PR #9, opened 2015-05-22
Allow passing a falsey value through `url` and still getting the default

mattrobenolt:
This is important if you're doing something like: 

```python
Client(url=os.environ.get('VAULT_ADDR'))
```

without repeating the logic and falling back to my own default.


ianunruh:
Thanks for the PR! I prefer `if not` checks over the shorthand `or` syntax, so I made the change in another commit.

mattrobenolt:
Awesome, thx. <3

PR #11, opened 2015-05-23
pypip.io is offline (expired domain) and seems to no longer be suppor…

zerosignal0:
…ted. Updating with badge.io replacement badge host.

this is a really minor update, but I figured I would be a good neighbor since I merged it earlier. :)


ianunruh:
Thanks! I was wondering why it wasn't rendering anymore :smiley:

PR #12, opened 2015-05-25
Add is_authenticated method

DavidWittman:
I found myself externally managing the authentication state of the
Vault client, and I thought a helper method may be useful here.

I also used this method to add a check of the token provided (if any)
during client initialization. If authentication fails, a `VaultError`
is raised.


ianunruh:
I'd like to avoid doing HTTP requests in the constructor. However, the rest of your PR looks good to me.

DavidWittman:
Yeah, I was a little on the fence about that one. It mostly stemmed from my desire to have the client fail fast and raise some exception other than an `InvalidRequest` when there was an authentication issue.

I'll amend that commit.

ianunruh:
LGTM

PR #15, opened 2015-05-28
Fix initialize() passed paramss

intr1nsic:


ianunruh:
Whoops, can't believe I missed that. Thanks for your contribution!

PR #21, opened 2015-07-25
Add support for building a wheel

mattrobenolt:
```
$ make publish
```


ianunruh:
I wasn't familiar with Wheel before but it seems useful. Thanks for the PR! The next release I publish will use this.

mattrobenolt:
Yup! Wheels are just generally faster to install. :)

PR #24, opened 2015-10-01
Require a modern `requests`

hobbeswalsh:
I recently had this break on me when my system already had an old version of Requests installed.


ianunruh:
Thanks for the PR. I've actually ran into this problem myself.

PR #26, opened 2015-10-11
Update __init__.py

DBkiller15:
Metods update and delete_field, if you do it will be useful.


ianunruh:
Thanks for submitting a PR. At this time, I don't feel like there is a common enough use case for this functionality to be included in the library.

PR #27, opened 2015-11-06
Fixing token-create metadata param name

yongtin:
https://vaultproject.io/docs/auth/token.html 

params is now called "meta" and not "metadata"

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -262,7 +262,7 @@ class Client(object):
         """
         self._delete('/v1/sys/audit/{}'.format(name))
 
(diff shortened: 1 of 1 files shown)


ianunruh:
Is this something that has changed in a recent release or has the documentation just been wrong?

yongtin:
It is what works for me. And from the vault code it seems like it has always been expecting "meta", align with the current doc. 

https://github.com/hashicorp/vault/blame/master/api/auth_token.go#L122

ianunruh:
Weird, the example responses and the parameter lists don't match up in the docs :confused: In that case, this PR looks good.

PR #28, opened 2015-11-09
Add support for pgp_keys when issuing a start_rekey

mwielgoszewski:
Also adds support for multiple keys to be passed to `rekey` via a new `rekey_multi` helper method, similar to the `unseal_multi` operation.

Integration tests pending.


ianunruh:
:+1:

PR #30, opened 2015-12-14
Adding caching capability to Vault client

hobbeswalsh:
This PR is just meant as a way of determining whether or not you want to include caching in the HVAC Vault client. Of course caching introduces a whole new suite of problems, so I wanted to get your input on it, @ianunruh .

This PR:

 * Adds a `cache=False` keyword arg to the `Client` instantiation. This doesn't change current behavior, but allows consumers to instantiate a caching client.
 * Adds a local (thread-safe) cache that does its best to respect Vault `lease_duration`s. If it can't find a `lease_duration` for a secret, it uses the default (30 days).
 * Adds ability to ignore the cache on any given `read()` operation.
 * Invalidates the cache when a `write()` or `delete()` operation is performed.

Again, this is just meant as an inquiry as to whether or not this behavior is desired in this Vault library. Thanks!

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -1,13 +1,14 @@
 from unittest import TestCase, skipIf
 
 from nose.tools import *
-import requests
 
 from hvac import Client, exceptions
 from hvac.tests import util
 
+
 def create_client(**kwargs):
     return Client(**kwargs)
+
 
 class IntegrationTest(TestCase):
     @classmethod
@@ -25,6 +26,9 @@ class IntegrationTest(TestCase):
         cls = type(self)
 
         self.client = create_client(token=cls.manager.root_token)
+        self.caching_client = create_client(
+            token=cls.manager.root_token,
+            cache=True)
 
     def test_unseal_multi(self):
         cls = type(self)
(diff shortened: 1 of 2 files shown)


ianunruh:
I think caching definitely belongs in HVAC and I like how you made it opt-in. Could you add some integration tests for when caching is enabled? I'd like to make sure that things like invalidation work as expected. Thanks!

hobbeswalsh:
@ianunruh Thanks, and sorry for not adding the integration tests earlier.

i couldn't find a way to exercise the cache expiry code path -- I don't know how to write a secret that has a TTL (or change the properties of a secret backend to change the global TTL of that backend). If you can think of a way to do that (and believe we should test it) then please let me know. Thanks!

hobbeswalsh:
Just realized I'd left this for a long time. Do you think I need to add some more tests here?

Thanks!

ianunruh:
After thinking about it for awhile, I think for now I'd rather leave caching to the users of hvac, rather than include it in the library. It's fairly trivial to add caching around calls in Python, and I'd like to keep the library as simple as possible. Thank you for your PR though, I appreciate it!

PR #32, opened 2015-12-29
Methods to simplify creating new userpass, app-id & user-id records

mtougeron:
I add 3 methods for helping make it easier (and obvious) to create new userpass, app-id & user-id records. I know that you could always call the `write` method with the full path but I felt that these methods are easier for users to understand.

The policies and app_id can be passed in as a string `policies='foo,bar'` or as a list `policies=['foo','bar']`. I felt it was easier to allow the list when a user has more than 1 policy instead of making sure the string was built properly.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -392,7 +392,7 @@ class Client(object):
 
         # Users can have more than 1 policy. It is easier for the user to pass in the
         # policies as a list so if they do, we need to convert to a , delimited string.
-        if type(policies) == list:
+        if isinstance(policies, (list, set, tuple)):
             policies = ','.join(policies)
 
         params = {
@@ -409,7 +409,7 @@ class Client(object):
 
         # app-id can have more than 1 policy. It is easier for the user to pass in the
         # policies as a list so if they do, we need to convert to a , delimited string.
-        if type(policies) == list:
+        if isinstance(policies, (list, set, tuple)):
             policies = ','.join(policies)
 
         params = {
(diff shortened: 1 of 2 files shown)


mtougeron:
FYI, I wasn't able to get `tox hvac.tests.test_tls_integration` tests to run. I kept getting a ssl cert verification error. So I only ran the `tox hvac.tests.test_integration` test suite.

hvac/v1/__init__.py
```diff
@@ -385,6 +385,69 @@ def auth_userpass(self, username, password, mount_point='userpass', use_token=Tr
...
         return self.auth('/v1/auth/{}/login/{}'.format(mount_point, username), json=params, use_token=use_token)
 
+    def create_userpass(self, username, password, policies, mount_point='userpass'):
+        """
+        POST /auth/<mount point>/users/<username>
+        """
+
+        # Users can have more than 1 policy. It is easier for the user to pass in the
+        # policies as a list so if they do, we need to convert to a , delimited string.
+        if type(policies) == list:
```
mwielgoszewski: Instead of `if type(policies) == list:`, consider `if isinstance(policies, (list, set, tuple)):` to capture `set` and `tuple` types too? Or the following (nicer, imo) syntax:

~~~
def create_userpass(self, username, password, mount_point='userpass', *policies):
    .. snip ..
    policies = ','.join(policies)
~~~

and do away with type checking altogether. Same would go for the `create_app_id` and `create_user_id` below.

mtougeron: Ah cool, wasn't aware of the `*policies` type syntax. Only downside of that is that I'd want it to be a required param where if it was after the `mount_point='userpass'` it would be considered optional, right?

mtougeron:
Awesome! Thanks.

ianunruh:
Thanks for your contribution!

Issue #34, opened 2016-01-23
client.write() returns 204


frodopwns:
In vault 0.4.0 I am getting 204 return codes from client.write() instead of 200.

Because of this check: https://github.com/ianunruh/hvac/blob/master/hvac/v1/__init__.py#L35

`None` is returned for both success and failure.

ianunruh:
This is expected behavior. If the write failed, then an exception would be thrown due to a non-200 response code. Returning `None` for `204 No Content` seems reasonable to me.

frodopwns:
What is the point of this then?

```py
 if response.status_code == 200:
            return response.json()
```

ianunruh:
Only some writes return content. For example, a write to the generic secret backend will return nothing and a `204 No Content` status code. But a write to the PKI secret backend to issue a certificate will return JSON about the new certificate and a `200 OK` status code.

It is up to the caller to know if there will be content from the `write` call they make.

PR #35, opened 2016-01-23
issue #34 - allow both 200 and 204 for success in client.write()

frodopwns:
This will allow users of 0.4.0 to check for success in client.write() calls.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -32,7 +32,7 @@ class Client(object):
         """
         response = self._put('/v1/{0}'.format(path), json=kwargs)
 
-        if response.status_code == 200:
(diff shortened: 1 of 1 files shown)


frodopwns:
Hmm I guess 204 means there is no json to be had now.  Guess there is more to be fixed.

frodopwns:
I'm not sure what is an appropriate content for the return from client.write() now that it does not seem to return the secret in json form.  I figured we should still return json for consistency but it had to be truthy for ease of use.

ianunruh:
This is expected behavior. If the write failed, then an exception would be thrown due to a non-200 response code. Returning None for 204 No Content seems reasonable to me.

PR #39, opened 2016-02-11
Add support for nonce during rekey operation

mwielgoszewski:
A nonce was added to the rekey operation in Vault 0.5.0. This PR provides an opportunity to provide that information.

Also, a test for JSON policies that I should have included in #33 from a while back.


ianunruh:
Thanks for the PR! Looks like the test scripts for Travis were a little out of date. I fixed them and tested your changes on my local machine.

PR #42, opened 2016-03-07
add license metadata to pkg

cburroughs:
This makes things slightly easier for computers to do automatic
checking


ianunruh:
LGTM

PR #43, opened 2016-03-16
add a 'get' companion for policies

cburroughs:
Basic rouding out of functionality.  Note that a simple `read` api
call is not sufficient to retrieve the original policy string.


ianunruh:
Looks good, thanks!

PR #44, opened 2016-03-16
fix rekey tests for vault 0.5.x

### LINKED ISSUE #40: Rekey tests fail against Vault 0.5.x
Due to the new requirement to provide `nonce` to rekey operations, the integration tests for HVAC fail against Vault 0.5.x. There should be a check around rekey operations for whether or not the nonce should be handled.

cburroughs:
To quote from the 0.5.x docs 'The rekey nonce operation must be
provided with each call.'.  This requires start_rekey to return
something other than None, which I believe is an innocuous api change.

fixes #40

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -294,10 +294,13 @@ class IntegrationTest(TestCase):
         self.client.cancel_rekey()
         assert not self.client.rekey_status['started']
 
-        self.client.start_rekey()
+        rekey_info = self.client.start_rekey()
 
+        nonce = None
+        if util.match_version('>0.5.0'):
+            nonce = rekey_info['nonce']
         for key in cls.manager.keys:
-            result = self.client.rekey(key)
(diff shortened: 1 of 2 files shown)


ianunruh:
Hello! Thanks for the PR. The Travis tests are failing just because they're being run against Vault v0.4.x. Either the integration tests need to be made to work against both versions or they can just be cleaned up to be tested against Vault v0.5.x only. I'm leaning towards the latter. Thoughts?

cburroughs:
Oh I see, `v1/sys/rekey/init` didn't return anything before.  I don't see a maintenance branch for 0.4.x or anything from hashicorp and I don't think there will be more 0.4.x releases.  So I'd also lean towards only testing against 0.5.x.

ianunruh:
Cleaned up the tests and fixed in 19f55ecd0baa4a2a9ea4d7c676ffa8f1aac3f8a6

PR #47, opened 2016-04-01
feat(list): add basic list code

drlkf:
from v0.5 onwards vault supports the `list' operation
https://www.hashicorp.com/blog/vault-0.5.html#list
there is not much to add to make it work with this package

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -69,6 +69,16 @@ class IntegrationTest(TestCase):
 
         self.client.delete('secret/foo')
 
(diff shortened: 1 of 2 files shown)


ianunruh:
Hi! Thanks for the PR. Before I accept this, could you fix a couple of things?

* Use `params={'list': 'true'}` to pass the `list=true` query param instead of embedding it into the URL
* Add a test case to the integration test suite

drlkf:
hello,

I have added both. sorry for the delay.

cheers

PR #49, opened 2016-04-14
POST

mattgiles:
`post` is required to access certain parts of the Vault API.


ianunruh:
Vault actually accepts POST/PUT interchangeably, according to the [documentation](https://www.vaultproject.io/docs/http/index.html). If you're looking for a generic PUT/POST call, you can use `Client#write`.

PR #50, opened 2016-05-04
Fix write errors when path contains unicode chars

pschmitt:
This fix is trivial, just import `unicode_literals`. Without this would fail when using Python 2:

```python
client = Client(token=myToken)
client.write('secret/Unicodeßstringé', value='unicode ftw!')
```


ianunruh:
Thanks for the PR! 😄

PR #53, opened 2016-05-30
Add requests session support

ahlinc:
Hi,

With multiple requests it's very useful to have one connection per host as it saves time significantly when a Vault host is far away.


ianunruh:
Thanks for the PR! Since one of my recent commits introduced a merge conflict for this PR, I have implemented this functionality in e17b4bf616e8abb7699cde1d3949865c296f9e2d

Issue #46, opened 2016-03-25
simplejson.scanner.JSONDecodeError: Expecting value: line 1 column 1 (char 0)


HeathNaylor:
For simplicity sake I have user `a` and `b` for my app_id and user_id respectively.

```
Traceback (most recent call last):
  File "/etc/salt/bash/vault.py", line 30, in <module>
    client.auth_app_id('a', 'b')
  File "/usr/local/lib/python2.7/dist-packages/hvac/v1/__init__.py", line 412, in auth_app_id
    return self.auth('/v1/auth/{0}/login'.format(mount_point), json=params, use_token=use_token)
  File "/usr/local/lib/python2.7/dist-packages/hvac/v1/__init__.py", line 524, in auth
    response = self._post(url, **kwargs).json()
  File "/usr/local/lib/python2.7/dist-packages/hvac/v1/__init__.py", line 561, in _post
    return self.__request('post', url, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/hvac/v1/__init__.py", line 587, in __request
    errors = response.json().get('errors')
  File "/usr/local/lib/python2.7/dist-packages/requests/models.py", line 808, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python2.7/dist-packages/simplejson/__init__.py", line 488, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/dist-packages/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/usr/lib/python2.7/dist-packages/simplejson/decoder.py", line 389, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.scanner.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Simplification of my script:
```
import os
import sys
import hvac

client = hvac.Client(url=os.environ.get("VAULT_ADDR"), verify=False)
client.auth_app_id('a', 'b')
```

ianunruh:
Which version of Vault are you using?

ianunruh:
Closing since I can't reproduce on my end. When using an invalid user id/app id combo, I get back an InvalidRequest exception, not a JSON parsing error.

ahlinc:
[...]

PR #54, opened 2016-06-01
Catch unhandled JSONDecodeError exception on text/plain responces [GH-46]

ahlinc:


ianunruh:
Thanks for the PR! One thought I had on this was to check that the `Content-Type` header is `application/json` before calling `response.json()`. I think it may be less hacky than relying on exception handling.

ahlinc:
Thank you for your comment, I've updated the PR.

ianunruh:
Thank you!

PR #55, opened 2016-06-01
Fix handling of Vault redirects

willusher:
Code currently appends redirected URL to original server URL. 
e.g: https://vault:8200https://vault-host0/v1/sys/mounts

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -612,7 +612,7 @@ class Client(object):
 
         # NOTE(ianunruh): workaround for https://github.com/ianunruh/hvac/issues/51
         while response.is_redirect and self.allow_redirects:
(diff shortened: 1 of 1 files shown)


ianunruh:
In my testing, the Vault server (v0.5.1) does not have an absolute URL in the Location header. Because of this, it is necessary to prepend the base URL to the relative URL provided from the Location header.

tilgovi:
Maybe `urljoin` is a better alternative that would work no matter what?

https://docs.python.org/2/library/urlparse.html#urlparse.urljoin

ianunruh:
@tilgovi thanks for the suggestion! I switched to `urljoin` in 327b73c33239ec0cb6df0b548e089096bcac9636

tilgovi:
Cheers!

PR #57, opened 2016-06-07
Added key backup for rekey endpoint

blarghmatey:
This commit adds the ability to backup sealing keys during a rekey
operation if there are PGP keys provided. This makes it easier to manage
automating the setup of Vault by not requiring to send the generated
sealing keys to the necessary parties at build time.


ianunruh:
Thanks for the PR!

blarghmatey:
No problem, thanks for the great library!

PR #58, opened 2016-06-14
Added an optional dependency of pyhcl

blarghmatey:
In order to more easily work with data returned from the `get_policy`
method I added the `pyhcl` parsing library to optionally return the
policy data as a Python dict. This has been added as an optional
dependency so that users of `hvac` are not required to have another
dependency in their projects if they don't need the additional functionality.


requirements.txt
```diff
@@ -1 +1,2 @@
 requests>=2.7.0
+pyhcl>=0.2.1,<0.3.0
```
TerryHowe: Should this just be test-requirements.txt?

blarghmatey: That is debatable, since the pyhcl library is used in the actual library if it is installed. It is an optional dependency in setup.py, but I think that for development purposes it makes sense to keep it in the regular requirements.txt

hvac/v1/__init__.py
```diff
@@ -271,12 +277,15 @@ def list_policies(self):
...
-    def get_policy(self, name):
+    def get_policy(self, name, parse=False):
         """
         GET /sys/policy/<name>
         """
         try:
-            return self._get('/v1/sys/policy/{0}'.format(name)).json()['rules']
+            policy = self._get('/v1/sys/policy/{0}'.format(name)).json()['rules']
+            if parse and has_parser:
+                policy = hcl.loads(policy)
```
ianunruh: If pyhcl isn't available but `parse=True`, I feel like this should raise an exception.

PR #60, opened 2016-06-16
Adding ttl option to create_userpass

stevenmanton:
Currently, you can't pass a `ttl` option through the `create_userpass` function. Many other functions accept `kwargs` and pass these to the POST/GET, so I followed a similar approach here. Also added a test.


ianunruh:
Thanks for the PR! lgtm

PR #61, opened 2016-06-16
[WIP] Adding EC2 authentication support

blarghmatey:
Added in primary methods for interacting with the EC2 authentication
backend that was introduced in the 0.6.0 release of Vault.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -479,6 +479,18 @@ class Client(object):
 
         return self.auth('/v1/auth/{0}/login/{1}'.format(mount_point, username), json=params, use_token=use_token)
 
+    def auth_ec2(self, pkcs7, nonce=None, role=None, use_token=True):
+        """
+        POST /auth/aws-ec2/login
+        """
+        params = {'pkcs7': pkcs7}
+        if nonce:
+            params['nonce'] = nonce
+        if role:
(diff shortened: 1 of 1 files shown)


hvac/v1/__init__.py
```diff
@@ -526,6 +538,102 @@ def create_user_id(self, user_id, app_id, cidr_block=None, mount_point='app-id',
...
+        """
+        POST /auth/aws-ec2/role/<role>
+        """
+        params = {
+            'role': role,
+            'bound_ami_id': bound_ami_id,
+            'disallow_reauthentication': disallow_reauthentication
+        }
+        for param in ['role_tag', 'max_ttl', 'policies']:
+            if eval(param) is not None:
```
ianunruh: I'm not a huge fan of using `eval`. Could you either make this more explicit or use `kwargs` instead?

ianunruh:
Thanks for fixing that up! Sorry it took so long to get merged, I didn't notice until now that your diff had been updated.

blarghmatey:
Not a problem. I had assumed that you were waiting on tests which I hadn't had a chance to create yet. I've just been working off of my fork for the time being. I've actually been pulled away from the work I was doing that needed this while I work on getting a major platform release completed.

Thanks for getting this merged!

PR #65, opened 2016-07-05
Fixing wrong parameter in function create_token

caiotomazelli:
As can be seen [in this line](https://github.com/hashicorp/vault/blob/v0.6.0/vault/token_store.go#L944) there is no "no_default_profile" parameter in vault api. 

This problem might have been motivated by [this issue](https://github.com/hashicorp/vault/issues/990)


ianunruh:
Thanks for the PR!

PR #69, opened 2016-07-27
Add accessor kwarg to token_lookup and token_revoke client methods

### LINKED ISSUE #68: Feature request: Lookup/revoke tokens by accessor
Vault 6.0 introduces token accessors. When a new token is created, a secondary id is provided. This can be retained to use to refer to the token but can't be used to access the resources the token holds permissions to.

Accessors can be used for lookups and revocations but additional API endpoints have been added to cater for them. The `revoke-orphan` and `revoke-path` operations do not support accessors.

It would be great if we could add an `accessor=False` kwarg to the `Client.lookup_token()` and `Client.revoke_token()` methods.

nicr9:
Setting `accessor=True` will change the API endpoints and raise an error if the user tries mixing them with other features (such as `revoke-orphan`).

This closes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,16 @@ client.revoke_token_prefix('zzz')
 client.renew_token('aaa')
 ```
 
+### Managing tokens using accessors
+
+```python
+token = client.create_token(policies=['root'], lease='1h')
+token_accessor = token['auth']['accessor']
+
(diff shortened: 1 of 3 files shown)


ianunruh:
Thank you! 😃

nicr9:
Happy to help!

PR #70, opened 2016-07-27
Support for Response wrapping

### LINKED ISSUE #67: Feature request: Response wrapping support
I saw that Vault 0.6 has a new [response wrapping](https://www.hashicorp.com/blog/vault-0.6.html#response-wrapping) feature. This could really simplify my workflow so I was hoping to add it to HVAC.

I've been looking through the Vault docs and CLI code to get a better idea of how this feature works and think it should be straight forward to implement.

Responses from vault will be wrapped if the request is sent with a `X-Vault-Wrap-TTL` header. This feature could be turned on by allowing `**kwargs` to be passed to any client method. They could then pass them along to `__request()` which would set the header if `wrap_ttl` is provided.

Unwrapping the secret would require adding an `unwrap(token)` method to the client. This would temporarily set `client.token` to match the provided `token` value so that it can read the wrapped secret from `cubbyhole/response`.

I'm planning to get started on implementing these changes today. When I have something worthwhile I'll update this thread.

nicr9:
Added the `wrap_ttl` kwarg to the following client methods: `read`, `unwrap`, `create_token`, `lookup_token`, `renew_token`, `get_app_id`, `get_user_id`.

Also added a new client method: `unwrap`.

This PR is built on top of the same branch in #69 because the unittests added here depend on accessors.

This closes #67.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,16 @@ client.revoke_token_prefix('zzz')
 client.renew_token('aaa')
 ```
 
+### Managing tokens using accessors
+
+```python
+token = client.create_token(policies=['root'], lease='1h')
+token_accessor = token['auth']['accessor']
+
+same_token = client.lookup_token(token_accessor, accessor=True)
+client.revoke_token(token_accessor, accessor=True)
+```
+
 ### Manipulate auth backends
 
 ```python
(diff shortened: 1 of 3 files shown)


nicr9:
Hey @ianunruh, any chance you could take a look at this when you have the time?

PR #79, opened 2016-09-15
Add support for 501 status code when Vault is uninitialized

ianwestcott:
Closes https://github.com/ianunruh/hvac/issues/78.

Adds an exception type of `VaultNotInitialized`, to be thrown when Vault responds with the status code 501. This behavior is new to Vault as of version 0.6.1. Details can be found here: https://www.vaultproject.io/docs/install/upgrade-to-0.6.1.html

I have tested this change by installing hvac as an editable package (with `pip install -e`) and running a script which utilizes hvac against an uninitialized cluster. I confirmed that hvac raises `hvac.exceptions.VaultNotInitialized` where it previously raised `hvac.exceptions.UnexpectedError`.

### DIFF
--- a/hvac/exceptions.py
+++ b/hvac/exceptions.py
@@ -25,6 +25,9 @@ class RateLimitExceeded(VaultError):
 class InternalServerError(VaultError):
     pass
 
+class VaultNotInitialized(VaultError):
+    pass
+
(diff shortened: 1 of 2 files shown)


ianunruh:
Thanks for the PR!

Issue #83, opened 2016-09-27
[Feature Request] Response wrapping for writes


aseppala:
Add response wrapping support for `write()`.

Here's an example usecase:

In AppRole authentication pull model it would be necessary to be able to create wrapped secret-ids.

This would be done like this with the vault client:
```
$ vault write -f -wrap-ttl="1m" auth/approle/role/testrole/secret-id 
Key                          	Value
---                          	-----
wrapping_token:              	f5afc0ab-d381-077a-5ab7-9372536f7e54
wrapping_token_ttl:          	1m0s
wrapping_token_creation_time:	2016-09-27 08:06:37.763722815 +0000 UTC
```

After `wrapping_token` is delivered to application, it can then fetch its `secret-id`:
```
$ vault unwrap f5afc0ab-d381-077a-5ab7-9372536f7e54
Key               	Value
---               	-----
secret_id         	144e13ce-b3af-5e6a-2a53-9704ef96b432
secret_id_accessor	0ec3f65c-52ad-480b-1d92-1f96d4524fea
```

Finally application can get its authentication token:
```
$ vault write auth/approle/login role_id=5b7cce50-1b32-8e32-6fc3-f4b99ba67f18 secret_id=4fdb8884-3795-5269-c24f-82e3069864c1
Key            	Value
---            	-----
token          	886e5dc6-ea74-82f4-ce77-579a7926ae71
token_accessor 	65050e7e-5183-7807-2121-c8520c324ee0
token_duration 	20m0s
token_renewable	true
token_policies 	[default]
```

aseppala:
Implemented in #85

PR #85, opened 2016-09-27
Support for write response wrapping

aseppala:
Adds generic response wrapping support for writes.
 
Related to #83


ianunruh:
Thanks for the PR!

PR #87, opened 2016-10-01
Added missing method(s) to tune a mount.

cypherhat:
Added tune_secret_backend and get_secret_backend_tuning. In the Vault documentation for the pki backend, there is an example:

```
$ vault mount-tune -max-lease-ttl=87600h pki
```

I couldn't find this in your module, so I added it.

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -112,13 +112,13 @@ class IntegrationTest(TestCase):
 
     def test_secret_backend_manipulation(self):
         assert 'test/' not in self.client.list_secret_backends()
-
+        print ("test")
         self.client.enable_secret_backend('generic', mount_point='test')
         assert 'test/' in self.client.list_secret_backends()
 
         self.client.tune_secret_backend('generic', mount_point='test', default_lease_ttl='3600s', max_lease_ttl='8600s')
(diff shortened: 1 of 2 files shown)


hvac/tests/test_integration.py
```diff
@@ -116,6 +116,10 @@ def test_secret_backend_manipulation(self):
         self.client.enable_secret_backend('generic', mount_point='test')
         assert 'test/' in self.client.list_secret_backends()
 
+        self.client.tune_secret_backend('generic', mount_point='test', default_lease_ttl='3600s', max_lease_ttl='8600s')
+        assert 'max_lease_ttl' in self.client.get_secret_backend_tuning('generic', mount_point='test')
+        assert 'default_lease_ttl' in self.client.get_secret_backend_tuning('generic', mount_point='test')
```
mtougeron: I think this needs to include a test that the tuning actually modified the `default_lease_ttl` and `max_lease_ttl` for the mount.

otakup0pe: Yes please. If we can get that test, we can merge.

marc-sensenich: I branched off of this PR and added the additional tests in #137

PR #89, opened 2016-10-11
Add use_token flag to the auth_approle function

mtougeron-gpsw:
Add `use_token` flag to the `auth_approle` function. Unlike the other methods, it needs to default to False for backwards compatibility.


mtougeron-gpsw:
@ianunruh Let me know if you'd rather me make the `use_token` default to `True` like the other methods. Not sure how much we need to worry about backwards compatibility with it but I went on the side of caution.

mtougeron-gpsw:
Doh, opened with wrong GH account.

PR #92, opened 2016-10-20
Fix zero length field name in format

mwielgoszewski:
Don't leave people stuck on Python2.6 in the dust.


ianunruh:
I think it would be good to support Python 2.6. I'll go ahead and merge this, fix up the remaining format statements, and add Python 2.6 to Travis.

PR #93, opened 2016-10-30
fix(approle): Update approle secret-id endpoints, tests

nicr9:
It seems the endpoints have changed in recent Vault versions because
including the `secret-id` in the url resulted in those ids being logged in 
cleartext. (See: https://github.com/hashicorp/vault/issues/1944)

Now get and delete operations are performed with POSTs against
`.../lookup` and `.../destroy` respectively with secret-ids specified in
the request body. (See: https://www.vaultproject.io/docs/auth/approle.html)

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -371,10 +371,10 @@ class IntegrationTest(TestCase):
         try:
             self.client.get_role_secret_id('testrole', secret_id)
             assert False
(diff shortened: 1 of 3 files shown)


otakup0pe:
It would be cool to get this merged!

ianunruh:
That's a strange change of them to make, but I suppose it is a newer API. Thanks for the PR!

Issue #95, opened 2016-11-02
New API proposal


ianunruh:
Hi all,

This library has been around for around 1.5 years now and in that time, Vault has changed quite a bit. The number of features and API surface has grown a huge amount. Just in the HVAC client class, there's 98 methods, each for a different API call. The problem is that the API is starting to become a bit unwieldy.

I'd like to get some feedback from the users of this library to figure out a better path going forward keep this client simple to use and maintainable as Vault adds more features in the future. I'd like to propose something similar to [Consulate](https://github.com/gmr/consulate)'s API (a Consul client library/CLI for Python).

We could start by breaking up related functionality into subclasses and then deprecate the old monolithic methods over time, as we get closer to a 1.0 release.

Here's an example of what the new API could look like:

```python
client = hvac.Client()

client.read('mykey')
client.write('mykey', foo='bar')

client.secret_backends.list()
client.secret_backends.enable('mysql')
client.secret_backends.disable('...')

client.token_roles.list()

client.app_roles.list()
client.app_roles.create('test')
client.app_roles.auth()
```

ianunruh:
@mwielgoszewski @nicr9 @mtougeron @otakup0pe @blarghmatey

nicr9:
This sounds like a great idea! :+1:

otakup0pe:
I'm game! As long as we pay attention to semantic versioning as we roll this out ;)

blarghmatey:
👍

nicr9:
I've tried to get the ball rolling with #96 

Let me know what you guys think!

mtougeron:
@ianunruh I like it.
@otakup0pe totally agree semantic versioning will be key.

Would be nice to also standardize some of the names for the method params too if time permits.

Also, let me know if you have a chunk that you'd like me to help out with. I'm happy to spend some time on it this weekend.

dsuch:
Hello,

this is just a reply from a user of the library, as originally requested.

While you are working on re-designing the API, could you please make __request and __raise_error methods read _request and _raise_error, respectively?

That is, I understand that they are private but the fact of like is that clients in this kind of libraries get embedded in larger projects and frameworks which means they also tend to get sub-classed to add custom functionality (e.g. bespoke headers or replacing the series of if's in __raise_error with a lookup dictionary),  and making methods private to a class simply makes it all a bit more cumbersome.

Other than that, the project looks very nice, nice job! :-)

nicr9:
[...]

PR #96, opened 2016-11-02
refactor(auth/token): Break out token roles functionality

nicr9:
I wanted to get a head start on this while the token roles stuff was fresh in my mind!

I've added two new classes: ClientFeature and TokenRoles. The former is a wrapper around the client for the basic API calls (post, read, list and delete). The later inherits the former and is where the new token roles functionality lives.

Also added further checks to current tests.

Motivation for changes is as discussed in #95.

**BREAKING CHANGES:**

* token_role(...) -> token_roles.read(...)
* list_token_roles(...) -> token_roles.list(...)
* create_token_role(...) -> token_roles.write(...)
* delete_token_role(...) -> token_roles.delete(...)

### DIFF
--- a/hvac/v1/token_roles.py
+++ b/hvac/v1/token_roles.py
@@ -1,0 +1,36 @@
+from .client import ClientFeature
+
+class TokenRoles(ClientFeature):
+    def write(self, role, allowed_policies=None, orphan=None, period=None,
+                          renewable=None, path_suffix=None, explicit_max_ttl=None):
+        """
+        POST /auth/token/roles/<role>
+        """
+        params = {
+            'allowed_policies': allowed_policies,
+            'orphan': orphan,
+            'period': period,
+            'renewable': renewable,
+            'path_suffix': path_suffix,
+            'explicit_max_ttl': explicit_max_ttl
+        }
+
+        return self._post('/v1/auth/token/roles/{0}'.format(role), json=params)
+
+    def read(self, role):
+        """
+        Returns the named token role.
+        """
+        return self._read('auth/token/roles/{0}'.format(role))
+
+    def delete(self, role):
+        """
+        Deletes the named token role.
+        """
+        return self._delete('auth/token/roles/{0}'.format(role))
+
+    def list(self):
+        """
+        GET /auth/token/roles?list=true
+        """
+        return self._list('auth/token/roles')
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -563,23 +563,23 @@ class IntegrationTest(TestCase):
         assert token['auth']['client_token'] == lookup['data']['id']
 
     def test_token_roles(self):
-        # No roles, list_token_roles == None
-        before = self.client.list_token_roles()
+        # No roles at beginning
+        before = self.client.token_roles.list()
         assert not before
 
         # Create token role
-        assert self.client.create_token_role('testrole').status_code == 204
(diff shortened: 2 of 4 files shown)


nicr9:
Hey @ianunruh,

Can I please get some feedback on this? It's my first attempt at tackling the new API design you proposed in #95.

If you think it looks good I can start working on applying the same strategy to other feature sections. I'd like to do as much as I can this month as part of [24pullrequests](24pullrequests.com).

P.S. If we go with a feature branch for the API changes let me know what the target branch should be and I'll re-target this PR.

hvac/v1/token_roles.py
```diff
@@ -0,0 +1,36 @@
+from .client import ClientFeature
+
+class TokenRoles(ClientFeature):
+    def write(self, role, allowed_policies=None, orphan=None, period=None,
```
ianunruh: Could we standardize on "create" instead of write?

hvac/v1/token_roles.py
```diff
@@ -0,0 +1,36 @@
...
+            'orphan': orphan,
+            'period': period,
+            'renewable': renewable,
+            'path_suffix': path_suffix,
+            'explicit_max_ttl': explicit_max_ttl
+        }
+
+        return self._post('/v1/auth/token/roles/{0}'.format(role), json=params)
+
+    def read(self, role):
```
ianunruh: Could we standardize on "get" instead of read?

nicr9:
Ah! :facepalm: There was a comment in the original issue about semantic versioning that make a lot more sense now...

I should have some spare time over the weekend to move this to `/v2/` and implement your suggested refactors. Thanks for the feedback!

otakup0pe:
I think I like where this is going.... Will look at some other PR's first, but as HCV adds more features this kind of structure will be key to grow `hvac`

nicr9:
If we get this merged I'd be happy to start working on some more of the existing features!

Keep in mind that it's Hacktoberfest so I've plenty of incentive to work toward feature parity with v1 for the rest of the month.

nicr9:
Hey @otakup0pe @ianunruh,

Have you guys got any further feedback for this?

otakup0pe:
I think we can merge this. How do we want to handle not maintaining duplicates of the functions? Maybe have the old v1 stuff call into the v2? Save that problem for when there is more api coverage?

nicr9:
@otakup0pe I think @ianunruh wanted to maintain the two versions [until v2 was ready](https://github.com/ianunruh/hvac/pull/96#pullrequestreview-13199827).

I like the idea of v1 functions calling into v2 but you might be right about holding off till we have more coverage. We want to be sure v2 doesn't introduce regressions before subjecting v1 to it without their cooperation. The only way we can be sure v2 is ready is if we have enough people using v2 first. What you think?

jeffwecan:
My two cents are that test coverage should be good enough to ensure no obvious regressions are introduced. I'm not a big fan of versioning a Python module interface as proposed here. Thats the point of the version on the package / module itself. So I would vote either for a v1.0.0 module branch that is worked on in parallel, or better yet, do the "strangler" pattern sort of idea and have the v1 functions call in to the newer code as alluded to.

With all that said, I'm a newcomer to this module in particular so I may be missing some historical context. In any case, did you still want to pursue getting this PR merged in @nicr9? I'm trying to get some of the outstanding PRs and issues cleaned up here...

PR #107, opened 2017-01-06
Decode the response object when listing ec2 roles

ianwestcott:
While writing code which manages EC2 roles in Vault, I noticed that the `list_ec2_roles()` method returns a `requests.Response` object, whereas other `GET`-based methods generally return a dict (decoded from the JSON response from Vault). I suspect this was an omission in the original code, but the fix is pretty straightforward. I also wrote some (very) minimal testing around EC2 role CRUD to ensure coverage of the fix. 

/cc @blarghmatey, who contributed the original code in https://github.com/ianunruh/hvac/pull/61, in case this change breaks something for you 😃

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -607,6 +607,11 @@ class IntegrationTest(TestCase):
         # create a policy to associate with the role
         self.prep_policy('ec2rolepolicy')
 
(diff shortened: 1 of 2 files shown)


otakup0pe:
Love the tests too 💖

ianwestcott:
@otakup0pe Thanks!

PR #110, opened 2017-01-24
Updated to make access and secret key optional

blarghmatey:
Rereading the documentation, the access key and secret key are not
required parameters as they can be retrieved from an IAM instance role
if it is present or from the ENV if they are set there on the Vault instance.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -124,24 +124,16 @@ class Client(object):
         """
         self._put('/v1/sys/seal')
 
-    def unseal_reset(self):
+    def unseal(self, key, reset=False):
         """
         PUT /sys/unseal
         """
         params = {
-            'reset': True,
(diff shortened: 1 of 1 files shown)


blarghmatey:
Hi @arthurlogilab I'm glad that you're finding some value in the Vault integration that I've been building. There is an [open PR](https://github.com/ianunruh/hvac/pull/110) for upstreaming that patch but it seems that the maintainer of the HVAC library has been on hiatus for a while as there are a number of other PRs that have not yet been merged. For my own use of HVAC I have been using the mitodl/hvac fork of the library which you are welcome to do as well.

jeffwecan:
@blarghmatey Looks like your fork's master branch may have drifted away from the original intent of this PR? I'm going to close it out in an effort to tidy things up, but feel free to open a new issue or PR as needed.

PR #112, opened 2017-02-02
Add delete_role method

freimer:
Self-explanatory, add delete_role method.


otakup0pe:
Can we get a test on this please ?

jeffwecan:
@freimer: I went ahead and made the small tweak of returning the DELETE response from this method and added an integration test for it. If you don't have any objections to the changes I pushed up to your branch, I would be happy to get the change merged in. 👍

PR #115, opened 2017-03-08
fix #88 use self.auth for approle auth.

### LINKED ISSUE #88: hvac.auth_approle does not store the token
Currently hvac.auth_approle doesn't call self.auth and instead calls the self._post directly. Due to this, the token in response is not stored internally.

gaelL:
This commit fix the issue #88 by using self.auth on approle auth.

Approle feature come from https://github.com/ianunruh/hvac/pull/77
but don't use `self.auth` as others auth methode like github, ...

This commit now allow you to auth directly with `client.auth_approle(role_id, secret_id)`.

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -381,6 +381,9 @@ class IntegrationTest(TestCase):
         role_id = self.client.get_role_id('testrole')
         result = self.client.auth_approle(role_id, secret_id)
         assert result['auth']['metadata']['foo'] == 'bar'
+        assert self.client.token == result['auth']['client_token']
+        assert self.client.is_authenticated()
+
         self.client.token = self.root_token()
         self.client.disable_auth_backend('approle')
 
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -921,7 +921,7 @@ class Client(object):
             params['meta'] = meta
         return self._post(url, json=params).json()
 
-    def auth_approle(self, role_id, secret_id=None):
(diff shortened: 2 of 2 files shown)


gaelL:
Is this project still maintained ?

Alan-R:
This looks like it probably fixes #122 also. Without the ability to specify use_token, you're kind of out of luck without doing the hack suggested in #122 (which I also currently use).

Alan-R:
I think this also fixes #122. Without the ability to specify use_token (and having it default to True), you're kind of out of luck...

wolftrouble:
Will this ever be merged? Going on nearly 6 months now for a 2-line change is a little frustrating. @otakup0pe you seem to have write access, can you help?

My only quibble would be the whole reason this was never caught is https://github.com/ianunruh/hvac/blob/b90d6745aadc405d1542c294737f578df9c74ae9/hvac/tests/test_integration.py#L391 doesn't ever try to assert is_authenticated() (unlike every other test). It might be better to add that to this commit.  @gaelL what do you think?

gaelL:
@wolftrouble indeed. I fixed it in my commit thanks

wolftrouble:
@gaelL Looks good, now let's just hope someone with commit permissions is willing to merge it.

Alan-R:
Hurray!  Thanks to @wolftrouble and @otakup0pe!

otakup0pe:
I should be able to get a new release on pypi by the weekend!

crgwbr:
@otakup0pe Any update on releasing this to PyPI?

otakup0pe:
Expect a legit 0.2.18 very soon !

PR #116, opened 2017-03-08
Add wrap_ttl to create_role_secret_id(..)

ramiamar:
1. Adding `wrap_ttl` to the method `Client.create_role_secret_id` allows getting a wrapped secret_id and only then passing it on to an app/machine/service which requires it.
2. `Client.auth_approle` now uses the `auth(...)` method and uses the client_token (instead of using `_post` and just returning the response)

Simple changes, important functionality.

🙏 thanks for considering

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='hvac',
-    version='0.2.17',
+    version='0.2.17.1',
     description='HashiCorp Vault API client',
     author='Ian Unruh',
     author_email='ianunruh@gmail.com',
--- a/setup.py
+++ b/setup.py
@@ -2,12 +2,12 @@
 from setuptools import setup, find_packages
 
 setup(
-    name='hvac',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,12 +2,12 @@
...
-    version='0.2.17',
+    name='alooma-hvac',
+    version='0.2.18',
     description='HashiCorp Vault API client',
-    author='Ian Unruh',
-    author_email='ianunruh@gmail.com',
-    url='https://github.com/ianunruh/hvac',
+    author='Ram Amar',
+    author_email='rami@alooma.com',
+    url='https://github.com/Aloomaio/alooma-hvac',
```
ericfode: I suspect you did not mean to add these changes to the pull request.

ramiamar: that is correct, @ericfode , apologies, as i changed these on our own fork .. i will undo these changes

ramiamar:
I'm closing this PR in favor of #159

PR #119, opened 2017-04-05
Fixes #71

### LINKED ISSUE #71: Feature request: Automate cubbyhole dance
Hashicorp has a blogpost detailing how to use cubbyhole backend for authentication with temporary tokens:

https://www.hashicorp.com/blog/vault-cubbyhole-principles.html

While that could be implemented on application side, it'd be great if hvac would have methods that could automate it. The API could be something simple like `client.auth_cubbyhole('TEMP_TOKEN')` which would then fetch the permanent token. I think that would suffice for both the push and pull models described in the blog.

tkinz27:
Small convenience method to use a temporary token to grab permanent token for future requests.

Fixes #71


tkinz27:
I've used it by subclassing hvac, but I dont have any unit tests for it.

otakup0pe:
If we could get a basic test in that would be amazing 💖

tkinz27:
Test added. Sorry should have done that originally, we use gatekeeper to setup the wrapped tokens and I never had actually created those before.

tkinz27:
The python 2.7 tests pass, but something is wrong with the setup of the python3.5

ianwestcott:
@tkinz27 the test issue should be fixed as of https://github.com/ianunruh/hvac/pull/143, can you merge master into your branch and see if that helps?

tkinz27:
I did a rebase not a merge. Not sure if travis will pick up the updates or not

jeffwecan:
Merged in master and tests are passing when I'm running them at least, 🚢 . 

Long-term, any method such as this that doesn't map directly to a Vault API route needs documentation if its to provide any sort of general utility (i.e., if there can be any expectation of folks actually knowing what its meant to be used for). However I think that increased documentation is needed across the board and can be done as a separate effort without blocking this fella.

PR #132, opened 2017-09-25
Add delete_role function

djlarsu:


jeffwecan:
Howdy @djlarsu! 

Looks like the test change that was added here ended up repurposing an existing test instead of adding a new one. We also have a similar PR open previously under https://github.com/ianunruh/hvac/pull/112 that should accomplish the same ends. So I'm going to go ahead and close out this PR, but if https://github.com/ianunruh/hvac/pull/112 doesn't cover exactly what you were after, feel free to reopen this PR or create a new issue with additional notes.

PR #137, opened 2017-10-24
Add Additional Testing for Mount Tuning

marc-sensenich:
Added the additional testing requested for #87 to assist with getting that PR merged in


otakup0pe:
@marc-sensenich can you pull in from master, the tests should be working now ✨

test/vault-tls.hcl
```diff
@@ -7,3 +7,6 @@ listener "tcp" {
 }
 
 disable_mlock = true
+
+default_lease_ttl = "768h"
```
marc-sensenich: I set these to the Vault defaults, so that the test for updating these values has a documented and configured base state to assert agains

jeffwecan:
Hey @marc-sensenich, I hope you don't mind but I went ahead and merged in the latest state of the master branch here so we could get this fella merged.

PR #139, opened 2017-10-27
enable passing the verify flag to requests

cariaso:
requests supports the ability to disable ssl verification or to customize the list of trusted certs. https://stackoverflow.com/a/12864892
http://docs.python-requests.org/en/master/user/advanced/?highlight=ssl#ssl-cert-verification

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -772,7 +772,7 @@ class Client(object):
             params['instance_id'] = instance_id
         return self._post('/v1/auth/aws-ec2/role/{0}/tag'.format(role), json=params).json()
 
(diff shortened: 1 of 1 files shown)


jeffwecan:
@cariaso: Turning off SSL verification can be done across all requests made by the hvac client by passing in False to the verify kwarg when constructing it. See: https://github.com/ianunruh/hvac/blob/master/hvac/v1/__init__.py#L22. If you have a compelling reason to have the ability to turn off SSL verification on a method-by-method basis we can discuss that in an issue, otherwise I am going to close this PR for now.

That said, it is worth making sure that option is documented generally if it isn't already. To that end I've opened this issue: https://github.com/ianunruh/hvac/issues/189

PR #143, opened 2017-11-08
Specify Python Versions To Test Against

ianwestcott:
as per Travis CI doc: https://docs.travis-ci.com/user/languages/python/

Note that specifying multiple Python versions in `.travis.yml` also requires defining the build matrix explicitly, otherwise Travis CI will attempt to build every env declaration against every Python version. See: https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

Closes https://github.com/ianunruh/hvac/issues/142.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,6 @@
 dist: trusty
 sudo: required
 language: python
-python:
-  - "2.6"
-  - "2.7"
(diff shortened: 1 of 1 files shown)


otakup0pe:
This is _awesome_ thank you 💖

PR #149, opened 2017-11-17
API Bug Fixes

mracter:
This PR includes two changes:

1. Changing the token unwrapping API call to call `POST /sys/wrapping/unwrap`  instead of calling the deprecated `GET /cubbyhole/response`. Commits between version `0.2.17` and `0.3.0` show some iterations that do call `/sys/wrapping/unwrap`, so the reversion may have been accidental.

2. Allowing integration tests to pass by changing the test for an intercepted token wrapper to assert `exceptions.InvalidPath` instead of `exceptions.Forbidden`, which is not returned by the Vault server.


jeffwecan:
This is somewhat duplicative of https://github.com/ianunruh/hvac/pull/119/ now that I see them both. Since the `auth_cubbyhole()` method covers this same use case, how would you feel about updating these changes to simply return the unwrap response without also updating the client token? 

For example, when using approle auth, its a good practice to push out a wrapped client secret-id. In that case, the end-user would want to unwrap the secret-id but it wouldn't have any business being used as a token immediately afterwards.

mracter:
That's fine, I've added commits with the suggested changes.

hvac/v1/__init__.py
```diff
@@ -77,16 +77,13 @@ def delete(self, path):
...
         """
-        path = "cubbyhole/response"
-        _token = self.token
-        try:
-            self.token = token
-            return json.loads(self.read(path)['data']['response'])
-        finally:
-            self.token = _token
+        payload = {
+            'token': token
```
jeffwecan: Would you mind making this token parameter optional? As mentioned in the [Vault documentation](https://www.vaultproject.io/api/system/wrapping-unwrap.html) regarding this parameter:
> Specifies the wrapping token ID. This is required if the client token is not the wrapping token. Do not use the wrapping token in both locations.

So as-is (without the token being optional), this method can be used to unwrap a token provided the *Client's* (i.e., the `hvac.Client` object) token attribute is valid. The other use case an optional token parameter would support is when you don't have any authentication token to begin with and want to do something like:

```
client = Client(token='some-wrapped-token')
unwrap_response = client.unwrap()
new_client_token = unwrap_response['auth']['client_token']
client.token = client_token

```

Though now that I've written up that example, I imagine that use case could be covered by calling the method with `client.unwrap(token='')` and/or `client.unwrap(token=None)`.

What do you think?

mracter: I think it would probably lead to less frustration if hvac inferred the location of the wrapped material should be instead of the user specifying it. The Vault API behaves similarly so no reason to add opacity. :)

I've made the changes, please would you take a look.

PR #155, opened 2017-12-20
Add IAM auth

s3u:
This fixes #127. Thanks in advance for review/merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,3 +235,16 @@ the latest `vault` binary is available in your `PATH`.
 ## Contributing
 
 Feel free to open pull requests with additional features or improvements!
+
+## IAM role auth
+
+If you're using an IAM role to authenticate:
+
+```
+import hvac
+vault_client = hvac.Client(url=VAULT_URL[, verify=somecert.pem])
+vault_client.auth_iam(role)
+vault_client.logout(revoke_token=True)
+```
+Add git+git://github.com/ExpediaDotCom/hvac.git@iam to your requirements.txt to use the fork.
+
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -1,6 +1,9 @@
 from __future__ import unicode_literals
 
 import json
+import base64
+from urllib.parse import urlparse
+import boto3
 
 try:
     import hcl
@@ -15,6 +18,7 @@ try:
     from urlparse import urljoin
 except ImportError:
     from urllib.parse import urljoin
+
 
 class Client(object):
     def __init__(self, url='http://localhost:8200', token=None,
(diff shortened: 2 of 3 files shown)


RevolutionTech:
This is a great feature! Our organization is doing this too, so I am very much looking forward to this PR being merged into master. 😃

RevolutionTech:
One issue I did notice with this PR however is that `boto3` is used as a new dependency and although it has been added to `requirements.txt` it has not been added to `setup.py`.

If this were merged in and uploaded to PyPi, then importing `hvac` in a fresh virtualenv would fail:
```
$ mkvirtualenv hvac-test
$ pip install hvac
$ python
>>> import hvac
...
ImportError: No module named boto3
```

README.md
```diff
@@ -235,3 +235,16 @@ the latest `vault` binary is available in your `PATH`.
 ## Contributing
 
 Feel free to open pull requests with additional features or improvements!
+
+## IAM role auth
```
dankolbman: This should probably be place in [Authenticate to Different Auth Backends](https://github.com/ianunruh/hvac#authenticate-to-different-auth-backends)

hvac/v1/__init__.py
```diff
@@ -978,6 +986,156 @@ def auth_approle(self, role_id, secret_id=None, mount_point='approle', use_token
...
+
+    def _authenticate_to_vault(self, vault_host, vault_port, role, verify):
+        payload = self._generate_vault_request(role, vault_host)
+
+        headers = {
+            'Content-type': 'application/json',
+            'Accept': 'text/plain'  # ,
+        }
+
+        response = requests.post('https://{}:{}/v1/auth/aws/login'.format(vault_host, vault_port),
```
jstensland: I just tried this and it worked great except I was running Vault with TLS disabled.  I think you should use the scheme of the url passed to the client instead of using https

hvac/v1/__init__.py
```diff
@@ -978,6 +986,156 @@ def auth_approle(self, role_id, secret_id=None, mount_point='approle', use_token
...
```
jstensland: Same here

flyinbutrs: just FYI, unless you're doing this on localhost or entirely within a controlled net, you're opening yourself up to having someone with a basic packet sniffer steal you AWS creds. Probably not a huge risk, but worth thinking about.

jstensland: My suggestion is that hvac support http since Vault does as well, but yes, it should only be used in dev environment and private network.

rberlind:
Any chance of getting this merged soon.  Would love to use.

pjelar:
I'm also interested in this

pjelar:
@s3u I've added a pr into your forked branch which includes updates to resolve the conflicts and setting urllib3 requirements.txt

omgapuppy:
also waiting for this to be merged

README.md
jeffwecan: Either that, or that *and* create a new third-level heading under that section.

README.md
```diff
@@ -235,3 +235,16 @@ the latest `vault` binary is available in your `PATH`.
...
+
+If you're using an IAM role to authenticate:
+
+```
+import hvac
+vault_client = hvac.Client(url=VAULT_URL[, verify=somecert.pem])
+vault_client.auth_iam(role)
+vault_client.logout(revoke_token=True)
+```
+Add git+git://github.com/ExpediaDotCom/hvac.git@iam to your requirements.txt to use the fork.
```
jeffwecan: Drop this and just add any new requirements needed to the upstream requirements.txt file please.

hvac/v1/__init__.py
```diff
@@ -978,6 +986,156 @@ def auth_approle(self, role_id, secret_id=None, mount_point='approle', use_token
...
+        :return:
+        """
+        parsed = urlparse(self._url)
+        self.token = self._authenticate_to_vault(parsed.hostname,
+                                                 parsed.port,
+                                                 iam_role,
+                                                 self._kwargs['verify'])
+        return self.token
+
+    def _authenticate_to_vault(self, vault_host, vault_port, role, verify):
```
jeffwecan: This is duplicative of other methods already extant in the library and should be dropped. E.g., the `auth_ec2()` method: https://github.com/ianunruh/hvac/blob/master/hvac/v1/__init__.py#L593-L603. I would recommend looking at the various other auth_ methods and making sure you are following the same pattern.

hvac/v1/__init__.py
jeffwecan: Wut? HTTPS / TLS is meant to encrypt data in transit...

hvac/v1/__init__.py
```diff
@@ -978,6 +986,156 @@ def auth_approle(self, role_id, secret_id=None, mount_point='approle', use_token
...
+        """
+        Generate a signed sts:GetCallerIdentity request to validate identify of the client.
+        The Vault server reconstructs the query and forwards it to STS service to authenticate
+        the client. See https://www.vaultproject.io/docs/auth/aws.html for more.
+
+        :param role: Role of this lambda
+        :return: Request body
+        """
+
+        client = boto3.client('sts')
```
jeffwecan: I personally would rather have any code that needs to lean on boto3 be handled outside of this client. If someone has cause to use hvac but doesn't need any of the AWS features, its not fair to expect them to install a boto dependency. The boto library is only being used for a couple of class in this PR, so leaving those calls to be handled by the caller instead of this library is not asking too much IMHO.

jeffwecan:
There are some largish problems with this PR that would prevent it from being merged as-is. If you would like any help addressing the points in the PR review I submitted though, I would be happy to assist. If this PR has languished long enough that the original submitter has lost interest, I can take a pass at implementing support for this authentication method sometime over the next month or so (🤞).

jeffwecan:
Since https://github.com/ianunruh/hvac/pull/170 has been merged and implemented a method with the same effective goal as this PR's changes, I'm going to go ahead and close this PR out. Feel free to update it or open a new one based on the related AWS IAM auth changes that have been merged in if needed!

RevolutionTech:
Sorry if this is a stupid question, but since the API in https://github.com/ianunruh/hvac/pull/170 now requires an access key ID and a secret access key, is there any way we can use IAM auth without having those at runtime? For the application I work on, we store those credentials in Vault currently.

I realize that we can infer credentials with `boto3` using the following code:
```python
import boto3
session = boto3.Session()
credentials = session.get_credentials()
print(credentials.access_key, credentials.secret_key)
```
but boto3 [attempts to collect](https://boto3.readthedocs.io/en/latest/guide/configuration.html) the access key and secret key from many different locations (such as environment variables, and from `~/.aws/credentials`) before it ultimately uses the IAM role, so if we were to rely on this we may not collect the credentials that correspond to the IAM role we want.

Is there some other way to extract an access key and secret key from an IAM role that I'm missing?

PR #164, opened 2018-01-24
fix #163 lookup_token function POST token not GET

### LINKED ISSUE #163: `lookup_token()` function returns a warning when looking up token
`lookup_token()` function returns a warning when looking up token.

solution: use POST instead of GET for token lookup

```
In [16]: client.lookup_token(token='d6b1a397-c412-da0e-e9c3-0549c32d118d')
Out[16]: 
{u'auth': None,
 u'data': {u'accessor': u'297ccbd4-4e40-5cb7-a75d-32caf0a7a035',
  u'creation_time': 1516646798,
  u'creation_ttl': 0,
  u'display_name': u'root',
  u'explicit_max_ttl': 0,
  u'id': u'd6b1a397-c412-da0e-e9c3-0549c32d118d',
  u'meta': None,
  u'num_uses': 0,
  u'orphan': True,
  u'path': u'auth/token/root',
  u'policies': [u'root'],
  u'ttl': 0},
 u'lease_duration': 0,
 u'lease_id': u'',
 u'renewable': False,
 u'request_id': u'd18f7738-571c-9bb7-fcee-70304d23a32a',
 u'warnings': [u'Using a token in the path is unsafe as the token can be logged in many places. Please use POST or PUT with the token passed in via the "token" parameter.'],
 u'wrap_info': None}

```

modifying the request to be with a POST + json `params` does nto produce the warning.
```
In [17]: client.lookup_token(token='d6b1a397-c412-da0e-e9c3-0549c32d118d')
Out[17]: 
{u'auth': None,
 u'data': {u'accessor': u'297ccbd4-4e40-5cb7-a75d-32caf0a7a035',
  u'creation_time': 1516646798,
  u'creation_ttl': 0,
  u'display_name': u'root',
  u'explicit_max_ttl': 0,
  u'id': u'd6b1a397-c412-da0e-e9c3-0549c32d118d',
  u'meta': None,
  u'num_uses': 0,
  u'orphan': True,
  u'path': u'auth/token/root',
  u'policies': [u'root'],
  u'ttl': 0},
 u'lease_duration': 0,
 u'lease_id': u'',
 u'renewable': False,
 u'request_id': u'0572f93f-9586-4160-e025-c4e88612eabf',
 u'warnings': None,
 u'wrap_info': None}

```

jnaulty:
lookup_token function should perform a POST to
avoid a warning about possibly leaking the token
```
 u'warnings': [u'Using a token in the path is unsafe as the token can be
 logged in many places. Please use POST or PUT with the token passed in
 via the "token" parameter.'],
```

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -450,14 +450,22 @@ class Client(object):
         GET /auth/token/lookup-accessor/<token-accessor>
         GET /auth/token/lookup-self
         """
+        token_param = {
+            'token': token,
+        }
+        accessor_param = {
+            'accessor': token,
+        }
         if token:
             if accessor:
-                path = '/v1/auth/token/lookup-accessor/{0}'.format(token)
-                return self._post(path, wrap_ttl=wrap_ttl).json()
+                path = '/v1/auth/token/lookup-accessor'
+                return self._post(path, json=accessor_param, wrap_ttl=wrap_ttl).json()
             else:
-                return self._get('/v1/auth/token/lookup/{0}'.format(token)).json()
+                path = '/v1/auth/token/lookup'
+                return self._post(path, json=token_param).json()
         else:
-            return self._get('/v1/auth/token/lookup-self', wrap_ttl=wrap_ttl).json()
+            path = '/v1/auth/token/lookup-self'
+            return self._get(path, wrap_ttl=wrap_ttl).json()
 
     def revoke_token(self, token, orphan=False, accessor=False):
         """


jnaulty:
@ti-mo We saw the same issue :smiley: 
I added some more changes to mine to make the format similar across all three requests.
It may be slightly more verbose, but I think it's clearer than putting everything inline.

jeffwecan:
Note (to self?): will need to call out that this breaks backwards compatibility with some older Vault versions in the changelog once its included in a release.

PR #166, opened 2018-01-25
Use request body for token lookups

ti-mo:
Vault responds with a 'warnings' field containing: 'Using a token in the path is unsafe as the token can be logged in many places. Please use POST or PUT with the token passed in via the "token" parameter.'


csawyerYumaed:
Any chance that this can get merged?

ti-mo:
@otakup0pe @jnaulty Which implementation are we going to pick?

otakup0pe:
I spoke with @ianunruh the other day and there is a concern about backwards compatibility. I will be extending the base `Client` class to have some element of (server) version detection, and then we will be able to merge in one of these changes (while maintaining compatibility with older HCV versions)

ti-mo:
@otakup0pe Alright, thanks for the info! Shall I leave this PR open as a reminder, or create an issue instead?

PR #168, opened 2018-02-01
Fix aws auth endpoint and returned value

iustinam:
* use endpoint specified in the docs (https://www.vaultproject.io/docs/auth/aws.html)
* remove duplicate '.json()' (the value from self.auth() is already a dict)  
* PEP8 formatting

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -16,7 +16,9 @@ try:
 except ImportError:
     from urllib.parse import urljoin
 
+
 class Client(object):
+
     def __init__(self, url='http://localhost:8200', token=None,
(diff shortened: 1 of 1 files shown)


iustinam:
I see that the same 'aws-ec2' endpoint is used in other functions as well , eg create_vault_ec2_client_configuration; I did not use/test those so I did not change them

jeffwecan:
@iustinam: If you're willing to update the other methods you referenced to parametrize the mount_point we can look at reopening and reviewing this PR. Otherwise I'll aim to get https://github.com/ianunruh/hvac/pull/181 merged in and released soonish which should cover your requirements.

PR #170, opened 2018-03-02
Add auth_iam_aws() method

ltm:
Support the [AWS IAM auth method](aws). This includes an implementation
of AWS SigV4 authentication rather than adding additional dependencies
(e.g. boto3).

[aws]: https://www.vaultproject.io/docs/auth/aws.html#iam-auth-method

### DIFF
--- a/hvac/aws_utils.py
+++ b/hvac/aws_utils.py
@@ -1,0 +1,42 @@
+import hmac
+from datetime import datetime
+from hashlib import sha256
+
+
+class SigV4Auth(object):
+    def __init__(self, access_key, secret_key, session_token=None):
+        self.access_key = access_key
+        self.secret_key = secret_key
+        self.session_token = session_token
+
+    def add_auth(self, request):
+        timestamp = datetime.utcnow().strftime('%Y%m%dT%H%M%SZ')
+        request.headers['X-Amz-Date'] = timestamp
+
+        if self.session_token:
+            request.headers['X-Amz-Security-Token'] = self.session_token
+
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
+        canonical_headers = ''.join('{0}:{1}\n'.format(k.lower(), request.headers[k]) for k in sorted(request.headers))
+        signed_headers = ';'.join(k.lower() for k in sorted(request.headers))
+        payload_hash = sha256(request.body.encode('utf-8')).hexdigest()
+        canonical_request = '\n'.join([request.method, '/', '', canonical_headers, signed_headers, payload_hash])
+
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4-create-string-to-sign.html
+        algorithm = 'AWS4-HMAC-SHA256'
+        credential_scope = '/'.join([timestamp[0:8], 'us-east-1', 'sts', 'aws4_request'])
+        canonical_request_hash = sha256(canonical_request.encode('utf-8')).hexdigest()
+        string_to_sign = '\n'.join([algorithm, timestamp, credential_scope, canonical_request_hash])
+
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4-calculate-signature.html
+        key = 'AWS4{0}'.format(self.secret_key).encode('utf-8')
+        key = hmac.new(key, timestamp[0:8].encode('utf-8'), sha256).digest()
+        key = hmac.new(key, 'us-east-1'.encode('utf-8'), sha256).digest()
+        key = hmac.new(key, 'sts'.encode('utf-8'), sha256).digest()
+        key = hmac.new(key, 'aws4_request'.encode('utf-8'), sha256).digest()
+        signature = hmac.new(key, string_to_sign.encode('utf-8'), sha256).hexdigest()
+
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4-add-signature-to-request.html
+        authorization = '{0} Credential={1}/{2}, SignedHeaders={3}, Signature={4}'.format(
+            algorithm, self.access_key, credential_scope, signed_headers, signature)
+        request.headers['Authorization'] = authorization
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -1,6 +1,10 @@
 from __future__ import unicode_literals
 
+import hmac
 import json
+from base64 import b64encode
+from datetime import datetime
+from hashlib import sha256
 
 try:
     import hcl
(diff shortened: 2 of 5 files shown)


rberlind:
I'm trying to use this by getting an error "hvac.exceptions.InvalidRequest".

Here is my code:
```
import boto3
client = hvac.Client(url='http://54.215.103.114:8200')
session = boto3.Session()
credentials = session.get_credentials()
client.auth_aws(credentials.access_key, credentials.secret_key, credentials.token)
```
My aws auth method is at path "aws" and does not have header_value set up at auth/aws/config/client.

Any ideas for me?  I would very much like to have this working tomorrow for a customer.  Thanks.

By the way, it seems to me that the new method should be called "auth_iam" rather than "auth_aws".

rberlind:
So, I figured out the problem and was able to get this to work.  You need to pass in the target Vault AWS authentication method role as an extra parameter.  I was able to modify your code by making the following changes:
line 594: `def auth_aws(self, iam_role, access_key, secret_key, session_token=None, header_value=None, mount_point='aws', use_token=True, **kwargs):`
line 642: `'role': iam_role,`

I then authenticated with: `client.auth_aws("lambda-qa-role", credentials.access_key, credentials.secret_key, credentials.token)`

Note that the "lambda-qa-role" is the Vault AWS auth role that I created with `vault write auth/aws/role/lambda-qa-role auth_type=iam bound_iam_principal_arn=arn:aws:iam::362381645759:role/lambda-vault-qa  policies=qa max_ttl=24h` while "lambda-vault-qa" is the IAM role the Lambda function is running as.

I hope that makes sense.

ltm:
@rberlind By default Vault will attempt to use a role with the same name as the IAM role. So it should work if you create the role like this:

```
vault write auth/aws/role/lambda-vault-qa auth_type=iam bound_iam_principal_arn=arn:aws:iam::362381645759:role/lambda-vault-qa policies=qa max_ttl=24h
```

However, if you do need to use a Vault role with different name than your IAM role you can provide it in the `kwargs`:

```
client.auth_aws(credentials.access_key, credentials.secret_key, credentials.token, role="lambda-qa-role")
```

rberlind:
Thanks for explaining, @ltm.  I figured you must have gotten it working somehow without the role.  I had not realized that kwargs could be used to pass arbitrary arguments. Even so, I think it would be much better to actually have the role as the first explicit argument since this would better match how authentication is done with the Vault CLI and with other client libraries such as the [Vault Ruby Client auth.rb](https://github.com/hashicorp/vault-ruby/blob/master/lib/vault/api/auth.rb).  (See line 196 of that code.)

Note that when setting up the Vault AWS roles with Vault commands, you can use a wildcard at the end of the role passed to the bound_iam_principal_arn parameter.  In that case, one would not necessarily have the Vault role correspond to any single AWS IAM role. 

Thanks again for having written this code.  It was very timely for me.

Please do consider changing the name of the method to "auth_iam" instead of "auth_aws".  Even better would be to change the existing auth_ec2 method to "auth_aws_ec2" and your method to "auth_aws_iam" since both are authenticating to Vault against the AWS auth method, one using the "ec2" option and the other using the "iam" option.

ltm:
@rberlind I'm hesitant to make `role` mandatory since it's optional in both the API and CLI (this PR is loosely based on [cli.go](https://github.com/hashicorp/vault/blob/master/builtin/credential/aws/cli.go)). However, I've added it as an explicit keyword argument.

I'm not opposed to renaming the method to `auth_iam()` or `auth_aws_iam`(), but it was my impression that AWS EC2 (`auth/aws-ec2`) is deprecated in favor of AWS IAM (`auth/aws`). As far as I can tell the CLI only supports IAM.

rberlind:
Thanks.  I work at HashiCorp and we do support both the ec2 and the iam options for the AWS auth method.  So, I think this library should keep both the ec2 and the iam auth methods and name them auth_aws_ec2 and auth _aws_iam to be clear what they are for.

I agree that role should not be mandatory since it is not mandatory in the auth/aws/login endpoint.

ltm:
Yeah, that makes sense then. I've renamed `auth_aws()` to `auth_aws_iam()`.

rberlind:
Thanks.  I guess renaming the original auth_ec2 method to auth_aws_ec2 might be tricky since it would break code for people already using.  Perhaps the solution is to add new auth_aws_ec2 method that calls original auth_ec2 method?

But that does not necessarily have to be done in this PR.

myoung34:
Any issues merging this? Its working for our test-cases

flyinbutrs:
Poke poke...  @otakup0pe / @ianunruh

jeffwecan:
@ltm: This looks mostly good to merge. Would you be willing to add in some test cases to ensure we have coverage for the new method? I recently opened a PR updating some of the auth_ec2 logic that included tests if you want something to reference along those lines: https://github.com/ianunruh/hvac/pull/181

ltm:
@jeffwecan I've added a test to `test_unit.py` and rebased with master.

hvac/v1/__init__.py
```diff
@@ -598,6 +602,63 @@ def auth_userpass(self, username, password, mount_point='userpass', use_token=Tr
...
+        canonical_request = '\n'.join([method, '/', '', canonical_headers, signed_headers, payload_hash])
+
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4-create-string-to-sign.html
+        algorithm = 'AWS4-HMAC-SHA256'
+        credential_scope = '/'.join([timestamp[0:8], 'us-east-1', 'sts', 'aws4_request'])
+        canonical_request_hash = sha256(canonical_request.encode('utf-8')).hexdigest()
+        string_to_sign = '\n'.join([algorithm, timestamp, credential_scope, canonical_request_hash])
+
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4-calculate-signature.html
+        key = 'AWS4{0}'.format(secret_key).encode('utf-8')
```
jeffwecan: The copious amount of `encode('utf-8')` calls give me some pause. I'm no expert at understanding all the implications that come with character encodings in python 2/3, but I have typically found if I'm having to fiddle with encodings this much, there is often an easier way of ending up with the same result. What was the motivation for including these?

hvac/v1/__init__.py
```diff
@@ -598,6 +602,63 @@ def auth_userpass(self, username, password, mount_point='userpass', use_token=Tr
...
+        key = hmac.new(key, timestamp[0:8].encode('utf-8'), sha256).digest()
+        key = hmac.new(key, 'us-east-1'.encode('utf-8'), sha256).digest()
+        key = hmac.new(key, 'sts'.encode('utf-8'), sha256).digest()
+        key = hmac.new(key, 'aws4_request'.encode('utf-8'), sha256).digest()
+        signature = hmac.new(key, string_to_sign.encode('utf-8'), sha256).hexdigest()
+
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4-add-signature-to-request.html
+        authorization = '{0} Credential={1}/{2}, SignedHeaders={3}, Signature={4}'.format(
+            algorithm, access_key, credential_scope, signed_headers, signature)
+        headers['Authorization'] = [authorization]
```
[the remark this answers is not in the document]
ltm: Yeah, makes sense. I've updated the PR.

hvac/v1/__init__.py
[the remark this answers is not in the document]
jeffwecan: I'm following ya, but I still have some vague fears that there is a better route for navigating that concern that could simplify things. However the output looks to be the same between python 2.7 and python 3.x. So I have no cause to block things on those fears. 👍

hvac/aws_utils.py
```diff
@@ -0,0 +1,42 @@
+import hmac
+from datetime import datetime
+from hashlib import sha256
+
+
+class SigV4Auth(object):
```
jeffwecan: This probably falls under personal preference, but I was hoping for this to go with a functional set up instead of a Class-based one. If you're amenable to it, I pulled down your branch and added some commits illustrating what I had in mind that could be merged in to your branch: https://github.com/ltm/hvac/compare/auth-aws...jeffwecan:auth-aws?expand=1

Since its not a deal breaker either way, just let me know if you'd rather stick with things as-is and we can get this merged.

ltm: The `SigV4Auth` approach is (very) loosely based on botocore (i.e. https://github.com/boto/botocore/blob/master/botocore/auth.py#L151). I do think it makes sense to keep the signature algorithm separate from the creation of the `GetCallerIdentity` request since this is specific to Vault. That said I don't have a strong preference either way so I'll defer to you.

jeffwecan: Make sense, I can live with this as our initial effort given that context.

jeffwecan:
FYI, I am aiming to publish a new version / release of the hvac package by the end of this week (i.e. June 5th) and that release should include this method. Thanks again for the contribution @ltm!

PR #174, opened 2018-03-23
add transit signature_algorithm parameter

crmulliner:


hvac/v1/__init__.py
```diff
@@ -1223,6 +1227,10 @@ def transit_verify_signed_data(self, name, input_data, algorithm=None, signature
             params['context'] = context
         if prehashed is not None:
             params['prehashed'] = prehashed
+        if sign_algorithm is not None:
+            params['signature_algorithm'] = sign_algorithm
+        else:
+            params['signature_algorithm'] = "pss"
```
jeffwecan: If this is going to be default, please just set it as such in the method signature. I.e., `, sign_algorithm = None):` => `, sign_algorithm='pss'):`

hvac/v1/__init__.py
```diff
@@ -1199,11 +1199,15 @@ def transit_sign_data(self, name, input_data, key_version=None, algorithm=None,
...
 
         return self._post(url, json=params).json()
 
     def transit_verify_signed_data(self, name, input_data, algorithm=None, signature=None, hmac=None, context=None,
-                                   prehashed=None, mount_point='transit'):
+                                   prehashed=None, mount_point='transit', sign_algorithm = None):
```
jeffwecan: nitpick: please remove the extra spaces around to be in line with PEP8.

crmulliner: will do

hvac/v1/__init__.py
crmulliner: will do

hvac/v1/__init__.py
```diff
@@ -1199,11 +1199,13 @@ def transit_sign_data(self, name, input_data, key_version=None, algorithm=None,
             params['context'] = context
         if prehashed is not None:
             params['prehashed'] = prehashed
+        if sign_algorithm is not None:
```
jeffwecan: These conditionals are probably safe to drop now, correct? Since we're supplying a default value for the `sign_algorithm` parameter, someone calling this method would have to explicitly pass in None to end up not meeting this conditional. Also if a caller did pass in None, I would imagine Vault would default to "pss" when receiving that value (but I don't know that for certain). 

How would you feel about simplifying this further and dropping these conditionals all together?

hvac/v1/__init__.py
```diff
@@ -1181,7 +1181,7 @@ def transit_generate_hmac(self, name, hmac_input, key_version=None, algorithm=No
         return self._post(url, json=params).json()
 
     def transit_sign_data(self, name, input_data, key_version=None, algorithm=None, context=None, prehashed=None,
-                          mount_point='transit'):
+                          mount_point='transit', sign_algorithm='pss'):
```
jeffwecan: I didn't catch this previously, but since the Vault API parameter is named "signature_algorithm", would you be okay using the same name for these method's for consistency?

jeffwecan:
@crmulliner: Thanks for getting back on this so quickly. After your last commit addressing my initial review comments, I realized there were a couple other points that could further simplify things here. Apologies for the back and forth, but let me know what you think.

PR #175, opened 2018-03-23
Add more methods for the userpass backend

ianwestcott:
There's a `create_userpass()` method and a `delete_userpass()` method, but no `list_userpass()`, `read_userpass()`, or policy/password update methods. _Until now._

Tests included.

### DIFF
--- a/hvac/tests/test_integration.py
+++ b/hvac/tests/test_integration.py
@@ -276,13 +276,20 @@ class IntegrationTest(TestCase):
         if 'userpass/' not in self.client.list_auth_backends():
             self.client.enable_auth_backend('userpass')
 
+        # add some users and confirm that they show up in the list
         self.client.create_userpass('testuserone', 'testuseronepass', policies='not_root')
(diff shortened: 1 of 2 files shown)


ianwestcott:
The test failure in https://github.com/ianunruh/hvac/pull/175/commits/0f87de971f94126ec87535fa3f0bb77f7fce1af4 seems to be caused by a breaking change in how Vault's userpass API returns "policies" data. I added an explanation of the problem alongside the fix in https://github.com/ianunruh/hvac/pull/175/commits/a7f9bfaa6be49915133e5314db3b9c9802a2c7c6. Note that I've stopped short of handling Vault's breaking change in hvac, as I'm not sure it's this library's place to mask breaking changes upstream.

jeffwecan:
Without looking over the changes quite yet, I think this module should aim to support the latest Vault release by default with a best case effort made to document the oldest Vault version it is compatible with. If folk's are unwilling or unable to keep their Vault service version up-to-date, they should make sure to pin the appropriate version of this module.

PR #176, opened 2018-03-26
added the ability to retain a url path for vault when needed

GuessRyansDevving:
added code to retain headers from sessions passed in to Client

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -13,20 +13,23 @@ from hvac import exceptions
 
 try:
     from urlparse import urljoin
+    from urlparse import urlsplit
 except ImportError:
     from urllib.parse import urljoin
+    from urllib.parse import urlsplit
 
 
 class Client(object):
     def __init__(self, url='http://localhost:8200', token=None,
                  cert=None, verify=True, timeout=30, proxies=None,
-                 allow_redirects=True, session=None):
+                 allow_redirects=True, session=None, keep_url_path=False):
 
         if not session:
             session = requests.Session()
         self.allow_redirects = allow_redirects
         self.session = session
         self.token = token
+        self.keep_url_path = keep_url_path
 
         self._url = url
         self._kwargs = {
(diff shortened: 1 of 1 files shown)


jeffwecan:
@GuessRyansDevving / @RyanSwedenburg-Turner : This could use a better description. I'm not entirely sure what the use case is here. The session headers makes enough sense, but I'm unclear what URL path is being retained...

Additionally, there are some test failures that need to be addressed before this PR could be considered.

hvac/v1/__init__.py
```diff
@@ -1244,10 +1247,19 @@ def _put(self, url, **kwargs):
     def _delete(self, url, **kwargs):
         return self.__request('delete', url, **kwargs)
 
-    def __request(self, method, url, headers=None, **kwargs):
+    def __request(self, method, url, **kwargs):
+        # NOTE(rswedenburg): added ability to retain the vault url's path when it's needed (for example when referencing a service path in Mesosphere)
```
jeffwecan: nit: There is no reason to prefix a comment with "NOTE"  (a comment is innately a note)  nor include the comment's author (that is what we use git for 😜).

hvac/v1/__init__.py
```diff
@@ -1244,10 +1247,19 @@ def _put(self, url, **kwargs):
...
+            split_url = urlsplit(self._url)
+            url = split_url[2] + url
         url = urljoin(self._url, url)
 
-        if not headers:
+        # NOTE(rswedenburg): assuming that the user wants to retain the headers on the session passed in, this grabs and uses them
+        if self.session:
+            headers = self.session.headers
+            if not headers:
+                headers = {}
```
jeffwecan: This breaks some previous usage of this method. If the method had custom headers passed in previously they no longer would be referenced here.

jeffwecan:
I'm hoping https://github.com/ianunruh/hvac/pull/212 will address at least part of this PR (the `keep_url_path` bits). If you'd like to see an option for passing in request sessions directly, please open up a new issue and/or PR for that in particular.

PR #182, opened 2018-05-17
client.write: kv uses POST instead of PUT

masteinhauser:
Vault 0.10 introduced a breaking change in the KV secrets enging HTTP
API. It requires HTTP POST instead of HTTP PUT, backwards compatible.

KV v2:
https://www.vaultproject.io/api/secret/kv/kv-v2.html#create-update-secret

KV v1:
https://www.vaultproject.io/api/secret/kv/kv-v1.html#create-update-secret


jeffwecan:
Looks sane to me, thanks! 👍

PR #190, opened 2018-06-10
Customizable approle path / mount_point

jeffwecan:
Solves for https://github.com/ianunruh/hvac/issues/178.

Auth backends can be mounted under arbitrary paths. This PR ensures custom paths can be used with the module and the approle auth backend in particular.

### DIFF
--- a/hvac/tests/test_approle_routes.py
+++ b/hvac/tests/test_approle_routes.py
@@ -1,0 +1,597 @@
+from unittest import TestCase
+
+import requests_mock
+from parameterized import parameterized
+
+from hvac import Client
+
+
+class TestApproleRoutes(TestCase):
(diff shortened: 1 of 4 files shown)


jeffwecan:
@bartwalczak1: This PR is meant to address your issue. I would appreciate your sign-off in the comments here as to whether it does that successfully or not. :D

bartwalczak1:
looks good thanks!

jeffwecan:
Noting for posterity: I added some unit tests for all the methods being changed here to ensure there were tests covering the specific changes (which ended up being good because I caught one potential bug while adding em!). As a result of that, I updated any approle methods that returned None to instead return their associated Vault response. That was needed for the test assertions but also provided additional information to the methods' callers if they desire it.

PR #195, opened 2018-06-11
Ensure mount_point Parameter for All AWS EC2 Methods

jeffwecan:
Builds on https://github.com/ianunruh/hvac/pull/181 and addresses the general issue raised in https://github.com/ianunruh/hvac/issues/194.

I did a quick first pass on these changes. I wanted to go ahead and open the PR to ensure the changes are staged but will give it a once over again before merging it in.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -1,14 +1,17 @@
 from __future__ import unicode_literals
 
 import json
+from base64 import b64encode
 
 try:
     import hcl
+
     has_hcl_parser = True
 except ImportError:
     has_hcl_parser = False
 import requests
 
+from hvac import aws_utils
 from hvac import exceptions
 
 try:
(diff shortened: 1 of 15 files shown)


hvac/v1/__init__.py
```diff
@@ -799,49 +799,49 @@ def create_vault_ec2_client_configuration(self, access_key, secret_key, endpoint
...
         """
         params = {'list': True}
-        return self._get('/v1/auth/aws-ec2/config/certificates', params=params).json()
+        return self._get('/v1/auth/{0}/config/certificates'.format(mount_point), params=params).json()
 
     def create_ec2_role(self, role, bound_ami_id=None, bound_account_id=None, bound_iam_role_arn=None,
                         bound_iam_instance_profile_arn=None, bound_region=None, bound_vpc_id=None, bound_subnet_id=None,
                         role_tag=None, max_ttl=None, policies=None, allow_instance_migration=False,
-                        disallow_reauthentication=False, **kwargs):
+                        disallow_reauthentication=False, mount_point='aws-ec2', **kwargs):
```
jeffwecan: note: worth considering how this could negatively callers of this function that took advantage of the "kwargs" bit of this method. Perhaps we should also fill in the rest of the corresponding API routes parameters as explicit keyword arguments to decrease the risk?

jeffwecan: I'm going to push up a commit to add in those parameters. However after giving it a go, I realize we'll potentially break usage one way or another regardless. E.g., positional arguments could end up as arguments for the wrong parameter. I'll just make sure to call out that fact in the changelog.

jeffwecan: Actually I just realized I had forgotten about the actual content of https://github.com/ianunruh/hvac/pull/140/ which overlaps here. I'm going to back out 91b263c and see about getting https://github.com/ianunruh/hvac/pull/140/ rebased on top of this mayhaps instead.

PR #206, opened 2018-06-28
Initial GCP auth backend implementation

seanmalloy:
I did some manual testing and I seems to be working. Need to write unit tests for this and do some additional testing. Might also add some additional GCP related methods too.

Any early feedback on this code would be appreciated.


hvac/v1/__init__.py
```diff
@@ -691,6 +691,18 @@ def auth_ec2(self, pkcs7, nonce=None, role=None, use_token=True, mount_point='aw
 
         return self.auth('/v1/auth/{0}/login'.format(mount_point), json=params, use_token=use_token)
 
+    def auth_gcp(self, role, jwt, use_token=True):
+        """
+        POST /auth/gcp/login
```
jeffwecan: Would be nice to give this the full docstring treatment. Historically what you have here has been the model, but I've been trying to provide more information on newly added methods. E.g.: https://github.com/ianunruh/hvac/blob/master/hvac/v1/__init__.py#L674-L685

seanmalloy: Fixed. Let me if the doc string needs further changes.

seanmalloy:
@jeffwecan I added unit tests and cleaned up my commit history. I believe this is now ready for final review. Let me know if any changes are needed.

jeffwecan:
In hindsight, I should have encouraged adding integration tests more. This codebase has historically relied on those test cases exclusively while unit tests of any kind are a recent addition. More importantly, there can and have been breaking changes in new Vault versions for this sort of method. Having integration tests provide some confidence around determining our minimum supported version of Vault. If you don't mind, please add a test case there. 

I'll grant that its a bit funky to add an integration test case in this sort of situation as we don't currently have any avenues for running those tests directly on a GCP instance / or a client/server with access to the GCP identity service. However there is still some value in having it. The `auth_ec2()` tests may be a decent model for to follow here (they test everything up to the actual authentication being successful): https://github.com/ianunruh/hvac/blob/master/hvac/tests/test_integration.py#L1191-L1217.

If you have any questions on that front. I'm also willing to chip in and push up an integration test to your branch if you'd like.

As an aside, I have felt that the entire integration suite could be improved in a few ways. One of which I went ahead and documented while writing up this comment: https://github.com/ianunruh/hvac/issues/207

seanmalloy:
Sounds good. I can create the integration tests.

seanmalloy:
@jeffwecan I created one integration test for this code change. Let me know if any additional changes are needed.

Issue #205, opened 2018-06-27
Silently fails when the vault address contains routes


damouse:
`urljoin` on line 1471 silently drops trailing routes on the Vault URL and fails, even though the URL is valid and works fine with the vault binary. 

Example: 
```
url = urljoin(`https://my.cloud.com/vault/`, '/v1/auth/token/lookup')
print(url)
# => https://my.cloud.com/v1/auth/token/lookup
```

Note that this works as expected with the vault command line binary:

```
export VAULT_ADDR=https://my.cloud.com/vault/
vault status
...
# what you expect to see
```

I'm sure using `urljoin` has benefits, but silently rewriting my URL in a case where Vault itself has no problem is a little off-putting.

Vault is hosted on a subroute because its sitting behind a reverse proxy via nginx, in order to share the system with other services on one server, domain, and one SSL config. Nginx rewrites the route when proxying and removes the `/vault`, of course.

jeffwecan:
Looks like https://github.com/ianunruh/hvac/pull/176 is perhaps related. Regardless this should definitely be addressed.

PR #212, opened 2018-07-05
Add custom urljoin method to retain original path

jeffwecan:
Addresses https://github.com/ianunruh/hvac/issues/205


jeffwecan:
@damouse this will hopefully resolve the issue you brought up in #205. Please let me know if that is true or not? :D

damouse:
It works, thank you kindly!

PR #216, opened 2018-07-06
Kvv2 support

Jukape:
Added kv v2 support with backwards compatibility with v1 for:

Read
Write
List
Delete

related to https://github.com/ianunruh/hvac/issues/183

Tested against:

Vault: 0.10.3, Backend v2
Vault: 0.10.3, Backend v1

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -26,14 +26,13 @@ class Client(object):
     # Kineticit - Add version parameter to support vault backend kv-v2
     def __init__(self, url='http://localhost:8200', token=None,
                  cert=None, verify=True, timeout=30, proxies=None,
-                 allow_redirects=True, session=None, version=1):
+                 allow_redirects=True, session=None):
 
         if not session:
             session = requests.Session()
         self.allow_redirects = allow_redirects
         self.session = session
         self.token = token
-        self._version = version
         self._url = url
         self._kwargs = {
             'cert': cert,
@@ -41,6 +40,15 @@ class Client(object):
             'timeout': timeout,
             'proxies': proxies,
         }
+
+        # auto detect backend version
+        try:
+            json_back = self.list_secret_backends()
+            if 'data' in json_back and 'kv/' in json_back['data']:
+                detected_version = json_back['data']['kv/']['options']['version']
+                self._version = int(detected_version)
+        except:
+            print('Cant determine backend version')
 
     def read(self, path, wrap_ttl=None):
         """
@@ -50,7 +58,7 @@ class Client(object):
             # Kineticit - Update URLs for vault backend kv-v2
             if self._version == 1:
                 return self._get('/v1/{0}'.format(path), wrap_ttl=wrap_ttl).json()
-            elif self._version == 2:
+            elif self._version >= 2:
                 return self._get('/v1/secret/data/{}'.format(path), wrap_ttl=wrap_ttl).json()
         except exceptions.InvalidPath:
             return None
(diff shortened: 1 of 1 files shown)


hvac/v1/__init__.py
```diff
@@ -39,12 +41,25 @@ def __init__(self, url='http://localhost:8200', token=None,
...
         }
 
+        # auto detect backend version
+        try:
+            json_back = self.list_secret_backends()
+            if 'data' in json_back and 'kv/' in json_back['data']:
+                detected_version = json_back['data']['kv/']['options']['version']
+                self._version = int(detected_version)
+        except:
+            print('Cant determine backend version')
```
jeffwecan: I'd like to avoid direct calls to print wherever possible. Assuming we simply want to log this behavior (i.e., raising an exception isn't needed), I would vote for use of the logging module. E.g.,:

Add `import logging` and `logger = logging.getLogger(__name__)` to the top-level of this source file. Then this line can be changed to `logger.warning('Cant determine backend version')` or something similar.

Jukape:
Updated to remove the print statement, instead there is an additional check for None so that no exception cases should happen in that block.

Out of interest how would you check the version without checking self._version ?

jeffwecan:
Could perhaps be done via a decorator plus an optional parameter on the methods that need to care about the versions. I'm not too concerned with that bit, but I'd be happy to try to whip up an example if you're interested in going that route.

I'd still like to see the print statement dropped entirely. That would ideally be as simple as:

* Adding `import logging`
* Declaring a logger at the top-level of the module: `logger = logging.getLogger(__name__)`
* Changing the print to a logger call: `logger.warn('Could not determine version etc. etc.')`

Jukape:
Hi, 

I did originally have it as a decorator and optional parameter, then I switch to the auto detect method.

I have dropped the print statement now.

hvac/v1/__init__.py
```diff
@@ -21,6 +21,9 @@
 
 
 class Client(object):
+    _version = 1
```
jeffwecan: Shouldn't this be a instance attribute rather than a class attribute? As-is, if someone wanted to instantiate two Client classes, one pointed to a Vault instance with kv version 1 and one with kv version 2, they wouldn't be able to. I could also imagine a case where someone wants to use the client with a kv version 1 secret backend under the path `secrets/` and a version 2 backend under `kv/`. In that case, we wouldn't want this to be an attribute at all. I.e., perhaps it makes more sense to detect the version every time we need to send any request to the kv backend? If we went that route, it would make sense to add a toggle for kv version auto-detection in case folks don't want the client to preform the extra requests needed to determine the version on each kv method call.

jeffwecan:
@Jukape: I don't see your commit changing the print statement, have you pushed that up?

Since this change has a chance to impact folks' use cases, I'm trying to be a bit more careful in considering it. Right now my tentative plans are to build on top of your changes, preform some testing myself locally (since I haven't yet switch my own vault usage over to kv version 2), and find a way to expand our test coverage to include cases for both kv versions. I hope to wrap up that work by next week and will check back in with some more concrete suggestions then.

Jukape:
Yeah it has been push on my fork.

jeffwecan:
Ultimately I went a different route to achieve this end: https://github.com/hvac/hvac/pull/260.

I did not feel comfortable adding any KV v2 accommodations to the existing read/write/etc. Client methods as they're most likely used for many purposes in addition to updating KV secrets.

As such, I'm going to close this PR out. However I welcome any feedback on the alternate proposal in #260.

PR #218, opened 2018-07-10
add read_lease

jvanbrunschot-coolblue:
This PR will add the 'read_lease' function so you can check how long it takes before you lease expires: 

https://www.vaultproject.io/api/system/leases.html#read-lease

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -272,10 +272,12 @@ class Client(object):
         return self._get('/v1/sys/leader').json()
 
     def read_lease(self, lease_id):
-        """
-        PUT /sys/leases/lookup
-        :param lease_id: str, specifies the ID of the lease to lookup.
-        :return dict, parsed JSON response from the leases PUT request
+        """PUT /sys/leases/lookup
+
+        :param lease_id: Specifies the ID of the lease to lookup.
+        :type lease_id: str.
+        :return: Parsed JSON response from the leases PUT request
+        :rtype: dict.
         """
         params = {
             'lease_id': lease_id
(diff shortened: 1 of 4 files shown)


hvac/v1/__init__.py
```diff
@@ -271,6 +271,15 @@ def ha_status(self):
         """
         return self._get('/v1/sys/leader').json()
 
+    def read_lease(self, lease_id):
+        """
+        PUT /sys/leases/lookup
```
jeffwecan: Could you please expand these docstrings a bit to capture the relevant information from the linked Vault documentation? E.g.: https://github.com/ianunruh/hvac/blob/master/hvac/v1/__init__.py#L670-L680

jvanbrunschot-coolblue: Added the additional docstrings..

about the integration tests: Looks like the `lease_id` is only available with dynamic secrets, from something like the `database` secrets engine. 

Enabling the `database` secrets engine and using it in the integration tests means that we need to spin up a mysql/postgres database to retreive a dynamic secret.

jeffwecan:
That's a fair point. If you don't mind sitting on this PR a few days, I'll try to come up with some potential solutions to more easily add test coverage for this sort of thing.

jvanbrunschot-coolblue:
Sure, I'm fine with that. If you need any help, please let me know.

jeffwecan:
Ended up finding this pro tip from the lead Vault developer re: easiest way to generate a lease ID without a bunch of extra setup being required: https://github.com/hashicorp/vault/issues/877#issuecomment-165988487

With that, I've added in an integration test case as well as a basic sort of unit test and opened a PR against your branch with those additions: https://github.com/jvanbrunschot-coolblue/hvac/pull/1. If you're down to merge in those test additions to the source branch on this PR, I can get everything merged and released. 👍

jeffwecan:
Thanks again @jvanbrunschot-coolblue! I am aiming to get this released in v0.6.2 in the next day or two.

Issue #208, opened 2018-07-03
InsecureRequestWarning


gerases:
Hi,

I have a custom CA and I'm getting `InsecureRequestWarning` because the cert is signed by my internal CA. I don't want to suppress the warning. What's the best way to handle this?

Thanks,
  S.

otakup0pe:
You can make use of the `REQUESTS_CA_BUNDLE` environment variable that is used by the underlying `requests` module.

Issue #111, opened 2017-01-25
Client Certificate CA


stongo:
Is there a way to specify a CA certificate when initializing client to go along with Certificate and Private Key in the `cert` tuple? The Absence of a way to specify CA file is a deal breaker for this library unfortunately.

ericksepulveda:
I think I'm having an issue related to this:

```bash
2017-01-30 16:59:37.543964 I | http: TLS handshake error from <host>: remote error: tls: unknown certificate authority
```

ganlaksh:
i am hitting the same problem.... using the verify=False option to workaround this.

bitprophet:
Fun note: the `verify` kwarg in the constructor and/or other methods, is passed through directly to an internal `requests.Session`, whose `verify` argument can _also_ take a string value which is the path to one's CA cert. Seems to work fine for me!

cscetbon:
It should at least be added to the documentation

otakup0pe:
I struggled with this a bit recently... Can add something to the README.

FabianFrank:
Passing a string that contains the path to a CA bundle, as described in the request docs works. You'll get a warning in some IDEs though, because the type hint in hvac is bool and not Union[str, bool] or just no type hint. I've addressed that in my PR.

jeffwecan:
Between #230 and #238 I am going to consider this issue closed. Please open up a new issue if you feel like additional work is needed.

PR #230, opened 2018-07-23
Add note on custom ca certs to README

otakup0pe:
This has come up a few times (#111 / #139 / #208) and the README changes covers a much better practice than disabling SSL validation.


jeffwecan:
Oh, though in our case, we set the relevant env var inside a helper method to get an authenticated hvac client. So might be worthwhile adding another example with something along the lines of:

```python
import os


# We use a self-signed certificate for the vault service itself, so we need to include our
# local ca bundle here for the underlying requests module.
os.environ['REQUESTS_CA_BUNDLE'] = '/etc/ssl/certs/ca-certificates.crt'
```

(In our case, this is referencing the CA bundle path found on Ubuntu distributions)

jeffwecan:
Hey @otakup0pe, I shuffled your addition around a bit and placed it under the `docs/` sub directory. Also added an additional example and pushed it up to your branch. Hope ya don't mind!

PR #232, opened 2018-07-26
Refactor adapters, Duplicate Auth Method

jeffwecan:
This PR intends to migrate the logic of the `Client.auth()` method into the Adapter class. This help facilitate moving auth methods (e.g., `Client.auth_aws_iam()` / `Client.auth_ec2()`) into separate classes in the future. In particular, I'm planning on breaking out aws auth method and secret backend methods into something like a `hvac.api.AWS` class in the future.

I also took this opportunity to move the logic of some of the more basic helper methods (`post()`, `get()`, etc.) into the abstract `Adapter` base class. Feels more logical to have those basic sort of methods in the base class since subclasses can still choose to override them or not as needed. Along those lines, I also moved the `Request` constructor method into the base class following the same reasoning.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -70,6 +70,11 @@ class Adapter(object):
 
         return '/'.join(map(lambda x: str(x).strip('/'), args))
 
+    def close(self):
+        """Close the underlying Requests session.
+        """
+        self.session.close()
+
     def get(self, url, **kwargs):
         """Performs a GET request.
 
@@ -108,11 +113,6 @@ class Adapter(object):
         :rtype: requests.Response
         """
         return self.request('put', url, **kwargs)
-
-    def close(self):
-        """Close the underlying Requests session.
-        """
-        self.session.close()
 
     def delete(self, url, **kwargs):
         """Performs a DELETE request.
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -151,6 +151,20 @@ class Adapter(object):
 
     @abstractmethod
     def request(self, method, url, headers=None, **kwargs):
+        """Main method for routing HTTP requests to the configured Vault base_uri. Intended to be implement by subclasses.
+
+        :param method: HTTP method to use with the request. E.g., GET, POST, etc.
+        :type method: str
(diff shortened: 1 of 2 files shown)


hvac/adapters.py
```diff
@@ -96,17 +58,24 @@ def __init__(self, base_uri='http://localhost:8200', token=None, cert=None, veri
...
+
+        return '/'.join(map(lambda x: str(x).strip('/'), args))
 
     def get(self, url, **kwargs):
         """Performs a GET request.
 
         :param url: Partial URL path to send the request to. This will be joined to the end of the instance's base_uri
             attribute.
-        :type url: str
+        :type url: basestring
```
dbwpe: `basestring` is in python2 to distinguish from unicode, but python3 has no such thing, it's all `str`.  as we discussed offline, defaulting to `str` seems more sensible, and tweak your IDE (if possible) to not make a fuss about an "incorrect" type (like unicode which is actually fine) being passed into the method

hvac/adapters.py
```diff
@@ -132,35 +101,84 @@ def put(self, url, **kwargs):
...
+            attribute.
+        :type url: basestring
+        :param headers: Additional headers to include with the request.
+        :type headers: dict
+        :param kwargs: Additional keyword arguments to include in the requests call.
+        :type kwargs: dict
+        :return: The response of the request.
+        :rtype: requests.Response
         """
+        raise NotImplemented
```
dbwpe: `NotImplementedError`

```python
>>> raise NotImplemented
  File "<stdin>", line 1, in <module>
TypeError: exceptions must be old-style classes or derived from BaseException, not NotImplementedType
>>> raise NotImplementedError
  File "<stdin>", line 1, in <module>
NotImplementedError
```

hvac/adapters.py
jeffwecan: Feast your eyes on this poetry: b853ae8

hvac/adapters.py
jeffwecan: As we discussed offline, I forgot which of these was which: df718a0

PR #238, opened 2018-07-31
Fix type and comment of verify parameter

FabianFrank:
So that it's clear that people can pass the path to a CA bundle.

Should address the confusion mentioned in #111.


jeffwecan:
@FabianFrank thank you so very much! 💯💯💯

Issue #129, opened 2017-09-11
Input validation might be good


ericfode:
It looks like most of these functions do nothing to keep you from passing invalid parameters to vault. It might be good if we did more.

otakup0pe:
It seems to me like addressing the root of #133 would help a _ton_ with input validation. If I'm not missing something, I'll close this one out in favor of 133.

jeffwecan:
This is something well keep in mind going forward, but without any recent input on the matter I figure we can close out the issue for now.

PR #244, opened 2018-08-06
New LDAP Auth Method Class

jeffwecan:
Similar idea to https://github.com/hvac/hvac/pull/242. Adding in a new "auth" class to cover the Vault LDAP auth method's API routes.


hvac/tests/integration_tests/api/auth/test_ldap.py
```diff
@@ -0,0 +1,356 @@
...
+]
+
+
+def load_test_cert():
+    """Load test cert for use when testing Ldap.configure's `certificate` parameter.
+
+    :return: Test certificate contents
+    :rtype: str | unicode
+    """
+    test_data_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', '..', '..', '..', 'test')
```
dbwpe: These '..'s smell funny - a helper method somewhere (hvac/tests/utils.py? or a new hvac/tests/test_files.py?) would be nice.

```python
from hvac.tests import test_files
test_cert = test_files.get_file_contents('server-cert.pem')
```

`get_file_contents(name)` could call a `get_filename(name)` that joins with a `get_base_dir()`.
removes some boilerplate, and later if test files are moved around, would only need to update `get_base_dir()`.

hvac/api/auth/ldap.py
```diff
@@ -0,0 +1,367 @@
...
+        :type policies: list
+        :param mount_point: The "path" the method/backend was mounted on.
+        :type mount_point: str | unicode
+        :return: The response of the create_or_update_group request.
+        :rtype: requests.Response
+        """
+        if policies is None:
+            policies = []
+        params = {
+            'policies': ','.join(policies),
```
dbwpe: I always ask myself in cases like this whether I want to enforce input types, or just say "rtfds" (read the frickin docstring).  this kind of thing has bitten me in the past:

```python
In [1]: policies = 'only-one-policy'

In [2]: ','.join(policies)
Out[2]: 'o,n,l,y,-,o,n,e,-,p,o,l,i,c,y'
```

If a bunk input type can result in Quite Bad Things, then I'll enforce the type (either with an exception, or help em out and make it a list), otherwise if the API (or whatever) handles it gracefully, just let the chips fall where they may.

jeffwecan: I added a ParamValidation exception a while back for this type of deal, would be reasonable to utilize it here. Could also make the case for casting things to a list if given a string where a list is expected, but I think its better to alert the user to the correct param type rather than hide it away implicitly...

hvac/tests/integration_tests/api/auth/test_ldap.py
jeffwecan: @dbwpe: Does 79fc92c1244743d0c7adf0d56800efac931c5242 cover what you had in mind?

hvac/api/auth/ldap.py
jeffwecan: 6210eeb. Touches on https://github.com/hvac/hvac/issues/129 to an extent.

hvac/tests/integration_tests/api/auth/test_ldap.py
jeffwecan: Then I read your actual comment and bc57308 is closer to the mark...

PR #245, opened 2018-08-06
New MFA Auth Method Class

jeffwecan:
Similar idea to https://github.com/hvac/hvac/pull/242 and https://github.com/hvac/hvac/pull/244. Adding in a new "auth" class to cover the Vault Mfa auth method's API routes.

Also related to the spirit of https://github.com/hvac/hvac/issues/95.


hvac/api/auth/__init__.py
```diff
@@ -1,7 +1,9 @@
 """Collection of classes for various Vault auth methods."""
 
 from hvac.api.auth.github import Github
+from hvac.api.auth.mfa import Mfa
```
dbwpe: this may end up conflicting with https://github.com/hvac/hvac/pull/244/files#diff-6584ecf07e2d3f55775b4244648aa47d once it's merged in

whether it does or not, I get to appear thoughtful and wise

hvac/api/auth/mfa.py
```diff
@@ -0,0 +1,172 @@
...
+        :return: The response of the configure_duo_behavior request.
+        :rtype: requests.Response
+        """
+        params = {
+            'username_format': username_format,
+        }
+        if push_info is not None:
+            params['push_info'] = push_info
+        if user_agent is not None:
+            params['user_agent'] = user_agent
```
dbwpe: "user_agent" is passed as a param, rather than a "User-Agent" header?  I guess the vault code [somewhere around hereabouts](https://github.com/hashicorp/vault/blob/8142b42d951119a73ce46daa3331921b5e21cdee/helper/mfa/duo/path_duo_config.go#L16-L19) translates it to a header when making a call to duo?

also, might it be reasonable to default to 'github.com/hvac' or similar?

hvac/tests/integration_tests/api/auth/test_mfa.py
```diff
@@ -0,0 +1,240 @@
...
+                    password=password,
+                    mount_point=TEST_AUTH_PATH,
+                )
+            self.assertIn(
+                member=exception_message,
+                container=str(cm.exception),
+            )
+        else:
+            raise SkipTest('Unable to fully test login flow (easily) without working TLS for mock Duo server')
+            login_response = self.client.auth_userpass(
```
dbwpe: having code under an unconditional `raise` smells funny.  not hurting anything, but maybe comment out the code code under raise? meh.

hvac/v1/__init__.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, url='http://localhost:8200', token=None,
 
         # Instantiate API classes to be exposed as properties on this class
         self._github = api.auth.Github(adapter=self._adapter)
+        self._mfa = api.auth.Mfa(adapter=self._adapter)
```
dbwpe: ditto here on possible mergeflict with https://github.com/hvac/hvac/pull/244/files#diff-b4888de0c71be81f2e659f73aefaa228R64

PR #250, opened 2018-08-22
Send token as a param, to avoid token been logged by the audit log

rastut:
Token should be sent as a parameter in a POST request to the renew endpoint. This will avoid token getting logged in an audit logs, it will also comply with current vault API policy (a warning is produced when token is sent as a URL parameter).

` Using a token in the path is unsafe as the token can be logged in many places. Please use POST or PUT with the token passed in via the "token" parameter.`

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -946,10 +946,10 @@ class Client(object):
         """
         params = {
             'increment': increment,
-            'token': token,
         }
 
-        if token:
+        if token is not None:
+            params['token'] = token
             return self._adapter.post('/v1/auth/token/renew', json=params, wrap_ttl=wrap_ttl).json()
         else:
             return self._adapter.post('/v1/auth/token/renew-self', json=params, wrap_ttl=wrap_ttl).json()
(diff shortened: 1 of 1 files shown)


hvac/v1/__init__.py
```diff
@@ -946,11 +946,11 @@ def renew_token(self, token=None, increment=None, wrap_ttl=None):
         """
         params = {
             'increment': increment,
+            'token': token,
```
jeffwecan: Perhaps it makes more sense to move setting this `token` key in the params dict to under the following conditional? I.e., no need to send it in the params if we're doing the renew-self call:

```python
def renew_token(self, token=None, increment=None, wrap_ttl=None):
    params = {
        'increment': increment,
    }

    if token is not None:
        params['token'] = token
        return self._adapter.post('/v1/auth/token/renew', json=params, wrap_ttl=wrap_ttl).json()
    else:
        return self._adapter.post('/v1/auth/token/renew-self', json=params, wrap_ttl=wrap_ttl).json()

```

rastut: Yes sure! :+1:

PR #254, opened 2018-08-30
Fix policy rules fetching from json

seuf:
Since a few version of Hashicorp Vault, the policy rules are returned inside a `data` object in the json.
This PR add a test if rules is present at the root of the json. Else it fetch rules from the `data.rules`.

Related issue : https://github.com/TerryHowe/ansible-modules-hashivault/issues/81

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -685,7 +685,11 @@ class Client(object):
         :rtype:
         """
         try:
-            policy = self._adapter.get('/v1/sys/policy/{0}'.format(name)).json()['rules']
+            data = self._get('/v1/sys/policy/{0}'.format(name)).json()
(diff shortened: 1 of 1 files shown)


drewmullen:
Thanks @seuf ! I copied these changes into my hvac install, removed the pre-complied __init__.pyc, and im still getting the error, same as before: "exception: [rules]"

am i missing something?

drewmullen:
Thanks @seuf this has fixed the policy issue! great work!

Dudesons:
@drewmullen There is an ETA for a new release with this code ?

jeffwecan:
@Dudesons I'm hoping to include this change in hvac v0.6.4 that is intended to be released by Monday, September 3rd (ideally).

PR #264, opened 2018-09-12
Fix sys top level injection on sys/ routes

jacquat:
In vault 0.11.0, there is a deprecation with the routes sys/*.

You can find out more about the changes here :  https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#0110-august-28th-2018 

This change will get the "data" field of the sys/* routes (or return None if there is no field "data").


jeffwecan:
Thanks for the contribution @jacquat! How would you feel about implementing some conditional logic to fallback to the previous response structure (i.e., from earlier versions of Vault) in order to maintain some backwards compatibility? There is an example of this in the related changes made in https://github.com/hvac/hvac/pull/254 (though worth noting I ended up tweaking things a bit more after merging in that PR: https://github.com/hvac/hvac/blob/v0.6.4/hvac/v1/__init__.py#L700-L711)

jeffwecan:
So I've fiddled with adding tests (including runs against Vault v0.10.4, v0.11.0, and v0.11.1) for the affected methods. In doing so, I realize I somewhat misunderstood the change that was made in v0.11.0 (the associated change has been reverted as of v0.11.1 but is still intended to be enacted against in the future). Long story short, I realize my request to add conditional logic around where we return the "data" key or not in these methods was not strictly necessary. In particular, we only would need to include that logic for support with Vault versions before v0.6.1. I'm okay dropping that support as that version is over two years old at this point.

I've pushed up some tweaks to your branch @jacquat, hope you don't mind!

PR #265, opened 2018-09-12
Add support to namespaces

stenio123:
Namespaces is a new feature released with Vault 0.11.0 enterprise. 

More information here: https://www.vaultproject.io/guides/operations/multi-tenant.html

This change shouldn't impact users not using namespaces. Meanwhile, users that are using namespaces will have the option to configure the client accordingly.

For reference, there is a similar PR under evaluation for vault-ruby: https://github.com/hashicorp/vault-ruby/pull/183


hvac/adapters.py
```diff
@@ -20,14 +20,16 @@ class Adapter(object):
     """Abstract base class used when constructing adapters for use with the Client class."""
     __metaclass__ = ABCMeta
 
-    def __init__(self, base_uri=DEFAULT_BASE_URI, token=None, cert=None, verify=True, timeout=30, proxies=None,
+    def __init__(self, base_uri=DEFAULT_BASE_URI, token=None, namespace=None, cert=None, verify=True, timeout=30, proxies=None,
```
jeffwecan: Will need to update the constructor args for the `Client` class as well if we want to enable the type of usage outlined in the additions to the readme. Also would you be down with putting this addition to the end of the argument list here so it wouldn't interfere with folks passing in positional arguments?

PR #271, opened 2018-09-21
Login and List Groups Methods for Okta Auth Method

kanna0110:


jeffwecan:
Howdy @kanna0110, I'm trying to limit any additions being made to the Client class here as its already grown quite large. For new methods (such as the Okta auth method-related logic being added here), I've been following a pattern of adding new specific classes. For instance, here is what we ended up adding for the MFA auth method: https://github.com/hvac/hvac/pull/245

You wouldn't need to implement all the API routes available for the Okat auth method, but I would like to see any methods you need added to a new class. If you're not interested in going down that route, I'd be happy to see about adding in Okta support along the lines I'm suggesting over the next month or so.

jeffwecan:
Closing this fella and going with #341

PR #272, opened 2018-09-23
added support for the status code in kv v2

MoisesRT:


jeffwecan:
Howdy @MoisesRT, this `write()` method is intended to be more of an analog to Vault CLI's `write` command. As such, I wasn't intending on adding KV-specific logic to these type of methods. 
If there is something with the existing KV-specific classes outlined at the following link, please submit an issue or PR making changes there: https://hvac.readthedocs.io/en/latest/usage/secrets_engines/kv.html

PR #277, opened 2018-09-28
Patch Method for KvV2 Class

### LINKED ISSUE #270: support for vault kv "patch" option
I don't see a way to do a "patch" operation which appends to a current key rather then overwriting it.   

vault kv put secret/test foo=bar

stores 
```
foo=bar
```

vault kv put secret/test foo=bar2 

stores
```
foo2=bar2
```
but
vault kv patch secret/test foo3=bar3

stores
```
foo2=bar2
foo3=bar3
```

so it appends rather then overwrites (as long as the key isn't the same).  

does HVAC have a way to support "patch" operations?

jeffwecan:
Closes #270

cc: @slimm609 in case you have any input on how these changes look given you created the original issue.

### DIFF
--- a/docs/usage/secrets_engines/kv_v2.rst
+++ b/docs/usage/secrets_engines/kv_v2.rst
@@ -149,6 +149,23 @@ Create/Update Secret
         cas=0,
     )  # => Raises hvac.exceptions.InvalidRequest
 
+Patch Existing Secret
+---------------------
+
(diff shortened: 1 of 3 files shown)


slimm609:
Was out last week, but gonna try to test it today/tomorrow.

jeffwecan:
That is an interesting idea... I may follow-up with an issue on that point (`create_or_patch_secret`) in particular to see if folks would be interested in the option. However I'll limit this PR to just adding the patch method (now that you've done me the solid of looking it over). Thanks @slimm609!

Issue #231, opened 2018-07-23
Python 2.6 incompatibility after 0.5.0


zhogov:
Following line is not compatible with Python 2.6, but works with Python 2.7:

https://github.com/hvac/hvac/blob/fa44c68a7ad1f38eaba3002f960ae62828c112b0/hvac/v1/__init__.py#L1111

Command used:
```
ansible -vvvvv -m debug -a msg="{{ lookup('hashi_vault', 'secret=path:contents token=my-token url=https://vault.my-company.net') }}" localhost
```
Error message:
```
ansible 2.5.5
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.6.6 (r266:84292, Aug 18 2016, 15:13:37) [GCC 4.4.7 20120313 (Red Hat 4.4.7-17)]
Using /etc/ansible/ansible.cfg as config file
setting up inventory plugins

The full traceback is:
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/ansible/executor/task_executor.py", line 138, in run
    res = self._execute()
  File "/usr/lib/python2.6/site-packages/ansible/executor/task_executor.py", line 506, in _execute
    self._task.post_validate(templar=templar)
  File "/usr/lib/python2.6/site-packages/ansible/playbook/task.py", line 260, in post_validate
    super(Task, self).post_validate(templar)
  File "/usr/lib/python2.6/site-packages/ansible/playbook/base.py", line 390, in post_validate
    value = templar.template(getattr(self, name))
  File "/usr/lib/python2.6/site-packages/ansible/template/__init__.py", line 525, in template
    disable_lookups=disable_lookups,
  File "/usr/lib/python2.6/site-packages/ansible/template/__init__.py", line 481, in template
    disable_lookups=disable_lookups,
  File "/usr/lib/python2.6/site-packages/ansible/template/__init__.py", line 717, in do_template
    res = j2_concat(rf)
  File "<template>", line 8, in root
  File "/usr/lib/python2.6/site-packages/Jinja2-2.6-py2.6.egg/jinja2/runtime.py", line 180, in call
    return __obj(*args, **kwargs)
  File "/usr/lib/python2.6/site-packages/ansible/template/__init__.py", line 615, in _lookup
[...]

PR #291, opened 2018-10-10
'list' function python 2.6 compatibility

jasonarewhy:
Other functions use '/v1/{0}' format, 'list' should as well. The {} format doesn't work in python 2.6.


jeffwecan:
@jasonarewhy: This module only supports Python 2.7 / 3.x. I'm fine merging in this PR as it certainly won't hurt anything but we make no guarantees that any sort of compatibility with Python 2.6 will be maintained over time.

Semi-related issue:https://github.com/hvac/hvac/issues/231

PR #326, opened 2018-10-31
Handle beta version string

jeffwecan:
This will hopefully fix the `ERROR: Failure: ValueError (invalid version number '1.0.0-beta1')` error seen on the "head" Vault version travis CI jobs...


jeffwecan:
Still have some failing tests on the HEAD Vault branch, but at least we can see those now! 👍

PR #303, opened 2018-10-28
[WIP] Implement Transit engine (#302)

alexandernst:
This is still a WIP. Implements #302

PR #327, opened 2018-10-31
Move Deprecated Transit Method Implementations to New Class

jeffwecan:
After merging in https://github.com/hvac/hvac/pull/303, I decided we could strip some of the transit logic out of the Client class to reduce the amount of repetitive code...

cc: @alexandernst


alexandernst:
Well... It's just a proxy to the same functionality, so it *should* work without any problems, but it's quite a big refactor for a minor release, plus the fact that it's refactoring code that is marked as deprecated anyways and will be removed in a couple of minor releases.

jeffwecan:
Thanks for the feedback @alexandernst, those are totally valid points! I'll think on it some before deciding whether to close this PR or merge it in...

Issue #339, opened 2018-11-20
sys.read_health_status() not working as expected


qeqar:
I have two problems with sys.read_health_status()

both seems to have todo with this code.
https://github.com/hvac/hvac/blob/6d7b5f5913f0fd3f024c111b8e228e2f66aefd78/hvac/adapters.py#L259-L266

When checking health on a standby the response code is `429`  but this code is throwing an `exceptions.RateLimitExceeded` even if in this case `429` is ok.

```  File "...../hvac/utils.py", line 37, in raise_for_error
    raise exceptions.RateLimitExceeded(message, errors=errors)
hvac.exceptions.RateLimitExceeded: {"initialized":true,"sealed":false,"standby":true,"performance_standby":false,"replication_performance_mode":"disabled","replication_dr_mode":"disabled","server_time_utc":1542718003,"version":"0.11.4","cluster_name":"vault-cluster-331bacf9","cluster_id":"e9b9ee72-ec96-5cbf-7ae8-95284eb8328c"}
```

the second problem is a when using `HEAD` and not `GET` we get a json decode error because we don't have a response.

curl -i output:
```% curl https://example.de:8200/v1/sys/health -I
HTTP/2 429 
cache-control: no-store
content-type: application/json
date: Tue, 20 Nov 2018 13:30:40 GMT
```

it is ok but the api throws
```  File "/.....t/lib/python3.7/site-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib64/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

I can prepare a pull request where i check if the response i empty and then do nothing. 
but i have no idea for the rate limit error?

Mark

qeqar:
ok a simple `and len(response.content) > 0 ` in line https://github.com/hvac/hvac/blob/6d7b5f5913f0fd3f024c111b8e228e2f66aefd78/hvac/adapters.py#L261
schould help with the JSON error if no content.

jeffwecan:
[...]

PR #340, opened 2018-11-20
fix json error when json header with no content

qeqar:
Fixes one part of #339


jeffwecan:
Thanks @qeqar! This looks like it would do the trick for that specific issue. I'd like to deliberate on this change a bit before merging this PR however.  I mainly want to explore if there isn't a more holistic fix that could improve how we're handling "error" responses from Vault generally...

PR #344, opened 2018-12-03
Bump Vault Versions - Vault v1.0.0

jeffwecan:
Bumping the versions of Vault we test again now that v1.0.0 is available. (Also fixing an oversight where Python 3.6 tests where using the py27 "TOXENV".)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,6 @@ sudo: false
 language: python
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


jeffwecan:
Note: This has a few failing test cases for v1.0.0. So those will need addressing before this is ready for a proper review...

jeffwecan:
Alrighty, hand to do some questionable conditionals to get all the failing test cases in v1.0.0 happy, but they should do the trick for now...

PR #346, opened 2018-12-06
Fix identity group conditionals

jeffwecan:
Turns out a bug from the first pass on the Identity class caused us to never set the `member_entity_ids` or `member_group_ids`. The original intention was to drop these params from the request sent to Vault for "external" group types as that results in an exception. Regression test cases have been added to catch the bug and a new `validate_member_id_params_for_group_type()` static method with the corrected logic has been added to the Identity class.

General points to note:

* There is some code duplication in the test cases that I am personally totally okay with. However I welcome any suggestions on how to DRY that bit up if y'all would like to see that addressed.
* This is the first time we've added a logger call in one of the API method classes. So it is worth pondering the implications that may entail. For context, the `logger.warning()` call seemed like a good fit between hiding the occurrence (i.e., that params were being dropped to allow the rest of the request to succeed) and continuing to allow some sort of exception to be throw to serve as the notification to the user.

### DIFF
--- a/hvac/api/secrets_engines/identity.py
+++ b/hvac/api/secrets_engines/identity.py
@@ -532,12 +532,19 @@ class Identity(VaultApiBase):
         }
         if group_id is not None:
             params['id'] = group_id
-        if group_type == 'external' and member_entity_ids is not None:
-            # InvalidRequest: member entities can't be set manually for external groups
+
+        if group_type == 'external':
+            if member_entity_ids is not None:
+                logger.warning("InvalidRequest: member entities can't be set manually for external groupsl ignoring member_entity_ids argument.")
+        else:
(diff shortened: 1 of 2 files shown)


lucascollino:
Is it possible the issue still remains for
hvac.api.secrets_engines.Identity.create_or_update_group_by_name()
I was using it but got the error:
`hvac.exceptions.InvalidRequest: member entities can't be set manually for external groups`

Using the same config with 
`hvac.api.secrets_engines.Identity.create_or_update_group()`
works like intended

config:
```
client.secrets.identity.create_or_update_group(
        name=app,
        group_type="external",
        policies="{0}".format(app),
    )
```

hvac: 0.7.2

PR #347, opened 2018-12-07
Fix For read_health_status() Exception Handling

### LINKED ISSUE #339: sys.read_health_status() not working as expected

jeffwecan:
Related PR: https://github.com/hvac/hvac/pull/340 (cc: @qeqar)

Closes #339 

Since the `/v1/sys/health` Vault API route allows the requester to set arbitrary response codes for various Vault states, it doesn't interact well with the exception handling we have for standard Vault response codes (i.e.: [`hvac.utils.raise_for_error()`](https://github.com/hvac/hvac/blob/a609d8fdbc1f3db5126f44fd98964b869c9b850c/hvac/utils.py#L13-L45)). So this PR adds a `raise_exception` parameter to our main requests adapter method that allows us to bypass that functionality for the health status endpoint.

Note: This PR also adds in a consul dependency for a subset of our integration test cases (in order to test the responses of the implicated method with a Vault HA cluster set up). It has been suggested that it makes more sense for these cases to be covered by unit testing. However, I feel good about going the integration test route since there are a lot of parameters at play with this method so overall it feels like the quicker solution to implement. It also doesn't necessarily require that consul dependency to be met for all contributors as the related test cases are skip when consul is not available.


requirements-dev.txt
```diff
@@ -1,5 +1,6 @@
 codecov
 coverage
+ipaddress>=1.0.22
```
jeffwecan: See the message on commit https://github.com/hvac/hvac/pull/347/commits/d0c2961a0d143e3fc7dd4574c8a720c341e96e49 for additional context on this addition...

tests/integration_tests/api/system_backend/test_health.py
```diff
@@ -2,30 +2,78 @@
...
+            ha_required=True,
+        ),
+        param(
+            'sealed standby node',
+            vault_addr='https://127.0.0.1:8199',
+            method='GET',
+            expected_status_code=503,
+            seal_first=True,
+            ha_required=True,
+        ),
```
jeffwecan: For posterity, here is the output from these regression test without the fix being added here:

```
ERROR: Test the Health system backend class's "read_health_status" method [with label='unsealed standby node', method='HEAD', vault_addr='https://127.0.0.1:8199', expected_status_code=429, seal_first=False, ha_required=True].
  File "/Users/jeffrey.hogan/.virtualenvs/hvac/lib/python3.7/site-packages/parameterized/parameterized.py", line 392, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/Users/jeffrey.hogan/workspace/hvac/tests/integration_tests/api/system_backend/test_health.py", line 70, in test_read_health_status
    method=method,
  File "/Users/jeffrey.hogan/workspace/hvac/hvac/api/system_backend/health.py", line 59, in read_health_status
    url=api_path,
  File "/Users/jeffrey.hogan/workspace/hvac/hvac/adapters.py", line 158, in head
    return self.request('head', url, **kwargs)
  File "/Users/jeffrey.hogan/workspace/hvac/hvac/adapters.py", line 266, in request
    errors = response.json().get('errors')
  File "/Users/jeffrey.hogan/.virtualenvs/hvac/lib/python3.7/site-packages/requests/models.py", line 897, in json
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 348, in loads
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py", line 337, in decode
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

ERROR: Test the Health system backend class's "read_health_status" method [with label='sealed standby node', method='GET', vault_addr='https://127.0.0.1:8199', expected_status_code=503, seal_first=True, ha_required=True].
  File "/Users/jeffrey.hogan/.virtualenvs/hvac/lib/python3.7/site-packages/parameterized/parameterized.py", line 392, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/Users/jeffrey.hogan/workspace/hvac/tests/integration_tests/api/system_backend/test_health.py", line 70, in test_read_health_status
    method=method,
  File "/Users/jeffrey.hogan/workspace/hvac/hvac/api/system_backend/health.py", line 66, in read_health_status
    json=params,
  File "/Users/jeffrey.hogan/workspace/hvac/hvac/adapters.py", line 93, in get
    return self.request('get', url, **kwargs)
  File "/Users/jeffrey.hogan/workspace/hvac/hvac/adapters.py", line 269, in request
    utils.raise_for_error(response.status_code, text, errors=errors)
  File "/Users/jeffrey.hogan/workspace/hvac/hvac/utils.py", line 43, in raise_for_error
    raise exceptions.VaultDown(message, errors=errors)
hvac.exceptions.VaultDown: {"initialized":true,"sealed":true,"standby":true,"performance_standby":false,"replication_performance_mode":"unknown","replication_dr_mode":"unknown","server_time_utc":1544189688,"version":"1.0.0"}
```

@qeqar: Are there any additional test cases we should add here to capture the issues you have outlined in #339?

jeffwecan:
Additional note: This change in behavior could potentially break usage for folks that were counting on a hvac.exceptions class being thrown when read_health_status() is called. So if we merge this PR, that needs be called out explicitly in the next hvac release's change log updates.

tests/integration_tests/api/system_backend/test_health.py
jeffwecan: After re-reading the issue, I realized we should go ahead and check both GET and HEAD methods for both types of scenarios.... 16ef641

qeqar: Sorry i missed the question :-)

ok we test one in 4xx one in 5xx, i am not sure if we should do positive test 200 too?

the unsealed standby with head  should cover my finding in #339

qeqar:
i am not familiar with unittest in python, but how do you make sure that u test against a `unsealed standby node`?

jeffwecan:
In this case, sealing is handled by restarting the test Vault cluster before running the test cases covering that mode. The standby node is determined by the port number specified on the second Vault server process started by the integration tests. Though with you question, and in thinking about it a bit more, I realize this PR should better document how that arrangement is set up and the test case should explicitly discover the port of the standby node rather than relying on the implicit ordering of the two HA Vault nodes to assume which one is active and which is not... Will push up a couple changes to fill in those gaps before merging in these changes...

Issue #51, opened 2016-05-11
Write fails when path contains backslash chars


pschmitt:
Hi,

the following script fails:

```python
import hvac
import os

client = hvac.Client(
    url=os.environ.get('VAULT_ADDR'),
    token=os.environ.get('VAULT_TOKEN'),
    verify=os.environ.get('VAULT_SKIP_VERIFY', False)
)

client.write(
    'keyring/http://some.url/sub/entry',
    foo='bar'
)
```

Output:

```
Traceback (most recent call last):
  File "hvac_test.py", line 12, in <module>
    foo='bar'
  File "/home/pschmitt/.local/share/virtualenvs/vault-keyring-import/lib/python2.7/site-packages/hvac/v1/__init__.py", line 45, in write
    response = self._put('/v1/{0}'.format(path), json=kwargs)
  File "/home/pschmitt/.local/share/virtualenvs/vault-keyring-import/lib/python2.7/site-packages/hvac/v1/__init__.py", line 576, in _put
    return self.__request('put', url, **kwargs)
  File "/home/pschmitt/.local/share/virtualenvs/vault-keyring-import/lib/python2.7/site-packages/hvac/v1/__init__.py", line 606, in __request
    raise exceptions.InvalidRequest(errors=errors)
hvac.exceptions.InvalidRequest: missing data fields
```

I don't see why this fails since the request `kwargs` are correctly set to `{'json': {'foo': 'bar'}}`

Any ideas?

EDIT: 

```bash
vault write "keyring/http://some.url/sub/entry" notes=okay                                                                                                 
Success! Data written to: keyring/http://some.url/sub/entr
```

This works just fine

TerryHowe:
I don't know, but I would guess the cli is doing something similar to [urlencoding](https://docs.python.org/2/library/urllib.html#urllib.urlencode) and urldecoding the key for you.  You would need to do that.  Having a key to a REST API, you'd have to.

ianunruh:
So, I'm able to reproduce this on my end, but I'm not sure how to fix it. In the native Vault client, there's no special escaping done on paths.

https://github.com/hashicorp/vault/blob/v0.5.2/api/logical.go#L47

Even when I escape that part of the URL with `urllib.quote` or `urllib.quote_plus`, I still get the `missing data fields` error.

v6:
// , @ianunruh, are you saying that this could be an issue with the native Vault client?

ianunruh:
[...]

PR #348, opened 2018-12-07
Fix Request Redirection Handling

### LINKED ISSUE #343: URL redirect partially broken in v0.6.1
In v0.6.0  v1/__init__.py there was code:
```
  # NOTE(ianunruh): workaround for https://github.com/ianunruh/hvac/issues/51
        while response.is_redirect and self.allow_redirects:
            url = urljoin(self._url, response.headers['Location'])
            response = self.session.request(method, url, headers=headers,
                                            allow_redirects=False, **_kwargs)
```
If a redirect occurred then urlparse:urljoin was used.  In v0.6.1 urljoin was switched to an internally written function, that just joins strings with a slash. 
```
 @staticmethod
    def urljoin(*args):
        """
        Joins given arguments into a url. Trailing and leading slashes are
        stripped for each argument.
        """

        return '/'.join(map(lambda x: str(x).strip('/'), args))

```
We have a vault cluster where when you hit a non-master node, you get a fully qualified URL.  The new code then redirects to something like https://vault.my.org:8200/https://10.100.10.1:8200/.  The urljoin from url parse has more advanced parsing that looks at the string being appended and if the second string starts with https:// is just uses the second string.
The new code would work for a redirect to a page on same site, but fails on redirects to a new site.

This change appears intentional, but breaks us.  I don't know what this commit fixed:
https://github.com/hvac/hvac/commit/3f08b5f8d599fcf65aa2a4018337dedcac7f3429

jeffwecan:
Closes #343. (cc: @stvdilln)
We've had three general issues with URL parsing and/or redirects in the past:

* https://github.com/hvac/hvac/issues/51 - Vault redirect / 301 response when a path contains `//` to a version of the path with a single `/`
* https://github.com/hvac/hvac/issues/205 - urllib's `urljoin()` method dropping paths from the base URL
* https://github.com/hvac/hvac/issues/343 - custom redirect logic and urljoin methods from the above two issues is causing Location headers in the response from Vault being incorrectly handled.

This PR ensures that all three issues are captured in test cases and reverts back to using the requests module's built-in redirect handling logic. The latter bit is accomplished by converting occurrences of `//` in a given path to `/` ourselves so that we don't end up with a 301 response back from Vault when it makes that change.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -247,14 +247,13 @@ class Request(Adapter):
         _kwargs = self._kwargs.copy()
         _kwargs.update(kwargs)
 
-        response = self.session.request(method, url, headers=headers,
-                                        allow_redirects=False, **_kwargs)
-
-        # NOTE(ianunruh): workaround for https://github.com/hvac/hvac/issues/51
-        while response.is_redirect and self.allow_redirects:
-            url = self.urljoin(self.base_uri, response.headers['Location'])
-            response = self.session.request(method, url, headers=headers,
-                                            allow_redirects=False, **_kwargs)
+        response = self.session.request(
+            method=method,
+            url=url,
+            headers=headers,
+            allow_redirects=self.allow_redirects,
+            **_kwargs,
+        )
 
         if 400 <= response.status_code < 600:
             text = errors = None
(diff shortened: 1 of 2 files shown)


dbwpe:
would be nice to verify that both request mocks you're registering are actually hit

PR #368, opened 2018-12-26
Use pip-compile For All Requirements

jeffwecan:
* Updating Contributing.md with a small blurb about using pip-compile
* Adding Makefile targets to run pip-compile for the various requirements files in this project
* ~Breaking out the pyhcl / hvac[parser] requirements to their own file so we can dynamically load them when running setup.py~. Turns out dynamically loading requirements in setup.py is quite a silly thing to do.

Overall, these changes / additions will ideally make it supremely easy to ensure we're keeping requirements up to date.


dotlambda:
Was there a reason to require `requests>=2.21.0`?

jeffwecan:
Honestly it was mostly an arbitrary decision to reset this modules version pinning to a current version. If that choice is causing issues for you or any other consumers of hvac, I would be glad to see an issue filed and we can see about tuning that version pinning down to the minimum required requests release :+1:.

PR #375, opened 2019-01-08
Fix delete_role_secret_id_accessor method

### LINKED ISSUE #297: Possible bug with `delete_role_secret_id_accessor`
The https://github.com/hvac/hvac/blob/master/hvac/v1/__init__.py#L2064 `delete_role_secret_id_accessor` method does not look like the documented https://www.vaultproject.io/api/auth/approle/#destroy-approle-secret-id-accessor method in the API.

1) the docstring has `DELETE` instead of `POST`
2) the docstring has `/secret-id/` instead of `/secret-id-accessor`
3) both the code, and docstring do not have the `/destroy` suffix as per the API

Most probably it should be similar to `delete_role_secret_id` https://github.com/hvac/hvac/blob/master/hvac/v1/__init__.py#L2046 where `POST` and `/destroy` are used for `/secret-id/` endpoints, it also has the `{ "secret_id": "" }` param, also like the API documentation shows for the accessor destruction.

frgaudet:
The current `delete_role_secret_id_accessor` method doesn't work. This patch fix this.

Resolves #297

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -1369,7 +1369,7 @@ class Client(object):
         return self._adapter.post(url, json=params)
 
     def delete_role_secret_id_accessor(self, role_name, secret_id_accessor, mount_point='approle'):
-        """DELETE /auth/<mount_point>/role/<role name>/secret-id/<secret_id_accessor>
+        """POST /auth/<mount_point>/role/<role name>/secret-id-accessor/destroy
 
         :param role_name:
         :type role_name:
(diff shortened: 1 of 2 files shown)


jeffwecan:
@frgaudet: Thanks for the contribution! At a glance this change seems reasonable. I just want to verify that the change proposed here would work for _all_ hvac's support Vault versions. We typically lean on the travis-ci builds / Vault integration cases to cover that side of things, but since this method only has a unit test we can't be as sure. So, in other words, so manual validation might be worthwhile and/or we may want to see if we can find a way to retrieve a more "integrate-y" version of this test case...

frgaudet:
I guess the API changed at some point, so I guess some old version should break. I currently use `0.11.4` vault version which is fine with this fix.

Issue #376, opened 2019-01-09
Cannot set length of random bytes in transit.generate_random_bytes()


kserrano:
No matter what the value 'n_bytes' is, the random bytes are always 32 bytes, which is the default value.
Assuming you have a valid token :
`client.secrets.transit.generate_random_bytes(n_bytes=length,output_format='hex')['data']['random_bytes']`
returns 32 bytes, independently of the value 'n_bytes'

jeffwecan:
Closed by https://github.com/hvac/hvac/pull/377?

PR #377, opened 2019-01-09
#376 : fix length in transit.generate_random_bytes()

kserrano:
The parameter in the HTTP request was incorrect according to the
latest vault API
https://www.vaultproject.io/api/secret/transit/index.html

### DIFF
--- a/hvac/api/secrets_engines/transit.py
+++ b/hvac/api/secrets_engines/transit.py
@@ -500,7 +500,7 @@ class Transit(VaultApiBase):
         :rtype: requests.Response
         """
         params = {
-            'n_bytes': n_bytes,
+            'bytes': n_bytes,
             'format': output_format,
         }
         api_path = '/v1/{mount_point}/random'.format(mount_point=mount_point)


jeffwecan:
I have a feeling the parameter name may have changed recently then...

@kserrano: Would you mind trying to add some regression tests for the issue you lodged as part of this PR? Having such test(s) would allow our travis-ci builds to verify this change is safe for the various versions of Vault the hvac module supports.

kserrano:
Alright, will do. My change is breaking compatibility with vault versions using the old parameter name 'n_bytes'. I don't know when they changed it.

jeffwecan:
Yup, that is a common issue with this module + maintaining support for the generally recent releases of Vault. I'll see about taking a go at adding some test cases that confirm which versions we would be concerned with in this context as a first step. Assuming the change from `n_bytes` to `bytes` happened within our support span of Vault releases, we might have to see about making accommodations (one way or another) to support both forms of the parameters until hvac drops support for the Vault versions that would require a parameter name of `n_bytes`...

PR #396, opened 2019-02-22
Update Test Runs With Latest Vault Versions (1.0.3 & 1.1.0)

singuliere:


jeffwecan:
@singuliere Thanks for the PR! Historically, I've been limiting our travis-ci runs to one Vault version per minor release in the interest of limiting test / build run times. I.e., how would you feel about dropping the entries for Vault v1.0.1 and v1.0.2; leaving just v1.0.3?

(One notable exception is Vault v0.11.0 that introduced some backwards incompatible changes that were subsequently backed out in following releases.)

singuliere:
> how would you feel about dropping the entries for Vault v1.0.1 and v1.0.2; leaving just v1.0.3?

Sounds good, changed & repushed :-)

Side question: do you know of a hvac based vault cli? Something like [vault-cli](https://github.com/peopledoc/vault-cli) but not implementing their own HTTP API layer and using hvac instead.

singuliere:
I think the [1.0.3 test error](https://travis-ci.org/hvac/hvac/jobs/497050644#L252) is real.

jeffwecan:
I haven't run across a python based Vault cli tool but I've considered putting together at some point! Would probably end up adding such a thing to a different repository under the hvac organization if it ever came up...

singuliere:
> Would probably end up adding such a thing to a different repository under the hvac organization if it ever came up...

I'm up for it. The latest work I've done is the [vault-keepass-import](https://vault-keepass-import.readthedocs.io/) script. Would you consider creating a repository so I can get started? It would be great to work with you (even if only as a reviewer for now ;-).

jeffwecan:
Deal! I just spun up https://github.com/hvac/hvac-cli and added you as a collaborator. Feel free to suggest a different name and/or hit me up with any other changes you'd like assistance with.

jeffwecan:
As to the substance of this PR, looks like we're being caught by this change from upstream v1.0.2:

> secret/aws: Role now returns credential_type instead of credential_types to match role input. If a legacy role that can supply more than one credential type, they will be concatenated with a ,.

https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#102-january-15th-2019

We don't have a solid pattern yet within hvac on how to handle this sort of functional change while still remaining support for older versions of Vault. I'll take a look at it sometime soon unless you want to give it a go. 

For context, in the past, I've debated adding conditional logic for such things that would key on the target Vault cluster's version. (e.g., `if vault_version >= '1.0.2': key = 'credential_type'` or what-have-you). However I haven't implemented anything along those lines outside of our integration tests thus far. Ultimately it comes down to a debate between designing hvac as a "dumb" API client that simply fulfills requests and returns responses directly from Vault versus adding extra quality-of-life features that tweak requests and/or responses in appropriate cases. The latter comes with the risk of increasing the amount of custom logic we have to support...

jeffwecan:
@singuliere: Hope ya don't mind, I rebased your branch on top of develop and pushed up some changes to enable test runs on Vault v1.1.0 (along with some test case changes to get things working with some of those version's changes). I'll merge this in ahead of cutting a new release today pending all the test runs passing. 👍

singuliere:
Thanks for completing this!

PR #400, opened 2019-03-06
Fix logging no-handlers warning

paulcaskey:
If a user requests a URI starting with a slash like "/some/path/...", it gets prepended with "/v1/" in the hvac Client class, resulting in a final URL here of "/v1//some/path/..." We then try to log a warning about this (see line 239). However, the python "logging" module requires at least some basic setup before it can be used, so the call on line 242 to logger.warning() actually throws its own non-critical error: 

   No handlers could be found for logger "hvac.adapters"

This change adds the easiest fix, which is to call logging.basicConfig() before instantiating our logger at the top of this file. 

P.S. As a shortcut, logging.warning() calls basicConfg() for you, if you haven't already started with getLogger(). So, confusingly, another fix here would be to remove this entirely:
    logger = logging.getLogger(__name__)
and call this:
    logging.warning('Replacing double-slashes...')
instead of this:
    logger.warning('Replacing double-slashes...')

I feel my fix here is cleaner.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -11,6 +11,7 @@ import requests.exceptions
 
 from hvac import utils
 
+logging.basicConfig()
 logger = logging.getLogger(__name__)
 
 DEFAULT_BASE_URI = 'http://localhost:8200'


jeffwecan:
I'd rather not set up basic config bits within hvac so that it's easier for users of the module to choose how the wish to handle logging. E.g., I'd rather avoid this type of deal: https://stackoverflow.com/a/12158233

I'd be interested to hear a bit more background on why the emitted "No handlers could be found for logger..." needs to be solved for.

paulcaskey:
It just seems odd, to me, that you instantiating a logger like this without any config, and then try to use it yourself, which throws an error. If I use this library, starting my URI with a slash (which is generally common and valid), I get this very unexpected behavior. I never see the warning of my URI starting with a slash, so I don't know how to fix my code to work with yours better and avoid the attempted log warning. 

Instead I get "No handlers could be found..." which is vague and gives me no clue what I'm doing wrong. I guess you're implying that all users of your tool should be totally familiar with the python logging library, recognize this error, and add their own logging config for logging("hvac.adapters") before calling hvac.Request.request(). 

I see your point that you don't want to clobber an end user's logging config with basicConfig(). In that case, I'd suggest my other idea: call logging.warning() instead of logger.warning() which I believe solves all of this. 

Or just don't log this case at all. It's arguably not even worth a warning, especially since you barely use logging anywhere else in this tool (apart from tests). 

You're the one prepending "/v1/" instead of "/v1" creating the double slash in the first place - which makes all this even more confusing. 

You could just remove all this:
```
        if '//' in url:
            # Vault CLI treats a double forward slash ('//') as a single forward slash for a given path.
            # To avoid issues with the requests module's redirection logic, we perform the same translation here.
            logger.warning('Replacing double-slashes ("//") in path with single slash ("/") to avoid Vault redirect response.')
            url = url.replace('//', '/')
```

And do it in your urljoin() function:
```
    return '/'.join(map(lambda x: str(x).strip('/'), args)).replace('//', '/')
```
That one liner is already Perl-worthy so, what the heck, make it worse, and move on. ;-)

jeffwecan:
@paulcaskey That all makes enough sense to me, thanks for providing that additional input! At this point, I'm figuring we can tweak this PR either a) drop the logger variable assignment and resulting call all together or b) just dropping the logger bit and call `logging.warning()` directly.

jeffwecan:
Will handle this in another PR since I'm unable to push to your fork's branch.

PR #403, opened 2019-03-12
AWS: ensure request parameters are sent with generate_credentials

steved:
The JSON body is not read on GET requests, we either need to POST or
pass the arguments as query parameters.

### DIFF
--- a/tests/integration_tests/api/secrets_engines/test_aws.py
+++ b/tests/integration_tests/api/secrets_engines/test_aws.py
@@ -254,7 +254,8 @@ class TestAws(HvacIntegrationTestCase, TestCase):
                     d1=json.loads(read_role_response['data']['policy']),
                     d2=self.TEST_POLICY_DOCUMENT,
                 )
-            elif 'credential_types' in read_role_response['data']:
+            # https://github.com/hashicorp/vault/commit/2dcd0aed2a242f53dae03318b4d68693f7d92b81
+            elif vault_version_lt('1.0.2'):
                 self.assertEqual(
                     first=read_role_response['data']['credential_types'],
                     second=['iam_user'],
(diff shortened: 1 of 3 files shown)


jeffwecan:
The documentation doesn't explicitly call out POST as an accepted method: https://www.vaultproject.io/api/secret/aws/index.html#generate-credentials.


I haven't dug into much, but given the premise and the current upstream documentation, shouldn't we just change the `json` keyword argument (that get passed through to the requests module) be changed to `params` instead of changing the request method to POST (assuming the upstream documentation is up-to-date)?

steved:
Yeah, unfortunately the docs aren't super consistent as the AWS secret engine documentation uses `vault write`: https://www.vaultproject.io/docs/secrets/aws/index.html#sts-assumerole.

Either way works for me though, so I'm happy to adhere to their API documentation.

tests/integration_tests/api/secrets_engines/test_aws.py
```diff
@@ -254,11 +254,16 @@ def test_read_role(self, label, configure_first=True, raises=None, exception_mes
                     d1=json.loads(read_role_response['data']['policy']),
                     d2=self.TEST_POLICY_DOCUMENT,
                 )
-            else:
+            elif 'credential_types' in read_role_response['data']:
```
jeffwecan: I don't have any issue with working around the `credential_type` / `credential_types` change in this secrets engine in this manner, but another option would be to go off the Vault version being tested against. E.g.:

```python
key_name = 'credential_types' if vault_version_lt('1.0.2') else 'credential_type'
self.assertEqual(
    first=read_role_response['data'][key_name],
    second=['iam_user'],
)
```

I like that route because it makes things a bit more explicit. What do you think?

jeffwecan:
Also for what it is worth, I would be fine with the originally proposed change to a POST method or with the current changes (now that you've done my due diligence in verifying a POST should work just as well 😄).

tests/integration_tests/api/secrets_engines/test_aws.py
steved: Sure! Changed and added a link to the corresponding vault commit.

steved:
GET works for me. I sort of understand the semantics of using a POST for this endpoint (creating credentials, ...), but for some reason it still makes more sense as a read.

Issue #294, opened 2018-10-16
Test cases should not be installed by setup.py


onitake:
`setup.py` installs `hvac/tests` together with the rest of hvac.

I don't think that makes a lot of sense, please consider excluding them, perhaps by adding the `exclude=` option to the `find_packages()` call.

jeffwecan:
@onitake: I agree 100%. Last I checked, there wasn't any clear standard (PEP's or otherwise) or whether test files should be included in the release artifacts or not. However since we're both leaning towards not including the tests I think its appropriate to go ahead and make that change.

kelleyblackmore:
What about just having a separate repo for the tests? So if you have some pipeline you can call that repo to run the tests. I'm willing to help out with this effort.

jeffwecan:
My current intention is just to move tests outside the "hvac" subdirectory targeted by setuptools as thats a pattern I personal enjoy. I don't think I want the added complexity of cross-dependencies with some other test repository. Aside from that I'm absolutely open to PRs with suggested solutions! :D

onitake:
@jeffwecan That sounds absolutely reasonable to me. A separate repo would be more hassle than worth, and it's common practice to keep unit tests and source code close together.

Perhaps these unit tests can be unified with the tox tests that are already in the repo root?

jeffwecan:
@kellyblackmore Thinking about it more, there is an outstanding issue that touches on your comment here: https://github.com/hvac/hvac/issues/99. I.e., while I want to keep the test cases for hvac within this repository generally, moving the bits orchestrating the Vault server parts of the integration test could conceivably be broken out to their our module.

onitake:
Is the mocking code big enough to warrant its own repository?
If not, I don't think that's the right approach, as it would prevent executing unit tests directly from the source tree, without checking out dependencies first.

jeffwecan:
Agreed! My comment was specifically targeted at any *integration* test components that aren't directly related to hvac (e.g., starting up a dev Vault server and tracking its state across integration test cases). I'm 100% in favor of unit tests being runable with the bare minimum of set up required across any environment folks may choose to run them in. 👍

kelleyblackmore:
Good job @jeffwecan

onitake:
Thanks, @jeffwecan !

PR #418, opened 2019-04-01
Actually do not install tests

eltoder:
In #294 `tests` directory was moved out from `hvac` with the intention to not
install tests into site-packages. However, find_packages() still finds and
installs it, this time as a top-level package. Pass it into exclude to really
skip it.

Same thing for `docs`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: Implementation :: CPython',
     ],
-    packages=find_packages(),
+    packages=find_packages(exclude=['docs*', 'tests*']),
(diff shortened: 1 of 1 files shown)


jeffwecan:
🤦‍♂️ I totally whiffed on that one and it now makes a bunch of sense. I ended up having to update https://github.com/hvac/hvac/blob/develop/MANIFEST.in to keep the tests out of the `hvac` package after the change you're proposing didin't seem to be sufficient without realizing that _both_ changes were (presumably) required.

eltoder:
I *think* you could have used `exclude=['hvac.tests*']` to exclude it from installation without moving it out, but having tests separately feels cleaner.

PR #430, opened 2019-04-24
* Generate credentials updated to a post request.

JadeHayes:
* Myself and https://github.com/hvac/hvac/pull/403/files received an error when passing in TTL.   GET requests do not recognize json, so the passed in TTL always defaulted to `1h`.

* This PR uses a POST request to call the endpoint.


jeffwecan:
Thanks @JadeHayes!

Issue #451, opened 2019-05-28
azure auth params missing and incorrect names


drewmullen:
missing param option:

- `num_uses`

the function takes params named incorrectly:

- `bound_location` -> `bound_locations`
- `bound_resource_group_names` -> `bound_resource_groups`

these issues stem from [hashi doc issues](https://github.com/hashicorp/vault/issues/6793). opening the ticket for awareness and documentation, will try to get to it this week

PR #452, opened 2019-05-28
issue-451 fix azure auth params

drewmullen:
fixes for https://github.com/hvac/hvac/issues/451

include new param: `num_uses`

change param names:

- `bound_location` -> `bound_locations`
- `bound_resource_group_names` -> `bound_resource_groups`


drewmullen:
@jeffwecan ik you just did a release this weekend. typically are they monthly cadence?

jeffwecan:
Aye, I have been shooting for a one release per month cadence over the past year or so. However, historically it has been 1-2ish months in reality. Releases normally get made when someone asks about an upcoming one enough (and/or when _I_ have something I'd like release hah 👼). That said, I've been coming around to the idea that quicker incremental releases done as as anything worth shipping is merged is probably the happier path long term. Which is to say, if shipping this change today would help you out, I'd be down to try and make that happen. Otherwise I'll get to it by this time next week.

drewmullen:
youre a kind soul, @jeffwecan ! dont rush yourself for today. however, I would **love** it if you could shoot for sometime next week!

jeffwecan:
Sounds good! I wouldn't mind lending a hand myself on filling in the documentation for _that_ secrets engine if we don't get any other takers before then (re: https://github.com/hvac/hvac/issues/453).

Issue #454, opened 2019-05-29
database secrets engine list_connections error


drewmullen:
regarding: https://github.com/hvac/hvac/pull/431

```
import hvac

def main():
    client = get_client()
    result = client.secrets.database.list_connections()
    print(result)
```
ERROR:
```
Traceback (most recent call last):
  File "db_config.py", line 29, in <module>
    main()
  File "db_config.py", line 10, in main
    result = client.secrets.database.list_connections()
  File "/Users/dmullen/venv/lib/python3.7/site-packages/hvac/api/vault_api_category.py", line 42, in __getattr__
    raise NotImplementedError('"%s" auth method class not currently implemented.')
```

```
$ pip list | grep hvac
hvac                       0.9.1
```

```
$ vault list database/config
Keys
----
test
```

my best guess is that this has to do with `list` possibly not being implemented yet since the error looks similar though im not sure why it mentions `auth method`... will continue to investigate. any thoughts @Dudesons ?
https://github.com/hvac/hvac/blob/9ad27d06658c1e5a8284c7ed02e3c97e8e93ae53/hvac/api/secrets_engines/database.py#L87

https://github.com/hvac/hvac/blob/9ad27d06658c1e5a8284c7ed02e3c97e8e93ae53/hvac/api/vault_api_category.py#L84

drewmullen:
so i noticed that i get the error no matter what function i call in `secrets.database`, even if the function doesnt exist (if its spelled wrong for example). not sure if this is local to my set up or if there's something wrong with how auth methods are called with the module

Dudesons:
@drewmillen I was busy today, I'll try to check that tomorrow

drewmullen:
im pretty sure this has to do with it
https://github.com/hvac/hvac/blob/129d1c036b9107e34fe755ef4e8fa3d919d4f120/hvac/api/secrets_engines/__init__.py#L44

drewmullen:
yep. definitely it. pushing a fix now

Issue #434, opened 2019-05-14
database secrets engine


drewmullen:
is anyone working on a specific [database secrets engine](https://www.vaultproject.io/docs/secrets/databases/index.html) implementation for HVAC? also, in light of DRY, is there any way we can use existing lib modules for configuring a database secrets engine? if not, if i were to write the database engine feature would we want to use current methodology of separating the secret engines to hvac lib endpoints or something similar to below?

example:

**azure secrets engine config**
`v1/<engine_type>/config` <-- pass it a json object
`v1/database/config` <-- pass it a json object

**create a role**
`v1/<engine_type>/role/<role_name>` <-- pass it a json object
`v1/database/role/<role_name>` <-- pass it a json object

jeffwecan:
One was added recently in https://github.com/hvac/hvac/pull/431 but is not yet released. I'll start working on getting a 0.8.3 release cut today since its overdue anywho. Leaving this issue open since there is certainly a lot of room to DRY up similar code across secrets engines, etc.

drewmullen:
another example, the `azure` secrets engine doesnt have a `read_role` function, but using `https://hvac.readthedocs.io/en/stable/usage/secrets_engines/aws.html#read-role` works great lol

`current = client.secrets.aws.read_role(name=name, mount_point=azure)`

PR #455, opened 2019-05-29
include db engine as implemented class

drewmullen:
database secrets engine wasnt enabled though the code was included. probably a good example for why we need unit tests :)

also - @jeffwecan may want to investigate why the error said `'"%s" auth method class` instead of `database secret engine class`

fix for https://github.com/hvac/hvac/issues/454

### DIFF
--- a/hvac/api/secrets_engines/__init__.py
+++ b/hvac/api/secrets_engines/__init__.py
@@ -8,6 +8,7 @@ from hvac.api.secrets_engines.pki import Pki
 from hvac.api.secrets_engines.kv_v1 import KvV1
 from hvac.api.secrets_engines.kv_v2 import KvV2
 from hvac.api.secrets_engines.transit import Transit
+from hvac.api.secrets_engines.database import Database
 from hvac.api.vault_api_category import VaultApiCategory
 
 __all__ = (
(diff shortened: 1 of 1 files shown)


jeffwecan:
Ah good catch! My bad for missing that bit in the initial PR (the fact it was easy enough to miss is more fodder for https://github.com/hvac/hvac/issues/434 imho ;) ). The `'"%s" auth method class` is just a copy/paste carry over that should have been updated to state `secrets engine class`. If you wanna grab that fix here as well I would be down!

drewmullen:
@jeffwecan so if i changed:
https://github.com/hvac/hvac/blob/9ad27d06658c1e5a8284c7ed02e3c97e8e93ae53/hvac/api/vault_api_category.py#L42

wouldnt that also change the error when its searching for all 3 implementation lists ('AuthMethods', 'SecretsEngines', 'SystemBackend',)? i was trying to figure that out yesterday 

the change might be more significant but im still pretty new to inheritance, etc

jeffwecan:
Ah yeah okay, didn't realize that bit was on the base class. We can skip it for now if you want and address that in another PR if you'd like in that case.

drewmullen:
yeah that sounds good. im gonna add a couple more bits to the secrets.database class too. will try to get that in by the weekend

PR #457, opened 2019-05-31
fixes to database secrets engine

drewmullen:
fix `delete_connection` return value. call does not return json object

include optional `mount_point` but default to 'database'

tested:
list_connections
read_connection
configure
delete_connection
create_role
delete_role
list_roles

### DIFF
--- a/hvac/api/secrets_engines/database.py
+++ b/hvac/api/secrets_engines/database.py
@@ -3,8 +3,7 @@
 """Database methods module."""
 from hvac.api.vault_api_base import VaultApiBase
 
-DEFAULT_MOUNT_POINT = "database"
-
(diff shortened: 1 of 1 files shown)


drewmullen:
@jeffwecan alright so afaik these are good to go. ive tested most of the functionality and am using them here: https://github.com/TerryHowe/ansible-modules-hashivault/pull/133

if you have a chance to merge and cut a release next week that would be AWESOME!

jeffwecan:
Sounds good, thanks again for all the contributions! I'll aim to cut a release over the next few days as I have time available.

drewmullen:
Absolutely. I'm looking at your unit tests trying to see if i can figure out a way to make some for this module also... havent messed with mocks much so might take me sometime but... its on the way!

Issue #459, opened 2019-06-03
tune mount config should accept force_no_cache


drewmullen:
https://www.vaultproject.io/api/system/mounts.html#force_no_cache

param missing from tune function

vault sets the var to False by default and returns it from a read_configuration call

Issue #435, opened 2019-05-15
tune_mount_configuration and tune_auth_method "options" parameter not recognized


artis3n:
The `tune_mount_configuration` [api docs](https://hvac.readthedocs.io/en/stable/source/hvac_api_system_backend.html#hvac.api.system_backend.Mount.tune_mount_configuration) list a parameter called `options` that "Specifies mount type specific options that are passed to the backend." The [tune_auth_method](https://hvac.readthedocs.io/en/stable/source/hvac_api_system_backend.html#hvac.api.system_backend.Auth.tune_auth_method) does not support this "options" parameter.

I am using this option to configure K8s auth method options. However, when I run the code I get the following error:
`tune_mount_configuration() got an unexpected keyword argument 'options'`. Sure, enough, in the `sys.tune_mount_configuration` function in hvac, no such `options` parameter exists.

I am attempting to create a Kubernetes auth method mount point. The `enable_auth_method` function does not appear to support passing in auth method configuration options, as per [the docs](https://hvac.readthedocs.io/en/stable/source/hvac_api_system_backend.html#hvac.api.system_backend.Auth.enable_auth_method). So I am doing so through the `tune_mount_configuration` function, which does appear to support this capability.

Here is the code. It is called inside of a custom Ansible module:
```python
def write_mount(module_args, client):
    """
    Enables a Vault auth mount point
    :param module_args: AnsibleModule arguments
    :param client: An instantiated hvac.Client
    :return: Status code of the response and the response body as JSON
    """
    try:
        enable_result = client.sys.enable_auth_method(
            path=module_args["mount_point"].capitalize(),
            method_type=module_args["type"],
        )
        tune_result = client.sys.tune_mount_configuration(
            path=module_args["mount_point"].capitalize(),
            options={
                "kubernetes_host": module_args["cluster_hostname"],
                "kubernetes_ca_cert": decode(module_args["cluster_ca_cert"]),
                "token_reviewer_jwt": module_args["cluster_service_token"],
            }
        )

        return {
            "enabled": {
                "status_code": enable_result.status_code,
                "result": enable_result.json(),
            },
            "tuned": {
                "status_code": tune_result.status_code,
                "result": tune_result.json()
            }
        }
    except (TypeError, hvac.exceptions.VaultError) as e:
        raise AnsibleModuleError(results={
            "msg": "Error while enabling and tuning the auth method %s: %s" % (module_args["mount_point"],
                                                                               to_native(e, nonstring='simplerepr'))
[...]

PR #461, opened 2019-06-03
add param to tune_mount_configuration

drewmullen:
fix for https://github.com/hvac/hvac/issues/459


drewmullen:
@jeffwecan per https://github.com/hvac/hvac/issues/435#issuecomment-498367898 maybe this should be **kwargs instead... how would that affect the 'optional_parameters` var?

akdor1154:
this broke `options` due to a missing comma

PR #475, opened 2019-06-14
Change AWS `generate_credentials` request method to GET

donjar:


hvac/api/secrets_engines/aws.py
```diff
@@ -326,7 +326,7 @@ def generate_credentials(self, name, role_arn=None, ttl="3600s", endpoint='creds
             name=name,
         )
 
-        response = self._adapter.post(
+        response = self._adapter.get(
             url=api_path,
             json=params,
```
jeffwecan: Does the `json` keyword are work with `requests.get()`? If not, this should be updated to `params=params`, no?

donjar: It does work when I tested it, but I think `params` makes more sense. Let me update it.

Issue #479, opened 2019-06-20
Support AWS authentication method


stevefranks:
It looks like AWS isn't fully supported as an authentication method yet within hvac. I'm going to work on adding this. I hope to have a PR ready for review shortly.

https://www.vaultproject.io/api/auth/aws/index.html

jeffwecan:
@stevefranks Sounds good and thanks! If you have any questions at all regarding the best way to implement this contribution, feel free to lodge those queries here or in the hvac gitter channel at: https://gitter.im/hvac/community

jeffwecan:
Also for what it's worth I went down this route way back when and pushed up that work that I was able to complete along those lines now at: https://github.com/jeffwecan/hvac/blob/1573406a231bb31c3779aa2722eeee130fc51247/hvac/api/aws.py

That version of the class doesn't incorporate the current auth class baseclass / directory structure since it was last touched before those elements were finalized. However it may still give you a good starting point for many of the AWS auth methods that remain to be implemented.

PR #482, opened 2019-06-28
Expanded Support for AWS Auth Method

stevefranks:
[Proposal]
This PR adds support for authentication via the AWS auth backend. I have tested this code against my local vault instance. I'll add more documentation once we are in agreement on the design.

References: https://github.com/hvac/hvac/issues/479

cc @jeffwecan

### DIFF
--- a/hvac/api/auth_methods/__init__.py
+++ b/hvac/api/auth_methods/__init__.py
@@ -10,6 +10,7 @@ from hvac.api.auth_methods.ldap import Ldap
 from hvac.api.auth_methods.mfa import Mfa
 from hvac.api.auth_methods.okta import Okta
 from hvac.api.auth_methods.radius import Radius
+from hvac.api.auth_methods.aws import Aws
 from hvac.api.vault_api_category import VaultApiCategory
 from hvac.utils import generate_method_deprecation_message
 
@@ -23,6 +24,7 @@ __all__ = (
     'Mfa',
     'Okta',
     'Radius',
+    'Aws',
 )
 
 
@@ -37,12 +39,12 @@ class AuthMethods(VaultApiCategory):
         Mfa,
         Okta,
         Radius,
+        Aws,
     ]
     unimplemented_classes = [
         'AppId',
         'AppRole',
         'AliCloud',
-        'Aws',
         'Jwt',
         'Cert',
         'Token',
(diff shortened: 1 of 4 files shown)


jeffwecan:
Just as a heads-up, I personally haven't had time to look over this PR but intend to do so as soon as I am able.

hvac/api/auth_methods/aws.py
```diff
@@ -0,0 +1,767 @@
...
+        :type cert_name: str | unicode
+        :param mount_point: The "path" the aws auth method was mounted on
+        :return: The data key from the JSON response of the request.
+        :rtype: dict
+        """
+        api_path = '/v1/auth/{0}/config/certificate/{1}'.format(mount_point, cert_name)
+        response = self._adapter.get(
+            url=api_path,
+        )
+        return response.json().get('data')
```
drewmullen: @jeffwecan we should consider standardizing json returns. personally i like this where the data key is returned directly, however, we have some that doesnt extract it. 

example: https://github.com/hvac/hvac/blob/c013371a43d30b5b54407f3fea2f2fd76bbe5934/hvac/api/auth_methods/ldap.py#L129

hvac/api/auth_methods/__init__.py
```diff
@@ -10,6 +10,7 @@
 from hvac.api.auth_methods.mfa import Mfa
 from hvac.api.auth_methods.okta import Okta
 from hvac.api.auth_methods.radius import Radius
+from hvac.api.auth_methods.aws import Aws
```
drewmullen: nit picky - the other imports are alphabetical

drewmullen:
LGTM! thanks for the PR! i dont have a strong exp with aws auth so ill let @jeffwecan comment b4 approving

jeffwecan:
Also I realize now that I skipped past the "I'll add more documentation once we are in agreement on the design" point you mentioned when opening the PR. However you're still welcome to fill in those gaps at any point in the future (and it would be much appreciated in fact!). The docstrings will at least cover some things since they end up in "autodoc" output at https://hvac.readthedocs.io/en/stable/source/hvac_api_auth_methods.html.

avoidik:
my apologies for bumping this old PR, but how to use iam_login?

my previous code is

```
vault_client = hvac.Client(url=vault_addr)
vault_client.auth_aws_iam(key_id, secret_id, session_token, vault_iam_header, vault_iam_mount, vault_iam_role)
```

avoidik:
nvm, it just

```
vault_client.auth.aws.iam_login()
```

PR #478, opened 2019-06-17
Start Using Enterprise (Trial) Version of Vault For Travis CI Builds

jeffwecan:
Building off @drewmullen's https://github.com/hvac/hvac/pull/474 PR here to start using Vault enterprise during Travis CI builds. This will allow us to add test cases for enterprise only features (e.g., Namespaces, MFA, etc.). While this is technically mergeable as is, I may take a bit more time to see about adding some additional bits. In particular:

- [x] Update documentation to clarify how to run tests with or without the enterprise Vault version locally.
- [x] Add at least one test case for an enterprise only feature that is skipped when testing against Vault OSS to serve as an example / pattern on how to do such things. (Would be nice to keep local test runs available for folks that only care about Vault OSS.)
- [x] Ideally fix whatever bits in the Vault HA (with Consul as the backend) aren't working with the Enterprise version of things. While the related test cases are automatically skipped without HA available in the test environment, and there aren't all that many of them at that, it would be nice to get this sorted so we're not reducing coverage overall.
- [x] Either see if HashiCorp can publish a 0.11.6 version of the Enterprise binary or update the Travis CI build matrix to drop this version.

PR #487, opened 2019-07-07
GCP Auth Test Case Updates For Changes in Vault v1.1.1+

jeffwecan:


jeffwecan:
Treating this as a prerequisite for https://github.com/hvac/hvac/pull/478.

PR #488, opened 2019-07-07
Update Travis CI Test Matrix With Latest Vault Version & Drop Python 3.6

jeffwecan:
* Tweaking supported Vault versions a smidge. Was inspired by #478 
* Dropping Python 3.6 test runs. We already have Python 3.7 coverage and dropping 3.6 will decrease the runtime of Travis CI builds quite a bit...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,9 @@ python:
   - '3.6'
   - '3.7'
 env:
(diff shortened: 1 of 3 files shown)


jeffwecan:
Treating this as a prerequisite for https://github.com/hvac/hvac/pull/478

PR #503, opened 2019-07-17
Add read_configuration For Mount System Backend Class

vamshideveloper:
Updated with read_configuration 
Need it to get full configuration response to check whether mount is configured or not

### DIFF
--- a/hvac/api/system_backend/mount.py
+++ b/hvac/api/system_backend/mount.py
@@ -106,6 +106,25 @@ class Mount(SystemBackendMixin):
             url=api_path,
         )
 
+    def read_configuration(self, path):
+        """Read the given mount's configuration.
+
+        This will return the mount configuration.
+
+        Supported methods:
+            GET: /{mount_point}/config. Produces: 200 application/json
+
+        :param path: Specifies the path where the secrets engine will be mounted. This is specified as part of the URL.
+        :type path: str | unicode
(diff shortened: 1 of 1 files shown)


jeffwecan:
Looks like the Travis ci build failed due to a small linting issue:

> flake8 run-test: commands[0] | flake8
./hvac/api/system_backend/mount.py:126:31: W291 trailing whitespace

hvac/api/system_backend/mount.py
```diff
@@ -106,6 +106,25 @@ def disable_secrets_engine(self, path):
             url=api_path,
         )
 
+    def read_configuration(self, path):
```
jeffwecan: So this doesn't seem like a `/sys/mount` sort of method in particular so I don't know if it fits within this class's scope. I.e., while _some_ secrets engines have a `/config` route it isn't universal. Is there a particular set of secrets engines, etc. that you would use this method for? If so, I would say the method would better fit under those engines' classes.

vamshideveloper: Sure then i will add this under the Active directory(secret engine) class
I read configuration to skip overriding configuration by reading the existing config

vamshideveloper:
Added read_configuration() to ad.py PR ( https://github.com/hvac/hvac/pull/507 )
you can close this PR
Thank you

PR #507, opened 2019-07-18
Add Active Directory Secrets Engine Support

vamshideveloper:
Module for Active directory
which will help us to manage AD through automation
I will also update hvac/hvac/api/secrets_engines/init.py to support ad.py


vamshideveloper:
Added read configuration to ad.py
with reference to https://github.com/hvac/hvac/pull/503/files

PR #512, opened 2019-08-01
Added list & create_update_policies definitions to access [child-namespaces] policies from [root-namespace]

vamshideveloper:
- Added definitions to list and create/update policy in specific namespace from root namespace
- Not sure if this is correct place for them to be
- This is added because I did not find a way to access the child-namespace/* from root-namespace, 
   Current module policy.py only allow us to GET/POST/LIST... to the namespace we are in

### DIFF
--- a/hvac/api/system_backend/namespace.py
+++ b/hvac/api/system_backend/namespace.py
@@ -99,4 +99,3 @@ class Namespace(SystemBackendMixin):
             url=api_path,
             json=params,
         )
-
--- a/hvac/api/system_backend/namespace.py
+++ b/hvac/api/system_backend/namespace.py
@@ -1,3 +1,4 @@
+import json
 from hvac.api.system_backend.system_backend_mixin import SystemBackendMixin
 
 
--- a/hvac/api/system_backend/namespace.py
+++ b/hvac/api/system_backend/namespace.py
@@ -47,3 +47,55 @@ class Namespace(SystemBackendMixin):
             url=api_path,
         )
         return response
+
+    def list_policies_namespace(self, path):
+        """List all configured policies in namespaces.
+
+        Supported methods:
+            GET: {path}/sys/policy. Produces: 200 application/json
+
+        :param path: Specifies the namespace_name to list policies. This is specified as part of the URL.
+        :type path: str | unicode
+        :return: The JSON response of the request.
+        :rtype: dict
+        """
+        api_path = '/v1/{path}/sys/policy'.format(path=path)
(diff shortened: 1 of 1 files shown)


hvac/api/system_backend/namespace.py
```diff
@@ -47,3 +48,54 @@ def delete_namespace(self, path):
             url=api_path,
         )
         return response
+
+    def list_policies_namespace(self, path):
```
jeffwecan: I'm not familiar with this API route. Do you know if its recorded under https://www.vaultproject.io/api/system/ or otherwise? It feels like it may fit better under https://github.com/hvac/hvac/blob/develop/hvac/api/system_backend/policy.py but I'd want to look up the upstream code / documentation to get a better idea of that scope before saying for sure...

vamshideveloper: https://learn.hashicorp.com/vault/operations/namespaces#api-call-using-curl-1

Above document explains how this will work

vamshideveloper: I am referring to below API call in the document
API call using cURL
To target a specific namespace, you can do one of the following:

Option 1: Pass the target namespace in the X-Vault-Namespace header

Option 2: Prepend the API endpoint with namespace name (e.g. <namespace_name>/sys/policies/acl)

drewmullen:
hey @vamshideveloper im sure this is frustrating to hear but these functions definitely do work in child namespaces. i use them all the time to list / create / update policies in nested namespaces.

the key is to make sure the namespaces are passed to the client when you're instantiating the class. heres and example:

```
client = hvac.Client(url=url, cert=cert, verify=verify, namespace=namespace)
```

If you have questions, we have a gitter that you can join to discuss: https://gitter.im/hvac/community

drewmullen:
the outcomes of the call are identical so it would be interesting to see why you need to perform it this one. this proposed design pattern could have far reaching implications involving duplicate functions for many, if not all api endpoints. some may work with the `mount_point` var but I havent tried that. 

is there any reason you cant use the header? 

I can think of at least 1 reason why **not** to use the path prepend, `root` does not work. whereas it does work if you pass via the header: https://github.com/hashicorp/vault/issues/5987#issuecomment-506019924

vamshideveloper:
https://github.com/hvac/hvac/pull/512#issuecomment-517723979
This works for me
I will close this PR
Thanks

drewmullen:
thanks very much! please feel free to join the gitter to discuss this or future questions! (or you can always open issues too) :D

Issue #515, opened 2019-08-19
New auth method to hit a local agent listener


semrov:
Add additional authentication method that will support passing authentication to Vault Agent. In this case you specify a **listener** in Agent’s configuration file and set **use_auto_auth_token** to **true**  (see [Vault Agent Caching](https://www.vaultproject.io/docs/agent/caching/index.html) for details, or [practical example](https://www.katacoda.com/hashicorp/scenarios/vault-agent) – step 3). This will allow fetching secrets without specifying authentication data, since Agent will handle it. 

![Agent's config file](http://shrani.si/f/26/OC/4DXLZ81Y/agentcachingconf.png)

From agent docs:
"... When this configuration is set, if the request doesn't already bear a token, then the auto-auth token will be used to forward the request to the Vault server. This configuration will be overridden if the request already has a token attached, in which case, the token present in the request will be used to forward the request to the Vault server."

![Vault caching](https://d33wubrfki0l68.cloudfront.net/a4d8046f2bd10378200cba23ee19545855da64a8/46415/static/img/vault-caching-3.png)

In simpler words:
1)	If no additional authentication data (like token, role id etc.) is provided - Just pass request to Vault agent - Agent is entirely responsible for authentication. This means, if you “curl” to some secret using Agent’s listener address, Agent first authenticates with Vault server (using data provided in configuration file) and forwards secret URL to Vault server - Vault server then forwards secret back to Agent which forwards secret back to “curl” (assuming Agent has valid credentials).

2)	Additional authentication data (like token, role id etc.) is provided - pass provided authentication data to Vault Agent, Vault Agent now uses authentication data provided  - overrides data in Agent's configuration file.

This would be useful when running Ansible scripts, users wouldn’t have to export authentication data into environment.

```yaml 
- name: Get secert from Vault
  set_fact:
     secret: "{{ lookup('hashi_vault', 'auth_method=agent secret={{ item.secret_path }} }}"
```

Currently HVAC module raises exception, because no authentication data is provided.

PR #521, opened 2019-09-11
Add Token Authentication Method Class

Tylerlhess:
Wrote the basic auth module for tokens. This allows you to authenticate with only setting the root_token variable or by using the auth_cubbyhole method to unwrap a wrapped token.

Also have some basic documentation for the databases secrets engine that I was working on.


Tylerlhess:
I guess this is only needed if using a token outside of the constructor.

drewmullen:
so can you help me with some examples for how this would be used? if we cant find an immediate usage, im thinking we could repurpose this for [this use case](https://github.com/hvac/hvac/issues/515)

jeffwecan:
Closing this PR as its gone a bit stale. Feel free to re-open it, or a new issue, as needed!

PR #525, opened 2019-09-18
When creating an alias that already exists this creates an error

tom-dierckx:
Fixed by using logic also used for other created objects (that already exists and returns status 204).
If you create an alias on an entity that already exists this gives following error:
```
ERROR: test_create_entity (__main__.TeamTests)
  File "/home/tom/repositories/gitlab-a/vault-provisioner/tests.py", line 40, in test_create_entity
    self.entity_handler.create(user)
  File "/home/tom/repositories/gitlab-a/vault-provisioner/handlers/entity_handler.py", line 29, in create
    self._create_alias(entity_id, user.id)
  File "/home/tom/repositories/gitlab-a/vault-provisioner/handlers/entity_handler.py", line 24, in _create_alias
    mount_accessor=self.__config.oidc_accessor,
  File "/home/tom/repositories/gitlab-a/vault-provisioner/venv/lib/python3.6/site-packages/hvac/api/secrets_engines/identity.py", line 373, in create_or_update_entity_alias
    return response.json()
  File "/home/tom/repositories/gitlab-a/vault-provisioner/venv/lib/python3.6/site-packages/requests/models.py", line 897, in json
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
  File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### DIFF
--- a/hvac/api/secrets_engines/identity.py
+++ b/hvac/api/secrets_engines/identity.py
@@ -370,7 +370,10 @@ class Identity(VaultApiBase):
             url=api_path,
             json=params,
         )
-        return response.json()
+        if response.status_code == 204:
+            return response
+        else:
+            return response.json()
 
     def read_entity_alias(self, alias_id, mount_point=DEFAULT_MOUNT_POINT):
         """Query the entity alias by its identifier.


llamasoft:
@jeffwecan Unfortunately, I think the safest solution is that the empty-or-json logic needs to be present for _all_ responses.  (Either that or always return the Response object directly, but that's a massive breaking change.)

Vault's [API documentation](https://www.vaultproject.io/api/overview.html#http-status-codes) even explicitly says that...
> Applications should be prepared to accept both 200 and 204 as success. 204 is simply an indication that there is no response body to parse, but API endpoints that indicate that they return a 204 may return a 200 if warnings are generated during the operation.

That means we can't even trust DELETE requests to always return a 204 either. ☹️

jeffwecan:
Word, thanks for sharing that. Given that caveat, perhaps it makes more sense to move this logic into the main request method... 🤔

llamasoft:
I could see that being a solution.  Additionally, I think I would prefer if the `Request` adapter was renamed to something like `RawRequest` and a new one with the above logic created with a name like `JSONRequest`.  `JSONRequest` would become the client's new default adapter and a user could supply `RawRequest` if they want unprocessed Response objects.

jeffwecan:
Yeah, I'm down with that. Are you imagining this proposed `JSONRequest` class would return a dict in all cases? I.e., if it gets a 204 response, it would return an empty dictionary, or?

llamasoft:
I had to give this one a good bit of thought.  On the one hand, even non-200 responses [can return JSON](https://www.vaultproject.io/api/overview.html#error-response).  On the other hand, that would probably require rewriting a lot of tests and would probably break a lot of people's code.

I think that in order to minimize breakage for existing users, a 204 response should return the Response itself, not an empty dict.  My decision is based entirely on the fact that [`bool(response204) == True`](https://github.com/psf/requests/blob/fab1fd10d0b115e635b9ef1364f8444089725000/requests/models.py#L673-L681) but `bool(empty_dict) == False`.  Returning an empty dictionary would cause any instances of `if client.do_a_thing()` to break if was previously capable of returning a 204 Response object.  
By extension, even if an _error_ Response contains a JSON body, the error Response object itself should be returned, otherwise the non-empty dictionary might be misinterpreted as success.

llamasoft:
I've started work on the JSON adapter as discussed.  I should have it ready in a day or two.

jeffwecan:
Closing this PR in favor of https://github.com/hvac/hvac/pull/537/ (which effectively includes the same sort of modification).

Issue #530, opened 2019-09-30
Test errors during setUpClass leave vault instance running


llamasoft:
If an integration test fails during a setUpClass call, the resulting vault instance is never terminated.  This causes all subsequent integration tests to fail because they expect to be interacting with an uninitialized vault instance.

For example, this is the first integration test to failure was caused by not having Java installed in my Docker image.  (Related to #529)

```
======================================================================
ERROR: test suite for <class 'tests.integration_tests.api.auth_methods.test_ldap.TestLdap'>
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/hvac/tests/integration_tests/api/auth_methods/test_ldap.py", line 22, in setUpClass
    cls.ldap_server = MockLdapServer()
  File "/hvac/tests/utils/mock_ldap_server.py", line 64, in __init__
    'entries': self.ldap_entries,
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/ldap_test/server.py", line 202, in __init__
    SERVER_PROCESS = run_jvm_server(java_gateway_port)
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/ldap_test/server.py", line 72, in run_jvm_server
    raise Exception("'java' executable not found in system path!")
Exception: 'java' executable not found in system path!
```

From this point forward, all of the remaining integration tests fail on the following:

```
Traceback (most recent call last):
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/hvac/.tox/py27/local/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
[...]

PR #531, opened 2019-10-10
Improve Integration Test Error Handling

### LINKED ISSUE #207: Integration Tests Should Reset State Between Test Cases
Relying on test cases to both 1.) verify the dev Vault instance is in an appropriate state to run and 2.) ensure they've reset the state of things after running assertions can:

* Make it harder to jump in and write a new test case (due to the implicit knowledge needed to write one "correctly" within the current scope of how things are configured).
* Opens us up to unexpected side-effects and test failures that are unrelated to the actual logic under test.

We should reset the state of the dev Vault cluster during setUp/tearDown (or something similar) as much as possible.

llamasoft:
This MR fixes issue #530 and partially fixes #207.

It includes the following changes:

- Throw an exception if a Vault instance is already running.
  - This is mainly to prevent interacting with someone's personal Vault instance or tunnel as if it were a test instance.
- If any integration test fails, the created Vault instance(s) are terminated.
  - The ServerManager's stop method works correctly regardless if any instances were actually created, so it's always safe to call it.
- An exception is now thrown if the Vault instance terminates before becoming ready.
  - This handles cases where the port is already in use or some other unspecified error occurs.
- If a Vault executable isn't found, the integration tests are skipped.
- If a Java executable isn't found, the LDAP integration tests are skipped.
- Added a few extra debug statements when starting/killing Vault instances to include process IDs.
- Minor rewording of variables and docstrings in the ServerManager class.

### DIFF
--- a/tests/integration_tests/api/auth_methods/test_ldap.py
+++ b/tests/integration_tests/api/auth_methods/test_ldap.py
@@ -27,7 +27,7 @@ class TestLdap(HvacIntegrationTestCase, TestCase):
             cls.mock_server_port = utils.get_free_port()
             cls.ldap_server = MockLdapServer()
             cls.ldap_server.start()
-        except:
+        except Exception:
             # Ensure that Vault server is taken down if setUpClass fails
             super(TestLdap, cls).tearDownClass()
             raise
--- a/tests/integration_tests/api/auth_methods/test_ldap.py
+++ b/tests/integration_tests/api/auth_methods/test_ldap.py
@@ -18,7 +18,7 @@ class TestLdap(HvacIntegrationTestCase, TestCase):
     def setUpClass(cls):
         # The mock LDAP server requires Java runtime
         if not distutils.spawn.find_executable('java'):
-            raise SkipTest
+            raise SkipTest('The mock LDAP server requires Java runtime')
 
         try:
             super(TestLdap, cls).setUpClass()
--- a/tests/integration_tests/api/auth_methods/test_ldap.py
+++ b/tests/integration_tests/api/auth_methods/test_ldap.py
@@ -1,5 +1,6 @@
+import distutils.spawn
 import logging
-from unittest import TestCase
+from unittest import TestCase, SkipTest
 
 from parameterized import parameterized, param
 
@@ -15,12 +16,21 @@ class TestLdap(HvacIntegrationTestCase, TestCase):
 
     @classmethod
     def setUpClass(cls):
-        super(TestLdap, cls).setUpClass()
-        logging.getLogger('ldap_test').setLevel(logging.ERROR)
+        # The mock LDAP server requires Java runtime
+        if not distutils.spawn.find_executable('java'):
+            raise SkipTest
 
-        cls.mock_server_port = utils.get_free_port()
-        cls.ldap_server = MockLdapServer()
-        cls.ldap_server.start()
+        try:
+            super(TestLdap, cls).setUpClass()
+            logging.getLogger('ldap_test').setLevel(logging.ERROR)
+
+            cls.mock_server_port = utils.get_free_port()
+            cls.ldap_server = MockLdapServer()
+            cls.ldap_server.start()
+        except:
+            # Ensure that Vault server is taken down if setUpClass fails
+            super(TestLdap, cls).tearDownClass()
+            raise
 
     @classmethod
     def tearDownClass(cls):
--- a/tests/utils/__init__.py
+++ b/tests/utils/__init__.py
@@ -20,8 +20,8 @@ LATEST_VAULT_VERSION = '1.1.3'
 
 
 def get_vault_version_string():
-    if not find_executable("vault"):
-        raise SkipTest
+    if not find_executable('vault'):
+        raise SkipTest('Vault executable not found')
     command = ['vault', '-version']
     process = subprocess.Popen(**get_popen_kwargs(args=command, stdout=subprocess.PIPE))
     output, _ = process.communicate()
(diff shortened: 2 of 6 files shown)


tests/integration_tests/api/auth_methods/test_ldap.py
```diff
@@ -15,12 +16,21 @@ class TestLdap(HvacIntegrationTestCase, TestCase):
 
     @classmethod
     def setUpClass(cls):
-        super(TestLdap, cls).setUpClass()
-        logging.getLogger('ldap_test').setLevel(logging.ERROR)
+        # The mock LDAP server requires Java runtime
+        if not distutils.spawn.find_executable('java'):
+            raise SkipTest
```
jeffwecan: This comment could be migrated to the exception instance being raised, no? It would still exist for folks looking 
```suggestion
            raise SkipTest('The mock LDAP server requires Java runtime')
```

tests/utils/__init__.py
```diff
@@ -18,6 +20,8 @@
 
 
 def get_vault_version_string():
+    if not find_executable("vault"):
+        raise SkipTest
```
jeffwecan: Same deal here:

```
$ nosetests -v tests/integration_tests/v1/test_integration.py
Failure: SkipTest () ... SKIP

Ran 1 test in 0.000s

OK (SKIP=1)
```
versus:

```
def get_vault_version_string():
    if not find_executable("vault"):
        raise SkipTest('CAN NOT FIND NO VAULT')
    command = ['vault', '-version']
```

```
$ nosetests -v tests/integration_tests/v1/test_integration.py
Failure: SkipTest (CAN NOT FIND NO VAULT) ... SKIP: CAN NOT FIND NO VAULT

Ran 1 test in 0.000s

OK (SKIP=1)
```

To be clear, the motivation in providing a message is mainly so developers have information about why tests are being skipped so they don't end up getting caught unaware by the CI tests run after opening a PR.

tests/utils/server_manager.py
```diff
@@ -33,31 +33,46 @@ def __init__(self, config_paths, client, use_consul=False):
...
         if self.use_consul:
             self.start_consul()
 
-        cluster_initialized = False
+        # If a vault server is already running then we won't be able to start another one.
+        # If we can't start our vault server then we don't know what we're testing against.
+        try:
+            self.client.sys.is_initialized()
+        except Exception:
+            pass
```
jeffwecan: Since this is a test utility, feels prudent to log (i.e., `logger.warn('Vault already initialized')` or what-have-you) when we hit this case so users have as much information as possible no?

tests/utils/server_manager.py
```diff
@@ -67,25 +82,35 @@ def start(self):
                 )
 
     def start_consul(self):
+        try:
+            requests.get('http://127.0.0.1:8500/v1/catalog/nodes')
+        except Exception:
+            pass
```
jeffwecan: nit: Also would like to see logging of this case as well.

tests/integration_tests/api/auth_methods/test_ldap.py
llamasoft: I'm not sure why I didn't think of that.  It should be an easy fix.

tests/utils/server_manager.py
llamasoft: It outputs the error as part of the exception in the `else` clause.  The [else clause ](https://docs.python.org/2.7/tutorial/errors.html#handling-exceptions) only runs if an exception _wasn't_ thrown by `self.client.sys.is_initialized()` which indicates that Vault is already running.

tests/utils/server_manager.py
llamasoft: Similar to the previous, the error message is displayed via thrown exception in the "else" clause of the try..catch..else block.

tests/utils/server_manager.py
jeffwecan: Ah right, checks out! 👍

llamasoft:
I tried throwing in a few tweaks to get the Travis builds to successfully built Vault but I haven't had much luck yet.  It no longer complains about the go version, but now it's complaining about [an unknown field](https://travis-ci.org/hvac/hvac/jobs/596792827#L261-L265) that [absolutely does exist](https://github.com/hashicorp/vault/blob/68750b70a2c139c68b4ea2aeecfd1317fef653ef/api/sys_raft.go#L17-L24).  
Unfortunately, I'm not at all experienced with go, so my ability to troubleshoot this is severely limited.  It's also hampered by the fact that I can't test under Travis directly, I can only push new commits and hope for the best.  That said, I managed to build the Vault binary locally in Docker using go 1.12.7, so the issue lies either in the Travis environment or the utility script used to build Vault.

Regardless, I'm moving my Travis tests to a new branch (as I should have done earlier!).

jeffwecan:
Oh those particular build failures are no major concern. Or at least not meant to block merging hah. 1.) the intention is having the source builds in place is sort of a leading indicator of potential upcoming issues 2.) sometimes the build failures are due to unrelated upstream issues. Which is why we have those particular test runs under the "allow failures" travis-ci distinction. I haven't looked into the specifics of the current failures to be clear. Still probably worth calling out those nuances in the contributing documentation w.r.t. upstream builds. 

That said, the Travis test matrix is probably out of date at this point and should most likely be updated. However, that need not block this very valuable addition in the meantime.

jeffwecan:
tl;dr: if you care to attempt to get the head vault branch builds passing in this PR, feel free. Otherwise I'm happy to merge things as is.

llamasoft:
I'll work on the head branch build issue in a separate PR.  I'll make one PR for testing, then a separate PR once I actually work out a fix.  👍

Issue #483, opened 2019-07-03
Standardize on What (If Any) Manipulation Should Be Done on Response Data Before Returning the Value


jeffwecan:
This issue inspired by @drewmullen's comment in https://github.com/hvac/hvac/pull/482#pullrequestreview-257492251:

> @jeffwecan we should consider standardizing json returns. personally i like this where the data key is returned directly, however, we have some that doesnt extract it. not sure its worth going back and changing since that would have dowstream effects but, we could consider it
> 
> example:
> 
> https://github.com/hvac/hvac/blob/c013371a43d30b5b54407f3fea2f2fd76bbe5934/hvac/api/auth_methods/ldap.py#L129

jeffwecan:
💯 agreed! Having a consistent pattern across the board is important in terms of reducing confusion, etc.

I can see an argument for just returning the unmarshalled response and nothing else (since there _is_ other metadata in the top-level keys that folks conceivable might want access to). On the other hand returning the 'data' key as seen here make working with the return far cleaner looking in most all cases. I suppose the first actionable things to track down to that first part are:

- Are there any reasonable use cases we can imagine where folks would want access to the top-level response keys (e.g., `request_id`, `warnings`, etc.)?
- Assuming there are such cases, can we provide some way to access those items while still returning `response.json()['data']` as the default? Perhaps add something to the adapter class (which may be the most appropriate place to "standardize" this logic anywho) that keeps a list of the last n response in full? Vaguely imaging something vaguely similar to the idea of how [python-requests exposes redirects](https://2.python-requests.org/en/master/user/quickstart/#redirection-and-history)...

drewmullen:
example output list of non-`data` dict keys:

- request_id
- lease_id
- renewable
- lease_duration
- wrap_info
- warnings
- auth

after looking at this list, i do think it's likely people would want to have access to this info, especially `lease_duration`. 

the question now is implementation! I like the idea of adding it to the `adapter`. we had also once [discussed manipulating the JSONEncoder](https://github.com/hvac/hvac/issues/382#issuecomment-456579828), that might be another management point(?)

Yakuza-UA:
[...]

PR #537, opened 2019-10-17
Make returned responses more consistent

llamasoft:
This is another big PR so apologies in advance.  This PR fixes issue #525 and arises from the discussion therein.

This PR does one thing: create a new JSONAdapter and makes it the default.  The "JSON adapter" works just like the previous "Request" adapter except the return value uses the following logic:

    if response.status_code == 200:
        try:
            return response.json()
        except ValueError:
            pass
    return response

The net result is that _all_ client functions now have consistent logic regarding what they return. 
 As a result, all instances of the following client logic have been removed:

    return response.json()

and 

    if response.status_code == 204:
        return response
    else:
        return response.json()

This _will_ likely cause some breaking changes for some users, especially with regards to the direct `client.[read,list,write,delete]` methods as they will no longer throw a ValueError from assuming the response _always_ contains a JSON body.


llamasoft:
I just had the last-second realization that the `login` method on the Adapter base class doesn't make sense.  The `request` method it depends on is abstract, so we have zero guarantees about the object type it returns.  Instead, I've moved the token extraction logic to a separate abstract method, `get_login_token`.

I also forgot to mention in the original PR that I added an adapter option called `ignore_exceptions` that optionally overrides the `raise_exception` parameter passed to `request`.  This allows a way to suppress exceptions and receive a raw Response object (normally not possible because the `raise_exception` argument isn't accessible).

llamasoft:
I just realized an additional breaking change regarding this PR: the `lookup_*` methods (e.g. `client.secrets.identity.lookup_[entity|group]`) now return a Response[204] (truthy) instead of None (falsy) if the lookup returned no results.  

Unfortunately, I can't think of a good way to resolve this issue.  The Requests Adapter is created from a user-specified class so we can't assume in advance what sort of data type it will return.  
Honestly, given that we can't guarantee return types, we probably shouldn't be interacting with response objects _at all_, including the helpers on `read_*` methods that return `response.get('data')`.

What are your thoughts on this matter?  I can revert the `lookup_*` to return None on 204 responses, but that logic may end up causing issues with custom adapters.

llamasoft:
Any word on if `response.get('data')` should be removed as well?  It's used in a few places, but not in others.  In my personal experience, it made return values unpredictable.  For example, most `read_*` methods use it, but `lookup_*` methods don't.

Unfortunately, removing them _will_ be a breaking change for some users.

jeffwecan:
> Any word on if response.get('data') should be removed as well? 

I suppose I feel it makes most sense to return the entire response in all cases. In general, this module is intended to be a pretty basic wrapper for Vault API (at least as far as I'm concerned).

drewmullen:
I agree with @jeffwecan . we had explored the idea of making all returns just the `data` portion but there are some useful values provided from vault that shouldnt be stripped (see below).

https://github.com/hvac/hvac/issues/483#issuecomment-509199918

Lucas-C:
This looks like an nice improvement !
What is the status of this PR ? :)

jeffwecan:
@Lucas-C: I figure it is about time to get this PR put in place! :D 

Will start working on getting it trued back up with the current state of the develop branch...

jeffwecan:
Planning on merging this in today. To close the loop on returning the entire response dicts versus only the data key: I still want to see use move to the former in all places eventually. However I'd rather tackle that in another PR / release since its most certainly a breaking change (as @llamasoft alluded to).

trishankatdatadog:
Okay, this one was really confusing, because the [documentation](https://github.com/hvac/hvac/blob/973c37d2624ef26cdecbff7a6cf8ba62285bdb1e/docs/usage/secrets_engines/transit.rst), which really copies doctstrings from the source code, still says to expect `requests` responses instead of the JSON body for non-204 responses. Should we update the docstrings?

jeffwecan:
I would imagine so, and sorry for the added confusion! @trishankatdatadog, I went ahead and quoted your comment into a new issue (#537) and will try to get a review of docstrings, etc. done, and any incorrect return types updated, sometime over the next week or so. (Though contributions along those lines are of course also very welcomed. ☺)

trishankatdatadog:
> I would imagine so, and sorry for the added confusion! @trishankatdatadog, I went ahead and quoted your comment into a new issue (#537) and will try to get a review of docstrings, etc. done, and any incorrect return types updated, sometime over the next week or so. (Though contributions along those lines are of course also very welcomed. ☺)

Thanks! Also wrong [link](https://github.com/hvac/hvac/issues/604) for issue 🙂

PR #545, opened 2019-11-21
Fix Policy.list_policies() to use _adapter.list()

hepoun:


hvac/api/system_backend/policy.py
```diff
@@ -16,7 +16,7 @@ def list_policies(self):
         :rtype: dict
         """
         api_path = '/v1/sys/policy'
-        response = self._adapter.get(
+        response = self._adapter.list(
```
jeffwecan: What is the motivation for this change? Upstream Vault API documentation appears to indicate that GET is the correct method here: https://www.vaultproject.io/api/system/policy.html#list-policies

hepoun: I checked again and it's my bad.

I followed HashiCorp's guide where just the `list` permission was used, but I missed it was for `sys/policies/acl` path. To have `read` permission for `sys/policy` was the correct solution to my problem, not to change `list_policies()` in hvac.

So to stick with the Vault documentation, this pull-request is to be closed without merging.

I'm sorry for wasting your time.

PR #533, opened 2019-10-15
Don't send optional parameters unless explicitly specified

llamasoft:
Apologies in advance, this is a big PR.  This fixes issue #527 in addition to a few other small things.

This PR primarily does two things:
1. Optional parameters now have a default value of None.
   - We no longer have to worry about our default values conflicting with Vault's default values.  We now just let Vault decide what the correct default should be, even if that value changes between versions.
1. Optional parameters are no longer sent when their value is None.
   - The "update" functions no longer clobbers existing values if the user didn't specify them.

**This PR _may_ contain breaking changes** (depending on what your definition of "breaking" is).  If anything, the "clobbering" side effect has been removed and some previously required parameters have been made optional.

I did my best to use the following pattern when creating the `params` value for API calls:

```
# Required parameters are always set
params = {
    'id': id,
    'username': username,
}

# Then optional parameters are added, assuming they're not None
params.update(
    utils.remove_nones({
        'ttl': ttl,
        'metadata': metadata,
    })
)

# Then we add optional parameters that require special handling
if policies is not None:
    params['policies'] = ','.join(policies)

# If we have a catch-all argument, it's added last
params.update(extras)
```

Please let me know if you have any questions!

PR #553, opened 2020-02-03
Fix a bug when creating external groups

GuillaumeRemyCSI:
When creating external group, the API call includes the members keys even if they are null, and this results in the API responding with the error "Exception: member entities can't be set manually for external groups"

### DIFF
--- a/hvac/api/secrets_engines/identity.py
+++ b/hvac/api/secrets_engines/identity.py
@@ -815,6 +815,9 @@ class Identity(VaultApiBase):
             mount_point=mount_point,
             name=name,
         )
+        filtered = {k: v for k, v in params.items() if v is not None}
+        params.clear()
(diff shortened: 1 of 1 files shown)


jeffwecan:
@GuillaumeRemyCSI My bad for not getting this related PR merged yet, but I believe https://github.com/hvac/hvac/pull/533 will cover this scenario as well. I'll double check and get that verified once I'm able to get #533 in.

GuillaumeRemyCSI:
@jeffwecan indeed, #533 will cover this use case. I'm waiting for it to be merged, thanks.

jeffwecan:
https://github.com/hvac/hvac/pull/533 has (finally) been merged. I'll look to get it released in a new version of the next day or so. Closing this PR out in the meantime.

PR #557, opened 2020-02-11
Fixes close quotes in example usage of read_secret_version

philherbert:
for those of us who are reckless copy-pasters


jeffwecan:
Ideally we'll move these examples over to doctest in the future to avoid such typo's in the future. :)

Issue #559, opened 2020-02-18
Userpass.create_or_update_user() does not allow to specify policies


stefanlasse:
Hi Folks,
according to [here](https://www.vaultproject.io/docs/auth/userpass/#configuration) and [here](https://www.vaultproject.io/api/auth/userpass/index.html#createupdate-user) one must provide a policy for a user to be created in order to receive a token with the respective policy attached.

Using `hvac.api.auth_methods.Userpass.create_or_update_user()` does not allow to specify a policy. How do you do that?

Thanks in advance for the support!

jeffwecan:
My understanding is that if no policies are not specified, the resulting token would simply receive the "default" policy.

> In addition, another built-in policy, default, is created. The default policy is attached to all tokens and provides common permissions.

(quoted from https://www.vaultproject.io/guides/identity/policies/)

stefanlasse:
Yes, that is what i understood, too. But if i want to attach a policy other than the default policy to a freshly baked user, how do i do that? For example with LDAP this is possible (see [here](https://hvac.readthedocs.io/en/stable/source/hvac_api_auth_methods.html#hvac.api.auth_methods.Ldap.create_or_update_user)), for userpass it's not, see [here](https://hvac.readthedocs.io/en/stable/source/hvac_api_auth_methods.html#hvac.api.auth_methods.Userpass.create_or_update_user).

vamshideveloper:
Opened https://github.com/hvac/hvac/pull/562 which can accept policies while creating users

PR #562, opened 2020-03-09
Add `policies` Parameter to Userpass `create_or_update_user()` Method

vamshideveloper:
Fixes https://github.com/hvac/hvac/issues/559


hvac/api/auth_methods/userpass.py
```diff
@@ -22,10 +23,13 @@ def create_or_update_user(self, username, password, mount_point=DEFAULT_MOUNT_PO
         :type username: str | unicode
         :param password: The password for the user. Only required when creating the user.
         :type password: str | unicode
+        :param policies: The list of policies to be set on username created.
+        :type policies: str | unicode
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
         """
         params = {
+            'policies': policies,
```
jeffwecan: According to the API docs this should instead be `token_policies`, no?: https://www.vaultproject.io/api/auth/userpass/index.html#inlinecode-token_policies-7

hvac/api/auth_methods/userpass.py
```diff
@@ -4,14 +4,15 @@
...
+DEFAULT_POLICIES = 'default'
 
 
 class Userpass(VaultApiBase):
     """USERPASS Auth Method (API).
     Reference: https://www.vaultproject.io/api/auth/userpass/index.html
     """
 
-    def create_or_update_user(self, username, password, mount_point=DEFAULT_MOUNT_POINT):
+    def create_or_update_user(self, username, password, policies=DEFAULT_POLICIES, mount_point=DEFAULT_MOUNT_POINT):
```
jeffwecan: We've found that specifying this sort of default argument and always passing it along in the API request leads to unexpected issues generally. How would you feel about instead defaulting to `None` and taking advantage of the recently added `remove_nones()` / https://github.com/hvac/hvac/blob/develop/hvac/utils.py#L289 in the subsequent params deal a la:

```python
        params = utils.remove_nones({
            'policies': policies,
            'password': password,
        })
```

hvac/api/auth_methods/userpass.py
vamshideveloper: But in https://www.vaultproject.io/api/auth/userpass/index.html#sample-payload it shows "policies"

hvac/api/auth_methods/userpass.py
vamshideveloper: Updated

hvac/api/auth_methods/userpass.py
lenaing: It's explained here https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#120-july-30th-2019 this PR tries to adapt hvac to this change : https://github.com/hvac/hvac/pull/563

PR #563, opened 2020-03-20
Switch to new common token fields for auth backends

lenaing:
Hello,

This PR adapts HVAC to the Token store changes mentioned here :
https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#120-july-30th-2019 

> Token store roles use new, common token fields for the values that overlap with other auth backends. period, explicit_max_ttl, and bound_cidrs will continue to work, with priority being given to the token_ prefixed versions of those parameters. They will also be returned when doing a read on the role if they were used to provide values initially; however, in Vault 1.4 if period or explicit_max_ttl is zero they will no longer be returned. (explicit_max_ttl was already not returned if empty.)

This include breaking changes as method parameters are renamed to their `token_` prefixed versions.
Kind regards,

### DIFF
--- a/docs/usage/auth_methods/azure.rst
+++ b/docs/usage/auth_methods/azure.rst
@@ -85,7 +85,7 @@ Create a Role
 
     client.auth.azure.create_role(
         name='my-role',
-        policies=policies,
+        token_policies=policies,
         bound_service_principal_ids=bound_service_principal_ids,
     )
 
@@ -105,7 +105,7 @@ Read A Role
     )
     print('Policies for role "{name}": {policies}'.format(
         name='my-role',
-        policies=','.join(read_role_response['policies']),
+        policies=','.join(read_role_response['token_policies']),
     ))
 
 List Roles
--- a/docs/usage/auth_methods/gcp.rst
+++ b/docs/usage/auth_methods/gcp.rst
@@ -142,7 +142,7 @@ Source reference: :py:meth:`hvac.api.auth.Gcp.read_role`
 
     print('Policies for role "{name}": {policies}'.format(
         name='my-role',
-        policies=','.join(read_role_response['policies']),
+        policies=','.join(read_role_response['token_policies']),
     ))
 
 List Roles
--- a/docs/usage/auth_methods/github.rst
+++ b/docs/usage/auth_methods/github.rst
@@ -39,7 +39,7 @@ Configure Connection Parameters
 
     client.auth.github.configure(
         organization='our-lovely-company',
-        max_ttl='48h',  # i.e., A given token can only be renewed for up to 48 hours
+        token_max_ttl='48h',  # i.e., A given token can only be renewed for up to 48 hours
     )
 
 Reading Configuration
@@ -54,7 +54,7 @@ Reading Configuration
 
     github_config = client.auth.github.read_configuration()
     print('The Github auth method is configured with a ttl of: {ttl}'.format(
-        ttl=github_config['data']['ttl']
+        ttl=github_config['data']['token_ttl']
     )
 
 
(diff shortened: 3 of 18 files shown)


lenaing:
Seems that it will also fix https://github.com/hvac/hvac/issues/559

jeffwecan:
Hi @lenaing, thanks for the PR!  It looks like some of the test cases need to be updated to accommodate these changes. I'll try to take a look at what would be required there in a few hours. 👍

lenaing:
Hello,
Should I only add the new fields and keep the old ones ? (Even if they are marked as "deprecated" in Vault's code and does not appear anymore in the documentation ?)
I believe this could be a fine transition without implying breaking changes.

As far as test cases are concerned, I believe to have made all the necessary changes matching mine. The failed test cases were already failing on my side before I start to work on this :)

jeffwecan:
I believe the failures we'd be concerned with here are for folks running versions of Vault earlier than 1.2.0. I'm working through some of this as part of updating hvac's test matrix in: https://github.com/hvac/hvac/pull/565. Which is to say, I don't have answers to your questions _yet_ because I'm still working through / thinking about them myself 😄.  E.g., while hvac is still officially supporting Vault <1.2, perhaps we should update the test cases to work with either scenario: https://github.com/hvac/hvac/pull/565/commits/d7a3f356162e016ed6868f7dd0a6f8c9f1770d21?

So my apologies for the delays there and just to reiterate thank you very much for taking the time to get this PR submitted!

jeffwecan:
Alrighty, so I took this PR and rebased it on top of develop (https://github.com/hvac/hvac/tree/tokenutilize_rebased) and we do indeed have some other fallout that would impacts integration tests for Vault <v1.2: https://travis-ci.org/github/hvac/hvac/builds/668876878

E.g. in the integration test suite for Vault `v1.1.5`:

```
ERROR: test_ec2_role_crud (tests.integration_tests.v1.test_integration.IntegrationTest)
  File "/home/travis/build/hvac/hvac/tests/integration_tests/v1/test_integration.py", line 647, in test_ec2_role_crud
    assert ('ec2rolepolicy' in foo_role['data']['token_policies'])
KeyError: 'token_policies'
-------------------- >> begin captured logging << --------------------
urllib3.connectionpool: DEBUG: Starting new HTTPS connection (1): localhost:8200
urllib3.connectionpool: DEBUG: https://localhost:8200 "GET /v1/sys/mounts HTTP/1.1" 200 None
urllib3.connectionpool: DEBUG: https://localhost:8200 "GET /v1/sys/auth HTTP/1.1" 200 1315
urllib3.connectionpool: DEBUG: https://localhost:8200 "POST /v1/sys/auth/aws-ec2 HTTP/1.1" 204 0
urllib3.connectionpool: DEBUG: https://localhost:8200 "PUT /v1/sys/policy/ec2rolepolicy HTTP/1.1" 200 [...]
urllib3.connectionpool: DEBUG: https://localhost:8200 "GET /v1/auth/aws-ec2/role/foo HTTP/1.1" 200 693
--------------------- >> end captured logging << ---------------------
```

I also worry about how hvac users of earlier versions of Vault might be affected. Mainly in the sense of this change potentially being confusing but also that it is a breaking change in terms of requiring end users to update kwargs, etc. in some cases.

All that said, I still want to get all tests passing with either version of these token fields as a generally Good and worthwhile thing to do. In the meantime, I'll keep thinking about what might be the best route forward for all hvac users (and I'm happy to take any suggestions and/or feedback along those lines!).

lenaing:
Hello,

What do you think about my proposal of only adding new fields and not removing the old ones at the moment ? :)

As much of the time it's just named parameters that are changing, I'm not sure why it would require end users to change their kwargs ?

Kind regards,

jeffwecan:
> What do you think about my proposal of only adding new fields and not removing the old ones at the moment ? :)

I think that may be the most reasonable option as long as this module is still supporting versions of Vault that use the previous parameter names.

> As much of the time it's just named parameters that are changing, I'm not sure why it would require end users to change their kwargs ?

I still need to understand this better myself. From what I've seen when updating hvac's tests, using the different parameter names in the requests to Vault leads to different behavior based on the Vault version. I unfortunately haven't found the time to investigate this fully and document the findings but I'll be able to better say if it is actually a concern or not once I do...

jeffwecan:
Since source has drifted quite a bit now since this PR's initial creation, I'm going to go ahead and close it out for now but will still consider revisit it in the future.

Issue #579, opened 2020-04-30
Add request info to exception objects


and-semakin:
Hello!

Sometimes it's difficult to figure out what's wrong with Vault ACL or paths inside of it. It would be very useful if `VaultError` objects contained some information about the request (not only about the response as it is now), for example, at least HTTP method and URL.

Thanks for the library!

and-semakin:
@jeffwecan if you like the idea would you like me to create a PR? I think it won't require a lot of changes.

jeffwecan:
Of course, feel free to give it a go! ☺

PR #583, opened 2020-05-04
Add request data to exception objects

and-semakin:
Follow up: https://github.com/hvac/hvac/issues/579

Please, feel free to request changes, I'm ready to fix/improve everything 😝

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -301,7 +301,13 @@ class RawAdapter(Adapter):
                     pass
             if errors is None:
                 text = response.text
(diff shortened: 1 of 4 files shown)


jeffwecan:
Looks good, thanks @and-semakin! I'm going to wait to merge this until I can see about maybe expanding some of the unit test cases first. I've seen scenarios in the past where including requests response attributes in exceptions doesn't quite work in all cases...

(To be clear, I don't have any specific concerns here. I just want to think things through a bit more once I can find the time before proceeding.)

PR #598, opened 2020-06-15
Extract "renew_self_token" from "renew_token"

jm96441n:
* Extract call to `renew-self` endpoint into it's own function
* Add deprecation message around calling `renew_token()` without the `token` parameter
---
I had mentioned this in the gitter channel, but I am extracting the `renew-self` endpoint into it's own function. As of now the `renew_token()` function can call either `renew` or `renew-self`. The issue with this comes into play when a client has permissions to renew its own token via the `renew-self` endpoint but passes a token in to the `renew_token()` function and the `renew` endpoint doesn't recognize this permission and fails with `permission denied`. I think that by extracting the `renew-self` endpoint functionality into it's own function will make the library clearer to use, while also having these functions more closely adhere to SRP.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -6,6 +6,7 @@ from base64 import b64encode
 
 from hvac import aws_utils, exceptions, adapters, utils, api
 from hvac.constants.client import DEPRECATED_PROPERTIES, DEFAULT_URL
+from hvac.utils import generate_property_deprecation_message
 
 try:
     import hcl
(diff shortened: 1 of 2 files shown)


jm96441n:
Currently investigating the test failure, the original test seems to fail on develop when the token is not provided to the `renew_token` function, which makes me think this is a pre-existing issue.

jm96441n:
Nailed it down, in the tests when you don't pass in the token and call `create_token` the token for the adapter is not updated so it is using an old token, by calling the setter `Client.token` the adapter now has the correct token and can correctly call the endpoint

PR #613, opened 2020-07-26
Add JWT/OIDC Authentication Method Classes

### LINKED ISSUE #607: Auth with OIDC
Is it possible to implement OIDC auth in this library?

https://www.vaultproject.io/docs/auth/jwt_oidc_providers

jeffwecan:
There integration test bits with dummy OIDC provider / server is perhaps a bit much, but it was helpful having around when validating things so 🤷 

Resolves #607


jeffwecan:
Also tagging https://github.com/ansible-collections/community.general/pull/154 in this PR seems it seems to cover a related topic...

PR #621, opened 2020-08-11
Support for Google IAP proxy authentication

Borg-Can-Code:
This proposed changed will allow services accounts that leverage a Google IAP secured Vault instanced to authenticate to the proxy prior to hitting the Vault API.

I designed it as a plug able mechanism for different advanced proxies.

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -93,8 +93,12 @@ hvac does not currently offer direct support of requests to a `Vault agent proce
 
 Using Vault behind Google IAP
 --------------------------------
+Official Google product page:  https://cloud.google.com/iap
 
-Vault instances secured behind Google IAP enjoy an extra level of protection due to the Google Cloud Platform role required to access the web application. In order to access your vault instance each request must contain a valid Google-issued OpenID Connect token in the authorization header.
+Vault instances secured behind Google IAP enjoy an extra level of protection due to the Google Cloud Platform role required to access the web application. In order to access your vault instance each request must contain a valid Google-issued OpenID Connect token in the authorization header via bearer tokens.
+
+Rather than build a static feature to allow only Google IAP it was decided to build an advanced_function parameter that could support a plugable proxy backend.  This was accomplished via passing in a dictionary container a string with the provider name which will need to match a valid key in `ProxyRouter.ADVANCED_PROXIES`
+
 .. code:: python
 
     advanced_proxy = {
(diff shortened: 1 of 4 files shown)


jeffwecan:
@Borg-Can-Code: My apologies again for it taking a bit of time before I was able to get to this PR. Now that I've looked over the proposed change a bit more, I'm not certain it makes sense to include it directly within the hvac module. In short, my thinking is along these lines:

1. There isn't anything involved with generating the Google IAP authorization header that is uniquely influenced by Vault being the service upstream of the IAP. 
2. hvac previously added the `adapter` attribute on its `Client` class explicitly to help navigate this sort of use case _without_ needing to include specialized code within hvac itself. 

To be clear, I'm happy to discuss the matter further as well if that explanation is unclear. Also, now that I've looked at the code more, here is an example of what I assume would work pretty much just as well (excepting requiring a bit more work on the end user's part) as the additions in this PR:


```python
#!/usr/bin/env python

import hvac
from google.auth.transport.requests import Request
from google.oauth2 import id_token


class GoogleIAPAdapter(JSONAdapter):
    """Adapter that handles Google IAP-related authentication headers for all requests."""

    def __init__(self, google_iap_client_id):
        self.proxy_id_token = id_token.fetch_id_token(Request(), client_id)

    def request(self, *args, headers=None, **kwargs):
        headers['Authorization'] = f"Bearer {self.proxy_id_token}"

        return super(JSONAdapter, self).request(*args, **kwargs)


client = hvac.Client(
    url='https://vault.behind.googleiap.com',
    adapter=GoogleIAPAdapter,
    # Note: hvac.Client passes all keyword arguments when instantiating the adapter class specified.
    # So extra arguments like `google_iap_client_id` in this example would get passed on through to
    # `GoogleIAPAdapter`'s constructor.
    google_iap_client_id='YOUR_CLIENT_ID',
)
```

The above implementation is small enough that I don't feel that including it would be a big ask. However that snippet and the PR's patch both include an extra requirement in the form of `google-auth-oauthlib. Since we try to limit the dependencies hvac has as much as possible, I think a compelling case would need to be made for why it is necessary to include in the hvac module over elsewhere.

Borg-Can-Code:
Jeff provided a better way to accomplish this in the accompanying issue.

Adapter I am using to resolve this issue:


`class GoogleIAPAdapter(JSONAdapter):
    """Adapter that handles Google IAP-related authentication headers for all requests."""

    def __init__(self, google_iap_client_id: str, **kwargs):
        self.proxy_id_token = id_token.fetch_id_token(Request(), google_iap_client_id)
        super(GoogleIAPAdapter, self).__init__(**kwargs)

    def request(self, *args, headers={}, **kwargs):
        headers['Authorization'] = f"Bearer {self.proxy_id_token}"
        return super(GoogleIAPAdapter, self).request(headers=headers, *args, **kwargs)`

PR #628, opened 2020-09-11
Add basic SSH Secrets Engine implementation

0az:
This PR includes minimal documentation, and it only implements the subset used internally. In addition, it does not support Python 2, and might require Python 3.8, though I may be misremembering when `Literal` was implemented.

This is the last day of my internship, so I'm unsure if I'll clean up this PR anytime soon – it's not a high priority, since I don't use Vault myself.

See also: #514 

Signed-off-by: Andrew Zhou <afzhou@ibm.com>

### DIFF
--- a/hvac/api/secrets_engines/ssh.py
+++ b/hvac/api/secrets_engines/ssh.py
@@ -1,0 +1,264 @@
+"""SSH vault secrets backend module."""
+
+from typing import Dict, List, Literal, Optional, Union
+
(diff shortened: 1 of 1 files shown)


jeffwecan:
I'd like to get back to this PR and get it tidied up enough to merge. However, as I haven't found the time over the past six months, I'm going to go ahead and close this PR until I (or anyone else really 😜 ) is able to find that time.

PR #629, opened 2020-09-11
Fix PUT method in secrets engine kv_v1 to use PUT instead of POST

ghTravis:
When specifying `method='PUT'` in the `.create_or_update_secret` function it will use POST no matter what. This should be changed to `.put` instead of `.post`


jeffwecan (approved):
Thanks!

PR #635, opened 2020-09-17
Update health.py to match new Vault API query parameters

leongyh:
Updated to match updated query parameters on Vault API.

https://www.vaultproject.io/api-docs/system/health

Seems like the parameters do not have `_` anymore.


jeffwecan (approved):
Checks out :)

Thanks for the contribution!

PR #636, opened 2020-09-18
remove legacy policy type check

dogfish182:
The below change allows any version of consul newer than 1.4 to create tokens with the new policy system. 

Current release of hvac means I can only create a token that maps to an existing policy if it's a management type token, which is not ideal as newer versions will warn that you are using 'legacy type tokens' and suggest going to documentation to upgrade your tokens.
A management token has global rights within the system, so this is expressly not wanted most of the time for generating tokens.


jeffwecan (approved):
Thank you!

PR #638, opened 2020-09-27
Fix oidc login and add working example

### LINKED ISSUE #631: Issues on OIDC callback function
Hello all, 

first of all thank you for this great library! Usualy it is just awesome to work with it! 🥳 

As we use OIDC for login, I was super happy to see this was implemented a while ago. 
During the try to implement it as decribed in the docs, I faced several issues and I am now at a point, where I think there is a tiny issue in the  `oidc_callback` function. 😕 
On my testing is looks like the paramters are not sended in the GET request as they should . 

The snipped where the issue happens should be this one: 
```python
params = {
            'state': state,
            'nonce': nonce,
            'code': code,
        }
        api_path = utils.format_url(
            '/v1/auth/{path}/oidc/callback',
            path=self.resolve_path(path),
        )
        return self._adapter.get(
            url=api_path,
            json=params,
        )
```
Source: https://github.com/hvac/hvac/blob/master/hvac/api/auth_methods/jwt.py#L364

I can fix my issue if I modify the code to: 
```python
        api_path = utils.format_url(
            '/v1/auth/{path}/oidc/callback?state={state}&nonce={nonce}&code={code}',
            path=self.resolve_path(path),
            state=state,
            nonce=nonce,
            code=code
        )
``` 
I think this is not the way you are looking for get requests with paramters, but (even with a little search) I found no parts in the code where it is done diffrently. 
How would you preffer - Should I open a PR with this coding or is it easier if you change it directly?

Greetings from Stuttgart, Germany

Malte

derBroBro:
This PR fixed #631.
In addition to the fixed get request, I added an working example as well. 

If anything should be change, just let me know :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@
 
 ### 🧰 Miscellaneous
 
-- Update Vault version test matrix / Oldest Support Vault Version. GH-610
 
 ## 0.10.5 (July 26th, 2020)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Changelog
 
+## 0.10.6 ()
+
+### 🚀 Features
+
+
+### 🐛 Bug Fixes
+
+- Fixed oidc login. GH-638
+
+### 📚 Documentation
+
+- Added more detailed example for the oidc usage. GH-638
+
+### 🧰 Miscellaneous
+
+- Update Vault version test matrix / Oldest Support Vault Version. GH-610
+
 ## 0.10.5 (July 26th, 2020)
 
 ### 🚀 Features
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## 0.10.6 ()
```
jeffwecan, line 3: FYI, these changelog entries actually get drafted as PRs are merged and then updated during the release process (https://github.com/hvac/hvac/blob/develop/CONTRIBUTING.md#creating--publishing-releases). Including them in this PR is no trouble either tho.

jeffwecan (approved):
Thanks for the contribution! 😄

PR #641, opened 2020-11-03
Update documentation with cleaner way of passing certificates to hvac…

trixpan:
… via requests.Session()

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -8,23 +8,23 @@ Advanced Usage
 Making Use of Private CA
 ------------------------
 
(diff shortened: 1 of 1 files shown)


trixpan:
By debugging an unrelated issue I noticed we also expose the `verify=` as part of hvac.Client() let me know your preferred path I and will be happy to adjust

jeffwecan (approved):
Thanks!

jeffwecan:
re:

> By debugging an unrelated issue I noticed we also expose the verify= as part of hvac.Client() let me know your preferred path I and will be happy to adjust

Longer term, I think it makes sense to avoid duplicating requests parameters on our Client class (e.g., like `verify`) and instead use dependency injection-esque things like you have documented with the `session` instance here. So I'm good with this direction. 👍

PR #649, opened 2020-12-01
Enable response wrapping of PKI secrets

tmcolby:
Methods `generate_intermediate` and `generate_root`, if called with `type='exported'`, expose the private key in the response.  Additionally, `generate_certificate` also exposes the private key in the response.
This PR adds the `wrap_ttl` keyword argument to these methods that enables response wrapping of these calls.
https://www.vaultproject.io/docs/concepts/response-wrapping

### DIFF
--- a/hvac/api/secrets_engines/pki.py
+++ b/hvac/api/secrets_engines/pki.py
@@ -242,7 +242,7 @@ class Pki(VaultApiBase):
             url=api_path,
         )
 
-    def generate_intermediate(self, type, common_name, extra_params=None, mount_point=DEFAULT_MOUNT_POINT):
+    def generate_intermediate(self, type, common_name, extra_params=None, mount_point=DEFAULT_MOUNT_POINT, wrap_ttl=None):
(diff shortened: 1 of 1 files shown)


tmcolby:
hi, what is the cadence for approving PR's and official releases?  i want to use hvac in a large enterprise project, but i need the feature in this PR.  i will need to be able to pull a release from pypi using requirements.txt to satisfy the devops folks - i dont think they will like pulling from my repo and building.
so with that said - can you tell me the likely hood of this getting approved and a realistic sense if it can make it into a release?  im not asking for special attention - just trying to understand if i need to go another route.  thanks for any feedback.

jeffwecan:
@tmcolby: Sorry for the delay in getting to this PR and your query! We typically aim to create new releases more-or-less anytime PRs have been merged in. In reality, I haven't been able to make as much time for this project to hit that cadence (obviously :P). That said, I was planning on release a new version, with these changes included, over the next hour or so. 👍

PR #651, opened 2020-12-02
Template "auth method not implemented" error message

### LINKED ISSUE #650: Unimplemented auth method error message is not templated
https://github.com/hvac/hvac/blob/7a8642e6ddc793ff80c69273987657b3c664f0c2/hvac/api/vault_api_category.py#L42

The error string looks like it's supposed to insert the auth method name but doesn't, so you end up with the raw `%s` in the error.

briantist:
Fixes #650

Error message for unimplemented auth method doesn't include the auth method name, looks like a simple oversight.


jeffwecan (approved):
Thanks!

Issue #640, opened 2020-10-28
Online Documentation does not seem to work for lambda/EC2 


lukepmanion:
Spent quite a while trying to get this code snippet to work:
https://hvac.readthedocs.io/en/stable/usage/auth_methods/aws.html#lambda-and-or-ec2-instance

I believe it is missing the session token which I've updated below.
```

def infer_credentials_from_iam_role(iam_role):
    on_lambda = 'AWS_LAMBDA_FUNCTION_NAME' in os.environ
    if on_lambda:
        return os.environ['AWS_ACCESS_KEY_ID'], os.environ['AWS_SECRET_ACCESS_KEY'], os.environ['AWS_SESSION_TOKEN']
    else:
        security_credentials = load_aws_ec2_role_iam_credentials(iam_role)
        return security_credentials['AccessKeyId'], security_credentials['SecretAccessKey'], security_credentials['Token']



access_key_id, secret_access_key, token = infer_credentials_from_iam_role('vaultSecretAccess')
client.auth.aws.iam_login(access_key_id, secret_access_key, token)
```

PR #675, opened 2021-03-10
Fix lambda authentication example in aws auth documentation

JPoser:
The current lambda example code is missing a session token reference and doesn't work as highlighted in [issue 640](https://github.com/hvac/hvac/issues/640). With it everything works as expected although I'd be tempted to split out the ec2 and Lambda implementations as it I found it a little confusing.


jeffwecan (approved):
Thanks!

Issue #656, opened 2020-12-18
hvac 0.10.6 - approle auth method - generate_secret_id : issue with metadata type


xtof-osd:
Using HVAC 0.10.6 with Vault v1.5.2

The following instruction:
  `client.auth.approle.generate_secret_id(
                  rolename, metadata=metadata, mount_point=VAULT_APPROLE_PATH
              )`
fails with ParamValidationError (hvac/api/auth_methods/approle.py", line 244) if metadata is not a dict.

BUT, the requests to Vault fails if metadata is a dict: 
`error converting input map[target:development] for field "metadata": '' expected type 'string', got unconvertible type 'map[string]interface {}'
, on post https://vault.certificare.vin:8200/v1/auth/apptoken/role/somerole/secret-id
`
If you change the check line 244 in approle.py to verify if metadat is str and not dict, everything works OK.

lperdereau:
Hello @jeffwecan, I'm trying to find a way for this issue on my repo (https://github.com/lperdereau/hvac/commit/096eb4d8302c99cdfa25bccece1002b63ed990e0)

Would you accept this type of PR ? I'll find a way to make test OK :). If you have another solution I'm open to it. And if you think there is no issue on this, could you says me how I add metadata to a secret ?

jeffwecan:
Yes a PR would be most welcome! ☺

lperdereau:
Nice I'll link on this issue

PR #680, opened 2021-04-02
Send AppRole generate_secret_id Method Metadata Parameter as String

lperdereau:
This PR respond to https://github.com/hvac/hvac/issues/656


jeffwecan (approved):
Thank you @lperdereau!

PR #684, opened 2021-04-06
AWS Secrets Engine: Add support for iam_tags when creating roles

nicholaswold:
See for more details:
- https://www.vaultproject.io/api/secret/aws#iam_tags
- https://github.com/hashicorp/vault/pull/10953
- https://github.com/hashicorp/vault/releases/tag/v1.7.0

This PR updates `create_or_update_role` in the AWS secrets engine code to accept an `iam_tags` parameter. This parameter assumes a list of strings in the pattern `key=value`.

### DIFF
--- a/hvac/api/secrets_engines/aws.py
+++ b/hvac/api/secrets_engines/aws.py
@@ -139,8 +139,7 @@ class Aws(VaultApiBase):
         )
 
     def create_or_update_role(self, name, credential_type, policy_document=None, default_sts_ttl=None, max_sts_ttl=None,
-                              role_arns=None, policy_arns=None, legacy_params=False, mount_point=DEFAULT_MOUNT_POINT,
-                              iam_tags=None):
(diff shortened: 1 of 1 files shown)


hvac/api/secrets_engines/aws.py
```diff
@@ -178,6 +179,9 @@ def create_or_update_role(self, name, credential_type, policy_document=None, def
         :type legacy_params: bool
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
+        :param iam_tags: A list of strings representing a key/value pair to be used for any IAM user that is created by
+            this role. Format is a key and value separated by an =.
+        :type iam_tags: list
```
jeffwecan, line 184:
```suggestion
        :param iam_tags: A list of strings representing a key/value pair to be used for any IAM user that is created by
            this role. Format is a key and value separated by an =.
        :type iam_tags: list
        :param mount_point: The "path" the method/backend was mounted on.
        :type mount_point: str | unicode
```

jeffwecan (dismissed):
Thanks @nicholaswold!

PR #688, opened 2021-04-21
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10

### LINKED ISSUE #687: threading.Thread.setDaemon has been deprecated in favor of setting daemon attribute directly in Python 3.10
Ref : python/cpython#25174

```
rg -t py -w 'currentThread|notifyAll|activeCount|isDaemon|setDaemon' 
tests/integration_tests/api/auth_methods/test_github.py
35:            cls.mock_server_thread.setDaemon(True)
```

tirkarthi:
Fixes #687


jeffwecan (approved):
Sweet, thanks @tirkarthi!

PR #689, opened 2021-04-27
Add handling of unsupported HTTP methods inside adapter

### LINKED ISSUE #551: Adapters list function
Hi,

I tried to use `secrets.kv.v2.list_secrets` wich use `adapters.list` function. But this function fails with the following error:

`hvac.exceptions.InvalidPath: None`

If I understand your code it calls `session.request` from `requests` module with `'list'` method. 
I'm a little bit confused by this because first of all I wasn't aware of such method for HTTP protocol, and if I turn on the debugger in my web browser, the method used is `'get'`.

Regards

sshishov:
### Background
We are using Kubernetes in our setup and Istio Gateway which is using Envoy behind the scenes.
- Unfortunately, envoy does not support custom HTTP methods, according to [the issue](https://github.com/envoyproxy/envoy/issues/8247).
- When [this issue](https://github.com/envoyproxy/envoy/issues/5155) will be closed, the parser will allow custom HTTP verbs but currently, we are locked out.

The only way to have a workaround, to replace `list` custom verb with `get` plus additional params.

### NOTE
Additional actions:
- now we can always use `LIST` everywhere and if the adapter does not support it, it will be converted before the request.
- there are a lot of places with logic if method == 'LIST' etc... And there are a lot of places where we are propagating list=true to params.

Fixes #551

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -18,7 +18,7 @@ class Adapter(object):
     __metaclass__ = ABCMeta
 
     def __init__(self, base_uri=DEFAULT_BASE_URI, token=None, cert=None, verify=True, timeout=30, proxies=None,
-                 allow_redirects=True, session=None, namespace=None, ignore_exceptions=False):
+                 allow_redirects=True, session=None, namespace=None, ignore_exceptions=False, strict_http=False):
         """Create a new request adapter instance.
 
         :param base_uri: Base URL for the Vault instance being addressed.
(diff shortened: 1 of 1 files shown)


Eslamanwar (approved):

sshishov:
If needed, I can try to update the code everywhere where this addition of `list=true` is used.
I just do not know if every endpoint which is using it supports `LIST` method of HTTP.
If not, then this inconsistency should be fixed, imho.

hvac/adapters.py
```diff
@@ -284,6 +287,14 @@ def request(self, method, url, headers=None, raise_exception=True, **kwargs):
         _kwargs = self._kwargs.copy()
         _kwargs.update(kwargs)
 
+        if self.strict_http and method.lower() in ('list',):
+            # Encty point for standard HTTP substitution
```
jeffwecan, line 291:
```suggestion
            # Entry point for standard HTTP substitution
```

PR #691, opened 2021-04-29
Add Certificate Authentication Methods

Tylerlhess:
Added Certificate Authentication

### DIFF
--- a/hvac/api/auth_methods/cert.py
+++ b/hvac/api/auth_methods/cert.py
@@ -69,26 +69,29 @@ class Cert(VaultApiBase):
                 cert = f_cert.read()
         except FileNotFoundError:
             cert = certificate
-        params = {
-            'name': name,
-            'certificate': cert,
-            'allowed_common_names': allowed_common_names,
-            'allowed_dns_sans': allowed_dns_sans,
-            'allowed_email_sans': allowed_email_sans,
-            'allowed_uri_sans': allowed_uri_sans,
-            'allowed_organizational_units': allowed_organizational_units,
-            'required_extensions': required_extensions,
-            'display_name': display_name,
-            'token_ttl': token_ttl,
-            'token_max_ttl': token_max_ttl,
-            'token_policies': token_policies,
-            'token_bound_cidrs': token_bound_cidrs,
-            'token_explicit_max_ttl': token_explicit_max_ttl,
-            'token_no_default_policy': token_no_default_policy,
-            'token_num_uses': token_num_uses,
-            'token_period': token_period,
-            'token_type': token_type
-        }
+
+        params = utils.remove_nones(
+            {
+                "name": name,
+                "certificate": cert,
+                "allowed_common_names": allowed_common_names,
+                "allowed_dns_sans": allowed_dns_sans,
+                "allowed_email_sans": allowed_email_sans,
+                "allowed_uri_sans": allowed_uri_sans,
+                "allowed_organizational_units": allowed_organizational_units,
+                "required_extensions": required_extensions,
+                "display_name": display_name,
+                "token_ttl": token_ttl,
+                "token_max_ttl": token_max_ttl,
+                "token_policies": token_policies,
+                "token_bound_cidrs": token_bound_cidrs,
+                "token_explicit_max_ttl": token_explicit_max_ttl,
+                "token_no_default_policy": token_no_default_policy,
+                "token_num_uses": token_num_uses,
+                "token_period": token_period,
+                "token_type": token_type,
+            }
+        )
 
         api_path = '/v1/auth/{mount_point}/certs/{name}'.format(mount_point=mount_point, name=name)
         return self._adapter.post(
@@ -249,4 +252,3 @@ class Cert(VaultApiBase):
 
     class CertificateAuthError(Exception):
         pass
-
--- a/hvac/api/auth_methods/cert.py
+++ b/hvac/api/auth_methods/cert.py
@@ -1,0 +1,252 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Cert methods module."""
+from hvac.api.vault_api_base import VaultApiBase
+from hvac.utils import validate_pem_format
+from hvac import exceptions
+import os
+
+
+class Cert(VaultApiBase):
+    """Cert Auth Method (API).
+
+    Reference: https://www.vaultproject.io/api/auth/cert/index.html
+    """
+
+    def create_ca_certificate_role(self, name, certificate, allowed_common_names="", allowed_dns_sans="", allowed_email_sans="",
+                  allowed_uri_sans="", allowed_organizational_units="", required_extensions="", display_name="", token_ttl=0,
+                  token_max_ttl=0, token_policies=[], token_bound_cidrs=[], token_explicit_max_ttl=0,
+                  token_no_default_policy=False, token_num_uses=0, token_period=0, token_type="", mount_point="cert"):
+        """
+        Create CA Certificate Role
+        Sets a CA cert and associated parameters in a role name.
+
+        Supported methods:
+            POST:	/auth/<mount point>/certs/:name. Produces: 204 (empty body)
+
+        »Parameters
+        :param name: The name of the certificate role.
+        :type name: str
+        :param certificate: The PEM-format CA certificate.
+        :type certificate: str
+        :param allowed_common_names: Constrain the Common Names in the client certificate with a globbed pattern. Value is a comma-separated list of patterns. Authentication requires at least one Name matching at least one pattern. If not set, defaults to allowing all names.
+        :type allowed_common_names: str | list
+        :param allowed_dns_sans: Constrain the Alternative Names in the client certificate with a globbed pattern. Value is a comma-separated list of patterns. Authentication requires at least one DNS matching at least one pattern. If not set, defaults to allowing all dns.
+        :type allowed_dns_sans: str | list
+        :param allowed_email_sans: Constrain the Alternative Names in the client certificate with a globbed pattern. Value is a comma-separated list of patterns. Authentication requires at least one Email matching at least one pattern. If not set, defaults to allowing all emails.
+        :type allowed_email_sans: str | list
+        :param allowed_uri_sans: Constrain the Alternative Names in the client certificate with a globbed pattern. Value is a comma-separated list of URI patterns. Authentication requires at least one URI matching at least one pattern. If not set, defaults to allowing all URIs.
+        :type allowed_uri_sans: str | list
+        :param allowed_organizational_units: Constrain the Organizational Units (OU) in the client certificate with a globbed pattern. Value is a comma-separated list of OU patterns. Authentication requires at least one OU matching at least one pattern. If not set, defaults to allowing all OUs.
+        :type allowed_organizational_units: str | list
+        :param required_extensions: Require specific Custom Extension OIDs to exist and match the pattern. Value is a comma separated string or array of oid:value. Expects the extension value to be some type of ASN1 encoded string. All conditions must be met. Supports globbing on value.
+        :type required_extensions: str | list
+        :param display_name: The display_name to set on tokens issued when authenticating against this CA certificate. If not set, defaults to the name of the role.
+        :type display_name: str | unicode
+        :param token_ttl: The incremental lifetime for generated tokens. This current value of this will be referenced at renewal time.
+        :type token_ttl: int | str
+        :param token_max_ttl: The maximum lifetime for generated tokens. This current value of this will be referenced at renewal time.
+        :type token_max_ttl: int | str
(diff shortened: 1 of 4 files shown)


jeffwecan:
Thanks for the contribution @Tylerlhess! Do you have any interest in attempting to include some [documentation](https://github.com/hvac/hvac/blob/develop/CONTRIBUTING.md#documentation) and/or [test coverage](https://github.com/hvac/hvac/blob/develop/CONTRIBUTING.md#testing) to cover this newly added TLS certificates auth method class? 😃 

(If not, that is perfectly okay, it may just take me a bit to see about filling into those aspects before I'd feel ready to merge this in.)

Tylerlhess:
Oh I figured I needed to write the test cases myself. I just need to get the adapter to give me the cert and it should be passing

Tylerlhess:
Helps to actually write tests and test code before submitting the PR. :)

hvac/v1/__init__.py
```diff
@@ -592,7 +592,7 @@ def auth_tls(self, mount_point='cert', use_token=True):
         :return:
         :rtype:
         """
-        return self.login('/v1/auth/{0}/login'.format(mount_point), use_token=use_token)
+        return self.auth.cert.login(mount_point=mount_point, use_token=use_token)
```
jeffwecan, line 595: Since we're updating this extant call site to use the new class, could you please tack on a related decorator for this `auth_tls()` method? E.g.:

```python
    @utils.deprecated_method(
        to_be_removed_in_version='0.13.0',
        new_method=api.auth_methods.Cert.login,
    )
```

(This `Client` class is a bit unwieldy, so we've been taking any opportunity to move folks off of using its methods where possible. Related reference: https://github.com/hvac/hvac/blob/develop/CONTRIBUTING.md#backwards-compatibility-breaking-changes )

jeffwecan (approved):
Thanks so much for working to fill in some test cases here! Looking things over, there is just one bit of cleanup I'd be in favor of seeing before we merge -> release this.

jeffwecan:
Ah shucks, looks like implementing that last suggestion may have exposed some gaps in our little `deprecated_method` handler 😂 :

```
_________ ERROR collecting tests/unit_tests/v1/test_aws_ec2_methods.py _________
tests/unit_tests/v1/test_aws_ec2_methods.py:6: in <module>
    from hvac import Client
hvac/__init__.py:1: in <module>
    from hvac.v1 import Client
hvac/v1/__init__.py:18: in <module>
    class Client(object):
hvac/v1/__init__.py:589: in Client
    def auth_tls(self, mount_point='cert', use_token=True):
hvac/utils.py:188: in decorator
    new_docstring=dedent(new_method.__doc__),
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/textwrap.py:430: in dedent
    text = _whitespace_only_re.sub('', text)
E   TypeError: expected string or bytes-like object
```

My intuition is this is due to some logic related to setting `__doc__` on the decorated method combined with the `Cert.login()` method not having any docstrings? I'll scope that out to confirm and get back to ya!

jeffwecan:
Alrighty, so I'll merge in https://github.com/hvac/hvac/pull/692 shortly and you could then rebase this branch on top of `develop` / merge develop into your branch to get those unrelated-to-your-changes test failures cleared.

Tylerlhess:
The Issue was no doc string in the new method replacing the deprecated method. More of a bug in my code than in the decorator

Tylerlhess:
Yeah hold off on merging this I might have to add setting the cert and ca directly into the sessions request. I also discovered that there may need to be a disclaimer that this will not work with the agent at present as it currently doesn't proxy the certificate request as far as I can see it would also need to be running in ssl. The agent can be used after certificate auth with the token. (This is what I had to do in my production environment as well as make the initial request through requests directly instead of being able to set the session variables needed to get the cert auth to work ) The current auth_tls method also has this failing currently and balks at passing the cacert kwarg down to the adapter even as the 'verify' kwarg. This is just the lack of passing kwargs down through to the adapter layer but I also want to make sure that my code is doing so properly. 
I am setting the kwargs to the adapter in my new_login method which should resolve the adapter getting these kwargs but I may also need to do this in the general login method. When I submitted the or I was under the impression that this was happening but it doesn't seem to be the case.

I don't think I will get the agent resolved in this pr as my experience with the agent is lacking

jeffwecan:
I don't have much context on cert auth usage but I'm down to help scope out the agent angle as I have time over the next week or so.

Tylerlhess:
OK this is now fully baked and ready to handle all the things.

hvac/api/auth_methods/cert.py
```diff
@@ -0,0 +1,252 @@
...
+            'token_ttl': token_ttl,
+            'token_max_ttl': token_max_ttl,
+            'token_policies': token_policies,
+            'token_bound_cidrs': token_bound_cidrs,
+            'token_explicit_max_ttl': token_explicit_max_ttl,
+            'token_no_default_policy': token_no_default_policy,
+            'token_num_uses': token_num_uses,
+            'token_period': token_period,
+            'token_type': token_type
+        }
```
jeffwecan, line 91: I think you'll want to wrap this in `utils.remove_nones()` to match similar methods' behavior (without that arrangement, it makes it harder for hvac users to update a single parameter without reverting the rest of them to their defaults generally):
```suggestion

        params = utils.remove_nones(
            {
                "name": name,
                "certificate": cert,
                "allowed_common_names": allowed_common_names,
                "allowed_dns_sans": allowed_dns_sans,
                "allowed_email_sans": allowed_email_sans,
                "allowed_uri_sans": allowed_uri_sans,
                "allowed_organizational_units": allowed_organizational_units,
                "required_extensions": required_extensions,
                "display_name": display_name,
                "token_ttl": token_ttl,
                "token_max_ttl": token_max_ttl,
                "token_policies": token_policies,
                "token_bound_cidrs": token_bound_cidrs,
                "token_explicit_max_ttl": token_explicit_max_ttl,
                "token_no_default_policy": token_no_default_policy,
                "token_num_uses": token_num_uses,
                "token_period": token_period,
                "token_type": token_type,
            }
        )
```

Also would need to add `from hvac import utils` to this file's imports to support that ^

jeffwecan (approved):
Looks good to me! Just had that one minor point / suggestion on the `create_ca_certificate_role()` method's params.

jeffwecan (approved):
Awesome, thanks again @Tylerlhess! 🎉

jeffwecan:
FYI I'll aim to get a release including these additions shipped over the next day or so.

HansAarneLiblik:
Since `hvac.Client` already accepts `cert` parameter, why is it needed to provide it to `auth.cert.login()` method again ? Can't it be retrieved from `hvac.Client` ?

hvac/api/auth_methods/cert.py
```diff
@@ -0,0 +1,254 @@
...
+            if not self._adapter._kwargs.get('verify'):
+                raise self.CertificateAuthError("cacert must be True, a file_path, or valid CA Certificate.")
+            else:
+                cacert = self._adapter._kwargs.get('verify')
+        else:
+            validate_pem_format(cacert, "verify")
+        # if cert_pem is a string its ready to be used and either has the key with it or the key is provided as an arg
+        try:
+            if validate_pem_format(cert_pem, "cert_pem"):
+                tls_update = True
```
intgr, line 228: Erm, this code is unreachable. `validate_pem_format()` is *documented* as returning `bool`, but it actually always returns `None`.

hvac/api/auth_methods/cert.py
```diff
@@ -0,0 +1,254 @@
...
+        except exceptions.VaultError as e:
+            if isinstance(e, type(exceptions.ParamValidationError())):
+                tls_update = {}
+                if not (os.path.exists(cert_pem) or self._adapter._kwargs.get('cert')):
+                    raise FileNotFoundError("Can't find the certificate.")
+                try:
+                    for tls_part, value in {'cert_pem': cert_pem, 'key_pem': key_pem}:
```
intgr, line 235: Iterating over a dict only yields keys, not key-value pairs. So this will always raise `ValueError` and the loop body is unreachable code.
```python
>>> for tls_part, value in {'cert_pem': '', 'key_pem': ''}:
...     pass

ValueError: too many values to unpack (expected 2)
```

hvac/api/auth_methods/cert.py
```diff
@@ -0,0 +1,254 @@
...
+            else:
+                cacert = self._adapter._kwargs.get('verify')
+        else:
+            validate_pem_format(cacert, "verify")
+        # if cert_pem is a string its ready to be used and either has the key with it or the key is provided as an arg
+        try:
+            if validate_pem_format(cert_pem, "cert_pem"):
+                tls_update = True
+        except exceptions.VaultError as e:
+            if isinstance(e, type(exceptions.ParamValidationError())):
```
intgr, line 230: `isinstance(e, type(exceptions.ParamValidationError()))`
->
`isinstance(e, exceptions.ParamValidationError)`
->
`except exceptions.ParamValidationError as e`

hvac/api/auth_methods/cert.py
intgr: Note that the requests library does not accept a PEM string directly s argument -- it must always be a path. So this logic could never have worked anyway.

hvac/api/auth_methods/cert.py
```diff
@@ -0,0 +1,254 @@
...
+        :param key_pem: Location of the public key.pem used to authenticate the host.
+        :param key_pem: str | unicode
+        :param mount_point:
+        :type mount_point:
+        :param use_token: If the returned token is stored in the client
+        :param use_token: bool
+        :return: The response of the login request.
+        :rtype: requests.Response
+        """
+        params = {'use_token': use_token}
```
intgr, line 210: Fixed in https://github.com/hvac/hvac/pull/720

hvac/api/auth_methods/cert.py
```diff
@@ -0,0 +1,254 @@
...
+                **self._adapter._kwargs,
+                **{'verify': cacert, 'cert': tuple([cert_pem, key_pem])} # need to define dict as cert is a tuple
+            }
+
+        return self._adapter.login(
+            url=api_path,
+            json=params,
+        )
+
+    class CertificateAuthError(Exception):
```
intgr, line 253: Shouldn't hvac exceptions derive from `exceptions.VaultError` class?

hvac/api/auth_methods/cert.py
xupefei: The signature of `validate_pem_format` is `(param_name, param_argument)` but here it's being called like `(param_argument, param_name)` This validation will never pass :(

hvac/api/auth_methods/cert.py
Tylerlhess: Good catch. The use_token was one of the things I hadn't originally had in scope as I wasn't originally planning on depricating the auth_tls method. I was originally only trying to create a method to create the certificate endpoints as well as allow an already created client to auth without the need to recreate it. This is because the certs are not saved as part of the adapter class and thus could only be used during instantiation of a new client object.

hvac/api/auth_methods/cert.py
Tylerlhess: It does not let you subclass Exceptions in this way. It has similar patterns in other modules with module specific error messages.

hvac/api/auth_methods/cert.py
Tylerlhess: Yes, this was from a miss reading of the requests verify params

:param verify: Either a boolean, in which case it controls whether we verify
            the server's TLS certificate, or a string, in which case it must be a path
            to a CA bundle to use

was read as 

:param verify: Either a boolean, in which case it controls whether we verify,
            the server's TLS certificate, or a string, in which case it must be a path
            to a CA bundle to use
            
I could never figure out why it thought that you could supply the tls cert as an argument but this logic was left in as it wasn't breaking. It wasn't working either.

as @xupefei pointed out I did also transpose the two arguments so this always failed.

hvac/api/auth_methods/cert.py
Tylerlhess: fixed in #743

hvac/api/auth_methods/cert.py
```diff
@@ -0,0 +1,254 @@
...
+        :param token_period: The period, if any, to set on the token.
+        :type token_period: int | str
+        :param token_type: The type of token that should be generated. Can be service, batch, or default to use the mount's tuned default (which unless changed will be service tokens). For token store roles, there are two additional possibilities: default-service and default-batch which specify the type to return unless the client requests a different type at generation time.
+        :type token_type: str
+        :param mount_point:
+        :type mount_point:
+        """
+        try:
+            with open(certificate, 'r') as f_cert:
+                cert = f_cert.read()
```
intgr, line 69: Reported bug #841 related to this code.

PR #701, opened 2021-05-14
Update approle.rst

mblau-leaffilter:
Under the function for generating the secret id the f-string incorrectly points to role id. This patch fixes it to correctly point to ['data']['secret_id']


jeffwecan (approved):
Thanks!

PR #720, opened 2021-07-01
Fix Cert.login() handling of use_token argument

intgr:
The `client.auth.cert.login(use_token=True)` argument should be passed internally to `Adapter.login()`. But instead it was sent to Vault in the POST body, which is not recognized by Vault: https://www.vaultproject.io/api/auth/cert#login-with-tls-certificate-method

Regression in #691 (7bb92f72da5968b4fc08f7b6434fdad631dd2f2d).


intgr:
Ping? @jeffwecan?

jeffwecan (approved):
Good catch, thanks @intgr!

PR #739, opened 2021-07-12
Drop Python 2.7 Support Explicitly

### LINKED ISSUE #582: Drop Python2 support
I understand that it may be too early to actually drop Python 2 support, because still ~ 40% of downloads come from Python 2 (according to [PyPI stats](https://pypistats.org/packages/hvac)). But it might be a good idea to announce plans in advance and tell users which version of `hvac` will be the latest with Python 2 support. Does the `hvac` project have any plan on this matter?

jeffwecan:
closes #582 (🤞🏻) 🎉 

TODO's:

- [ ] Validate expected behavior installing the module in both Python 2.7 and 3.x using test.pypi.org before shipping this!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,4 +3,4 @@ description = README.md
 license_files = LICENSE.txt
 
 [bdist_wheel]
-universal = 1
+universal = 0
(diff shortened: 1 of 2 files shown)


Kludex:
Is the TODO point the only thing missing? Can I help somehow?

hugovk:
This can be closed: https://github.com/hvac/hvac/issues/582 was fixed by https://github.com/hvac/hvac/pull/819.

PR #745, opened 2021-07-22
Fix Client TLS `verify` Behavior 

bpatterson971:
Currently, the HVAC client defaults the 'verify' parameter to True.  Later, in the init method, if 'verify' is not None, then it will be overridden by environmental variables, if present.  If a user specifically sets a value of 'verify' (say to False), the if block that overrides the parameter is ran clobbering whatever the user specified when creating the client.

The comments indicate that this if block should only run if the user did not specify anything to verify.  This pull will default verify to None and update the if block to only run if verify is set to None.  If no environmental variables are available, then 'verify' will get set to True.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -31,7 +31,7 @@ class Client(object):
         url=None,
         token=None,
         cert=None,
-        verify=True,
+        verify=None,
         timeout=30,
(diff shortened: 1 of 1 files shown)


Tylerlhess:
Good catch. Thanks for fixing my code :)

PR #762, opened 2021-09-22
Add X-Vault-Request header to all requests by default

### LINKED ISSUE #761: Add option to add X-Vault-Request header to all requests
Some Vault agent listeners may be configured with the `require_request_header` [option](https://www.vaultproject.io/docs/agent#listener-stanza) which will reject all requests made to the server without the `X-Vault-Request: true` header set.

The official Go client also adds this header to all requests [by default](https://github.com/hashicorp/vault/blob/master/api/client.go#L459), regardless of whether requests are issued to a Vault Server or Vault Agent.

While it's possible to add custom headers to requests made from `hvac` using an [HTTP Adapter](https://hvac.readthedocs.io/en/stable/advanced_usage.html#custom-requests-http-adapter) perhaps it would be ideal to add a `use_request_header` option or similar to the `Client`'s constructor which enables this behavior (if it is not to be already enabled by default)? This may make the library easier to use, and otherwise encourage the usage of secure defaults.

bobmshannon:
Closes #761.

This change is intended to update the behavior of `hvac` to match the [Go client](https://github.com/hashicorp/vault/blob/main/api/client.go#L479) where the `X-Vault-Request: true` header is added to all requests regardless of whether they are made to a Vault Server or Vault Agent.

This header is used for providing protection against SSRF vulnerabilities. Since adding it should be backwards compatible, I've enabled it by default. This will match the Go client behavior and also ensure that `hvac` uses secure defaults where possible.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -30,6 +30,7 @@ class Adapter(object):
         namespace=None,
         ignore_exceptions=False,
         strict_http=False,
+        request_header=True,
     ):
         """Create a new request adapter instance.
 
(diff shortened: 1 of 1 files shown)


bobmshannon:
@jeffwecan it would be much appreciated if you can take a look at this whenever you get a chance. :)

PR #767, opened 2021-09-23
Re-add six to requirements

### LINKED ISSUE #765: v0.11.1 won't load, six is no longer in setup.py but is used in utils.py
Release v0.11.1 commit c9ff353 is labeled "Drop six from requirements" and removed six from requirements-dev.txt, requirements.in, requirements.txt, and setup.py.  However, [line 11 in hvac/utils.py](https://github.com/hvac/hvac/blob/0e78da45711f6d30e924e95a38e456c5fd104eec/hvac/utils.py#L11)  still calls `import six`, causing hvac to fail to load.

bobmshannon:
Closes #765.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,2 +1,3 @@
 pyhcl
 requests
+six
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,5 +14,7 @@ pyhcl==0.4.4
     # via -r requirements.in
 requests==2.25.1
     # via -r requirements.in
+six==1.16.0
+    # via -r requirements.in
 urllib3==1.26.6
     # via requests
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ setup(
     python_requires=">=2.7",
     install_requires=[
         "requests>=2.25.1",
+        "six>=1.5.0",
     ],
     include_package_data=True,
     package_data={"hvac": ["version"]},


ayav09:
I believe v1.0.0 is the target to drop python2 support so we should be adding six back to the requirements.

bobmshannon:
I'm not sure what the plan there is, I'll let @jeffwecan chime in for more details. Happy to make any changes as needed, just trying to unbreak `0.11.1` for those trying to upgrade it.

bobmshannon:
Ok yeah, it does seem based on the changelog `0.11.1` supports Python 2, not sure why `six` was dropped but I'll re-add it.

briantist:
The change to `requests` also broke stuff by way of other (unrelated) packages needing a version that conflicts... imo this shouldn't have been done in a patch release.

jeffwecan:
Yeah that was my bad y'all. My sincere apologies. Will cut a quick break fix / 0.11.2 release here shortly

jeffwecan:
I am going to go ahead and just do a full revert of the related offending (and unintentionally released 😞 🙃 ) commits though. So closing this PR and will proceed with https://github.com/hvac/hvac/pull/768 instead

briantist:
thanks for this @jeffwecan , out of curiosity for 1.0.0 too, what is needed in the newer `requests` version?

jeffwecan:
@briantist: I'm not sure quite yet. My intention was to validate that sort of thing before the v1 release and I got a bit ahead of myself merging it into our main integration branch here before performing that due diligence. Ultimately, the main motivation was to get `requests`'s implicit `urllib3` dependency to >=1.26.5. Which I supposes implies a [requests version >=2.25.0](https://github.com/psf/requests/blob/fc106ab586b9bfeca37c54780387b24b5d2b13b4/HISTORY.md#2250-2020-11-11)?

In any case, I'll move those musing to an issue for follow-up: https://github.com/hvac/hvac/issues/770

Also, if you can expand a bit on the "broke stuff by way of other (unrelated) packages needing a version that conflicts"  note with any example of impacted packages I would appreciate having that extra context. 🙏🏻

PR #771, opened 2021-09-27
Fix premature read on stream requests in the `sys.take_raft_snapshot` method

dekimsey:
When a caller (such as `sys.take_raft_snapshot`) performs a stream
request, the act of attempting to parse the response as JSON causes the
entire response body to be read and the underlying connection to be
closed. This renders the streaming response moot.

This change addresses that issue by examining if the caller requested a
stream response, and if so returning the response as is.

Without the change, it is impossible to read raft snapshots that are
larger than memory as the entire response is read into memory to attempt
to read as JSON.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -370,6 +370,10 @@ class JSONAdapter(RawAdapter):
         :rtype: dict | requests.Response
         """
         response = super().request(*args, **kwargs)
+        # Do not attempt to jsonify stream requests, doing so will empty the
+        # socket and the caller has explicitly opted out
+        if kwargs.get('stream', False):
+            return response
         if response.status_code == 200:
             try:
                 return response.json()
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -397,10 +397,6 @@ class JSONAdapter(RawAdapter):
         :rtype: dict | requests.Response
         """
         response = super().request(*args, **kwargs)
-        # Do not attempt to jsonify stream requests, doing so will empty the
-        # socket and the caller has explicitly opted out
-        if kwargs.get('stream', False):
-            return response
         if response.status_code == 200:
             try:
                 return response.json()
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -14,6 +14,33 @@ from hvac.constants.client import DEFAULT_URL
 class Adapter(metaclass=ABCMeta):
     """Abstract base class used when constructing adapters for use with the Client class."""
 
+    @classmethod
+    def from_adapter(
+        cls,
+        adapter,
+    ):
+        """Create a new adapter based on an existing Adapter instance.
+        This can be used to create a new type of adapter that inherits the properties of an existing one.
+
+        :param adapter: The existing Adapter instance.
+        :type adapter: hvac.Adapters.Adapter
+        """
+
+        return cls(
+            base_uri=adapter.base_uri,
+            token=adapter.token,
+            cert=adapter._kwargs.get("cert"),
+            verify=adapter._kwargs.get("verify"),
+            timeout=adapter._kwargs.get("timeout"),
+            proxies=adapter._kwargs.get("proxies"),
+            allow_redirects=adapter.allow_redirects,
+            session=adapter.session,
+            namespace=adapter.namespace,
+            ignore_exceptions=adapter.ignore_exceptions,
+            strict_http=adapter.strict_http,
+            request_header=adapter.request_header,
+        )
+
     def __init__(
         self,
         base_uri=DEFAULT_URL,
--- a/hvac/api/system_backend/raft.py
+++ b/hvac/api/system_backend/raft.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """Raft methods module."""
 from hvac.api.system_backend.system_backend_mixin import SystemBackendMixin
-from hvac import utils
+from hvac import utils, adapters
 
 
 class Raft(SystemBackendMixin):
@@ -102,14 +102,17 @@ class Raft(SystemBackendMixin):
 
         The snapshot is returned as binary data and should be redirected to a file.
 
+        This endpoint will ignore your chosen adapter and always uses a RawAdapter.
+
         Supported methods:
             GET: /sys/storage/raft/snapshot.
 
-        :return: The response of the s request.
+        :return: The response of the snapshot request.
         :rtype: requests.Response
         """
         api_path = "/v1/sys/storage/raft/snapshot"
-        return self._adapter.get(
+        raw_adapter = adapters.RawAdapter.from_adapter(self._adapter)
+        return raw_adapter.get(
             url=api_path,
             stream=True,
         )
--- a/tests/unit_tests/test_adapters.py
+++ b/tests/unit_tests/test_adapters.py
@@ -12,6 +12,52 @@ from hvac import exceptions
 from hvac import adapters
 
 
+class TestAdapters:
+    CONSTRUCTOR_ARGS = (
+        "base_uri",
+        "token",
+        "cert",
+        "verify",
+        "timeout",
+        "proxies",
+        "allow_redirects",
+        "session",
+        "namespace",
+        "ignore_exceptions",
+        "strict_http",
+        "request_header",
+    )
+
+    INTERNAL_KWARGS = (
+        "cert",
+        "verify",
+        "timeout",
+        "proxies",
+    )
+
+    @pytest.mark.parametrize(
+        "conargs",
+        [
+            {arg: arg.capitalize() for arg in CONSTRUCTOR_ARGS},
+            {arg: arg.upper() for arg in CONSTRUCTOR_ARGS},
+        ],
+    )
+    def test_from_adapter(self, conargs):
+        expected = conargs.copy()
+        for internal_kwarg in self.INTERNAL_KWARGS:
+            expected.setdefault("_kwargs", {})[internal_kwarg] = expected.pop(
+                internal_kwarg
+            )
+
+        # let's start with a JSONAdapter, and make a RawAdapter out of it
+        json_adapter = adapters.JSONAdapter(**conargs)
+
+        raw_adapter = adapters.RawAdapter.from_adapter(json_adapter)
+
+        for property, value in expected.items():
+            assert getattr(raw_adapter, property) == value
+
+
 class TestRequest(TestCase):
     """Unit tests providing coverage for requests-related methods in the hvac Client class."""
 


hvac/adapters.py
```diff
@@ -362,6 +362,10 @@ def request(self, *args, **kwargs):
         :rtype: dict | requests.Response
         """
         response = super(JSONAdapter, self).request(*args, **kwargs)
+        # Do not attempt to jsonify stream requests, doing so will empty the
+        # socket and the caller has explicitly opted out
+        if kwargs.get('stream', False):
+            return response
         if response.status_code == 200:
```
dekimsey, line 369: FWIW, I'm of the opinion this check should instead only attempt to JSON() the response iif the content-type is `application/json`.  It seemed out of scope for this particular change, so I didn't modify it.

Something like:
```
if response.ok  and 'application/json' in response.headers.get('content-type', ''):
    ...
```

I do see some API responses are documented to 204 status and an empty body though they clearly return data, it's unclear to me what the intent is. So I'm leaving this alone.

briantist: https://www.vaultproject.io/api#http-status-codes says the following:

> Applications should be prepared to accept both 200 and 204 as success. 204 is simply an indication that there is no response body to parse, but API endpoints that indicate that they return a 204 may return a 200 if warnings are generated during the operation.

So it seems that 204 should be a solid indication there is no body, while if there is a body (due to warnings or other circumstances) then the code should be 200.

If you've seen instances of that not being the case, it would be good to know.

dekimsey: Ooh nice @briantist, I was unaware of that!

briantist:
Hi @dekimsey ,

Re: the PR, as I've taken a  more active role in maintaining the project and become more aware of its inner workings, I've found a lot of issues with the Adapter system that I think are going to require large structural changes.

This change makes some sense, but in a way it breaks the (already tenuous) contract of the `JSONAdapter` in that it's supposed to return objects, but in "stream" mode it would not do that.

The change _could_ be made "upstream" in the `RequestsAdapter` where there is no such guarantee, and I think it would make more sense implemented there.

It would still do the same thing in that `JSONAdapter` will pass the `kwargs` along, which is still kind of unfortunate because it's not obvious unless someone looks at the implementation.

Anyway I am happy to discuss this change and other Adapter stuff more if you are still interested, I know it's been a while.
See also #898 for some stuff I was working on (but we will need even bigger changes than that imo).

dekimsey:
Hi @briantist, it has been awhile , but that's okay! I appreciate you updating the PR's target for me.

I'll try to look at this tomorrow and see if I understand where to put the change. At first glance, I'm not sure. It's not clear to me how I might obtain a RawAdapter object from [this point](https://github.com/hvac/hvac/blob/3b4e5ca5502cf01fde40cda51a36ea463f8f435c/hvac/api/system_backend/raft.py#L112) in the code.

But this is just from vague memory and I'm on mobile right now. I'll see if I can dig into this better tomorrow.

Thank you!

PS I did identify I think this issue was caused by #537, when adapters were changed.

briantist:
> Hi @briantist, it has been awhile , but that's okay! I appreciate you updating the PR's target for me.
> 
> I'll try to look at this tomorrow and see if I understand where to put the change. At first glance, I'm not sure. It's not clear to me how I might obtain a RawAdapter object from [this point](https://github.com/hvac/hvac/blob/3b4e5ca5502cf01fde40cda51a36ea463f8f435c/hvac/api/system_backend/raft.py#L112) in the code.

This is actually one of the things that can make the adapter system we have now a little confusing, and at times, restrictive.

At this point in the code, you _cannot_ choose which adapter is in use. The library is designed that a method such as this is supposedly agnostic to the adapter. But many methods then assume things about the response, based on what the project's default Adapter is.

But perhaps more to the point for this PR is that you don't have to worry about that in this case, my recommendation to move the code from `JSONAdapter` to `RawAdapter` is because it's more at home at that level, but `JSONAdapter` inherits from `RAWAdapter` so the result you're going for will be the same. This is why I say it's unfortunate, because the `JSONAdapter` in theory should probably not support returning streams (maybe it could support deserializing _from_ a stream?).

If you made this change in `JSONAdapter`, it would also not work for anyone who might be using `RAWAdapter` (for some reason) nor any other adapter based on `RAWAdatper`, so that's another reason I think it belongs there, if anywhere.

Not sure if I've helped with this explanation or made it more confusing 😅 , happy to hash it out further.

The raft snapshot endpoint is an interesting one for two reasons:
1. Since it forces a streamed response, the default adapter being `JSONAdapter` doesn't work (added in #537 as you found), and the entire idea of using a passed in Adapter doesn't work well for this method unless there's a way for Adapters to assert stream support. This is something I will want to address in #898 and other work I think we need to do.
2. I cannot find this method listed in Vault API documentation anymore, I only see [`sys/storage/raft/snapshot-auto`](https://developer.hashicorp.com/vault/api-docs/system/storage/raftautosnapshots) even when I look at quite old Vault versions, do you have any ideas on that?
 
> But this is just from vague memory and I'm on mobile right now. I'll see if I can dig into this better tomorrow.
> 
> Thank you!
> 
> PS I did identify I think this issue was caused by #537, when adapters were changed.

No worries, thanks very much for submitting this. Stream support was not something I had my mind at all when thinking about the Adapter system recently so it's quite helpful.

dekimsey:
> I cannot find this method listed in Vault API documentation anymore, I only see [sys/storage/raft/snapshot-auto](https://developer.hashicorp.com/vault/api-docs/system/storage/raftautosnapshots) even when I look at quite old Vault versions, do you have any ideas on that?

It's [there](https://developer.hashicorp.com/vault/api-docs/system/storage/raft#take-a-snapshot-of-the-raft-cluster), it's just under a different link on the parent page. Makes it easy to miss

briantist:
> It's [here](https://developer.hashicorp.com/vault/api-docs/system/storage/raft#take-a-snapshot-of-the-raft-cluster), it's just under a different link on the parent page. Makes it easy to miss.
 🤦 oof I tried checking there before and still missed it, thank you!

> > If you made this change in JSONAdapter, it would also not work for anyone who might be using RAWAdapter (for some reason) nor any other adapter based on RAWAdatper, so that's another reason I think it belongs there, if anywhere.
> 
> I don't quite follow this, from my read RawAdapter passes kwargs through to requests and doesn't manipulate the response object at all. From my read, it already works as is.

Actually, you are completely right.

I am starting to think that any fix we could make right now in the Adapter system would not be great long term (due to the other issues with the system that need to be changed).

Because the need to return a stream is so rare (this is the only call I'm aware of that returns raw binary data, if you know of more, that would be good to know), I'm starting to warm up to the idea of having this one method _not_ use the `_adaper` set on the client, and to instead have this snapshot method always use a `RawAdapter`.

What do you think about that idea?

---

The "tricky" (annoying) part is having to pass in all the fields from the current adapter into the constructor of `RawAdapter` to create a new instance with the all the same fields preserved. 

```python
raw_adapter = hvac.adapters.RawAdapter(base_url=self._adapter.base_uri, token=self._adapter.token, ..., ..., ..., ...)
return raw_adapter.get(
    url=api_path,
    stream=True,
)
```

Although we could do that directly in the method like that, it would be prone to accidental breakage if parameters are changed in any adapters in the chain, so I'd prefer we add a classmethod to the `Adapter` base class that takes care of it, so that it's somewhat more obvious where the responsibility lies.
- https://github.com/hvac/hvac/blob/3b4e5ca5502cf01fde40cda51a36ea463f8f435c/hvac/adapters.py#L14

```python
class Adapter(metaclass=ABCMeta):
    def __init__(
        self,
        base_uri=DEFAULT_URL,
        token=None,
        cert=None,
        verify=True,
        timeout=30,
        proxies=None,
        allow_redirects=True,
        session=None,
        namespace=None,
        ignore_exceptions=False,
        strict_http=False,
        request_header=True,
    ):
        # real implementation here
        pass

    @classmethod
    def from_adapter(
        cls,
        adapter,
    ):
        return cls(
            base_uri=adapter.base_uri,
            token=adapter.token,
            cert=adapter.cert,
            verify=adapter.verify,
            timeout=adapter.timeout,
            proxies=adapter.proxies,
            allow_redirects=adapter.allow_redirects,
            session=adapter.session,
            namespace=adapter.namespace,
            ignore_exceptions=adapter.ignore_exceptions,
            strict_http=adapter.strict_http,
            request_header=adapter.request_header,
        )
```

And then this would make the actual update for raft snapshots look more like this:

```python
raw_adapter = hvac.adapters.RawAdapter.from_adapter(self._adapter)
return raw_adapter.get(
    url=api_path,
    stream=True,
)
```

---

Eager to hear your thoughts! If that sounds good, let me know how you'd like to proceed, I could put up a separate PR with the classmethod so that you're working off of that, or it can be implemented here as well, either by you or I could push up some commits, or let me know what works.

briantist:
Hey @dekimsey wondering if you've had a chance to look at the above. I think this is a viable fix for the issue and that we could get it out in the next release. I'm happy to implement it myself, but I appreciate your report and patience, and if you're still interested in contributing to `hvac` I want to give you the opportunity to do that. 

We're [actively looking for new contributors](https://github.com/hvac/hvac/discussions/994) but I also understand you posted this a while ago and may not be interested or able to do that now. Either way I thank you!

briantist:
@dekimsey I've gone ahead and implemented the changes I proposed, but pushed to this PR, hope that's ok. I would appreciate if you have a few minutes to check it out. I'd be especially grateful if you're able to actually test the raft snapshot method itself, since we don't have any raft tests. 😢 

@kaypeter87 do you have a sense of how difficult it might be to get raft set up in a Vault dev server? Is it even possible with a single node? Wondering if we might be able to get minimal testing going or if that's going to be infeasible.

kaypeter87:
@briantist that shouldn't be too difficult. Every feature of Vault is available in "dev" mode. The -dev flag just short-circuits a lot of setup to insecure defaults. All data is stored (encrypted) in-memory.

I haven't personally tried enabling dev mode with raft, but it should be possible from what the documentation states.

adammike (commented):
Is this the only potential use case for these streaming results? Should this change be more broad to cover any potential larger than system memory use cases?

briantist:
> Is this the only potential use case for these streaming results? Should this change be more broad to cover any potential larger than system memory use cases?

@adammike right now, it is the only known use case. The original PR sought to solve it more broadly, but due to the way Adapters are defined, and the leaky abstraction we have, there was not actually a good place to make such a broad change (this cannot work with the JSONAdapter generally). Some of the issues we've discussed in the past about Adapters, and some of the changes in #898 should lead us toward a future where that might make sense, but I believe that for now this is best solved for this one method.

adammike (approved):
LG2M

Issue #683, opened 2021-04-06
generate_secret_id doesnt have wrap_ttl input


Sanz009:
I have been using hvac=0.10.4 for some time now, and has shifted to 0.10.8.
Saw some deprecation warning wrt approle authentication methods, and wanted to use the new methods.

But for create_role_secret_id method alternative, which is generate_secret_id in approle.py, the new method doesnt accept wrap_ttl value. 

Old Method
   ```
 @utils.deprecated_method(
        to_be_removed_in_version='0.12.0',
        new_method=api.auth_methods.AppRole.generate_secret_id,
    )
    def create_role_secret_id(self, role_name, meta=None, cidr_list=None, token_bound_cidrs=None, wrap_ttl=None, mount_point='approle'):
        """POST /auth/<mount_point>/role/<role name>/secret-id

```
New Method

```
    def generate_secret_id(self, role_name, metadata=None, cidr_list=None, token_bound_cidrs=None, mount_point=DEFAULT_MOUNT_POINT):
        """
        Generates and issues a new Secret ID on a role in the auth method.

        Supported methods:
            POST: /auth/{mount_point}/role/{role_name}/secret-id. Produces: 200 application/json
```

Is there a reason why this functionality has been dropped in the new method? 
Vault accepts wrap_ttl to be given over POST.

briantist:
Hi @Sanz009 , thank you for submitting this! I'm sorry this was not addressed before the deprecated method was removed, it should be resolved in #782 .

PR #778, opened 2021-11-08
Add wrap_ttl parameter to approle generate_secret_id

Prividen:
#683


briantist:
@Prividen sorry I just saw this PR, but I actually added this functionality to your other PR #782 🤦‍♂️ since that one has tests and also addresses the custom secret ID method, let's keep it in that one if you don't want. Merging that one will close this one. 
Thank you of course, I hope we'll be able to address these more quickly and have a smaller backlog so these things aren't missed.

PR #782, opened 2021-11-09
approle - fix metadata for generated secret IDs, re-add `wrap_ttl`

### LINKED ISSUE #683: generate_secret_id doesnt have wrap_ttl input

### LINKED ISSUE #781: approle generate_secret_id(): metadata are always provided, even if empty
Function generate_secret_id is always include metadata in Vault request, even it is empty.

Code like this:
```
secret_request = client.auth.approle.generate_secret_id(
    role_name=role_name,
)
```

produce the following request to Vault:
```json
{ "request": {
...
    "data": {
      "metadata": null
    },
...
}
```

Usually it has no any side effect, unless you have restriction policy with the only allowed parameters, not included the ```metadata``` itself:
```hcl
path "auth/approle/role/client-role-*" {
  capabilities = ["create", "update"]
  allowed_parameters = {
    "token_policies" = ["client-policy"]
    "token_ttl" = ["30s"]
    "token_max_ttl" = ["1h"]
  }
}
```

Then you always got exception:
```
hvac.exceptions.Forbidden: 1 error occurred:
	* permission denied

, on post https://my.vault.org/v1/auth/approle/role/client-role-1/secret-id
```

The fix is easy - do not provide ```metadata``` parameter, if it empty.

### LINKED ISSUE #950: approle.create_custom_secret_id() metadata parameter doesn't work
Similar to: 
- #782 

The metadata handling for this method is broken, as it doesn't serialize the value to JSON. In theory it would work if a JSON string was sent through, but in that case it would fail parameter validation anyway.

Prividen:
Fixes #781 
Fixes #950
Fixes #683 
Closes #778

### DIFF
--- a/hvac/api/auth_methods/approle.py
+++ b/hvac/api/auth_methods/approle.py
@@ -267,7 +267,9 @@ class AppRole(VaultApiBase):
                 )
             )
 
-        params = {"metadata": json.dumps(metadata) if metadata else metadata}
+        params = {}
+        if metadata:
+            params = {"metadata": json.dumps(metadata)}
 
         list_of_strings_params = {
             "cidr_list": cidr_list,
--- a/hvac/api/auth_methods/approle.py
+++ b/hvac/api/auth_methods/approle.py
@@ -238,6 +238,7 @@ class AppRole(VaultApiBase):
         cidr_list=None,
         token_bound_cidrs=None,
         mount_point=DEFAULT_MOUNT_POINT,
+        wrap_ttl=None,
     ):
         """
         Generates and issues a new Secret ID on a role in the auth method.
@@ -255,6 +256,9 @@ class AppRole(VaultApiBase):
         :type token_bound_cidrs: list
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
+        :param wrap_ttl: Returns the request as a response-wrapping token.
+            Can be either an integer number of seconds or a string duration of seconds (`15s`), minutes (`20m`), or hours (`25h`).
+        :type wrap_ttl: int | str
         :return: The JSON response of the read_role_id request.
         :rtype: dict
         """
@@ -288,7 +292,7 @@ class AppRole(VaultApiBase):
             mount_point=mount_point,
             role_name=role_name,
         )
-        return self._adapter.post(url=api_path, json=params)
+        return self._adapter.post(url=api_path, json=params, wrap_ttl=wrap_ttl)
 
     def create_custom_secret_id(
         self,
@@ -298,6 +302,7 @@ class AppRole(VaultApiBase):
         cidr_list=None,
         token_bound_cidrs=None,
         mount_point=DEFAULT_MOUNT_POINT,
+        wrap_ttl=None,
     ):
         """
         Generates and issues a new Secret ID on a role in the auth method.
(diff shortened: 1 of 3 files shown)


briantist (approved):
I don't have commit rights or anything but this looks good to me.

@Prividen  use "Fixes #781" in your description so that GitHub properly links them.

briantist:
> I don't have commit rights or anything but this looks good to me.
> 
> @Prividen use "Fixes #781" in your description so that GitHub properly links them.

Update: I'm a maintainer now (funny how that happens) 😅

I've made some updates to the PR:
- rebased
- updated the unit and integration tests to prevent regression
- also addressed `create_custom_secret_id` which suffered from even worse bug (#950)

Thank you for the original issue and the PR @Prividen ! Sorry it took so long to get addressed.

briantist (approved):

briantist (approved):

hvac/api/auth_methods/approle.py
```diff
@@ -286,7 +292,7 @@ def generate_secret_id(
             mount_point=mount_point,
             role_name=role_name,
         )
-        return self._adapter.post(url=api_path, json=params)
+        return self._adapter.post(url=api_path, json=params, wrap_ttl=wrap_ttl)
```
dhuckins, line 295: should `wrap_ttl` be part of the payload? I don't see a `wrap_ttl` parameter on [requests](https://github.com/python/typeshed/blob/b0dc6a340523638df63e1790741689fda693a297/stubs/requests/requests/sessions.pyi#L215-L233)

briantist: it shouldn't be in the payload in this case. `_adapter` is our adapter rather an adapter from `requests`. The `post` method will call the `request` method:
https://github.com/hvac/hvac/blob/develop/hvac/adapters.py#L112-L123

Our `request` method looks for `wrap_ttl` and if it's supplied, it pops it out of the kwargs and makes it into a header:
https://github.com/hvac/hvac/blob/develop/hvac/adapters.py#L288-L290

briantist: I also added both integration and unit test coverage for `wrap_ttl`, if you want to see it in action

dhuckins (approved):

Issue #780, opened 2021-11-09
LIST v1/auth/token/accessors


jessecooper:
Is there a way to get a list of token accessors? 
If not, would the proper style for this package be to create a method called `list_accessors`? I could see how this could also go into the `token_lookup` method if no token is provided but `accessors=True`. I think `list_accessors` method is clearer but please let me know or if there is a different way that is already implemented. I would be willing to open I PR if this is not implemented yet and is desired.

jessecooper:
Found it is implemented already here: https://github.com/hvac/hvac/blob/develop/hvac/api/auth_methods/token.py#L132

PR #784, opened 2021-11-23
#780 add list accessors method

jessecooper:
Add `client.list_accessors()` for `/auth/token/accessors` endpoint.


jessecooper:
Found it is implemented already here: https://github.com/hvac/hvac/blob/develop/hvac/api/auth_methods/token.py#L132

PR #789, opened 2022-01-10
Update wrapping.rst - example for unauthenticated unwrap

nneul:
Provide example of using unwrap with just the wrapping token. Previous example is non-functional unless already authenticated.

### DIFF
--- a/docs/usage/system_backend/wrapping.rst
+++ b/docs/usage/system_backend/wrapping.rst
@@ -8,10 +8,8 @@ Wrapping
 
 Unwrap
 ------
-Is Sealed
----------
 
-.. automethod:: hvac.api.system_backend.Seal.is_sealed
+.. automethod:: hvac.api.system_backend.Wrapping.is_sealed
    :noindex:
 
 Examples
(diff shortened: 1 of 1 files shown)


nneul:
ping or close it please so it's not hanging around

mbieren:
I had this problem mentioned in : https://stackoverflow.com/questions/74531428/hashicorp-vault-approle-based-authentication-unwrap-secret-id-got-permission-den?noredirect=1#comment131590172_74531428
This solves my problem

nneul:
HVAC maintainers - please consider actually merging this.

docs/usage/system_backend/wrapping.rst
```diff
@@ -42,6 +42,17 @@ Examples
...
     print('Unwrapped approle role token secret id accessor: "%s"' % unwrap_response['data']['secret_id_accessor'])
 
+.. testcode:: sys_wrapping
+
+    import hvac
+
+    client = hvac.Client(url='https://127.0.0.1:8200')
+    client.token = "s.xxxxxxxxxxxxxxxxxxxxxx"
+
+    # When authenticating with just the wrapping token, should not pass token into unwrap call
```
colin-pm, line 52: suggestion: I think this could be reworded better

```suggestion
    # Do not pass the token to unwrap when authenticating with the wrapping token
```

colin-pm (dismissed):
One small nit, but overall the doc change looks good to me. Sorry for the delay.

docs/usage/system_backend/wrapping.rst
nneul: Updated.

briantist:
Hi @nneul ! I'd also like to apologize for the delays. I've updated your PR today to rebase it, and fix up the example so that it passes (we run live tests on examples and it needed some updates).

Calling attention to this doc also made me realize it was totally not showing the correct unwrap method documentation at all, and also included something from the Seal backend, so I've cleaned that up as well.

Thanks very much for the contribution!

briantist (approved):

Issue #648, opened 2020-11-30
Add SSH-CA Support to HVAC


evilmog:
Add SSH-CA Support to HVAC as per https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates and https://www.vaultproject.io/api/secret/ssh

jxsl13:
Any updates for this issue?

Roxyrob:
SSH Secret Engine and particularly SSH Certificate support would be very helpful, any news ?

emper0r:
this is very usefull

PR #811, opened 2022-05-11
ssh: Add secret engine

colin-pm:
Added SSH secret engine and tests.

Addresses #648


mjs:
Great! I was literally about to propose a similar PR based on an older abandoned PR (#628) which also implements this functionality.

The differences I see after a quick look:
- Most of the implementation is quite similar (unsurprisingly)
- This PR seems to have better test coverage
- This PR has more comprehensive documentation
- The older PR does a lot of validation in `create_or_update_role` which isn't done herer. I'm not sure how important this is.
- The other PR uses type hints (somewhat inconsistent given that the rest of hvac doesn't seem to include type hints)
- This PR has functionality referring to "zeroaddress" roles. I'm not sure what that's about. Can you explain?
- `sign_ssh_key` (here) is probably a better name than `sign_ca_key`

We're already using a vendored copy of hvac with SSH support patched in and we're very keen to get SSH secrets support merged upstream. @jeffwecan what do you think this PR needs to make it acceptable for merging?

mjs:
@colin-pm It looks like the lint failures are because of black needing to reformat 2 files. That should be an easy fix.

colin-pm:
Working on fixing the CI in a separate PR #812

PR #812, opened 2022-05-23
Fix CI

colin-pm:
The latest version of black appeared to have issues with conf.py and hvac_doctest.py.  Updated files so pipeline errors could be fixed for any rebased pull requests.


colin-pm:
So it looks like the linting CI job installs black without specifying a version, which installs black 22.x, while dev-requirements.txt specifies version 21.6.  Version 22 has a problem with the two files, where 21 does not, which does explain why I had trouble reproducing locally.

I'd also like to submit a PR to fix Documentation Tests next.

jeffwecan (approved):

PR #818, opened 2022-06-14
Don't override user warning filters

intgr:
Currently, whenever a deprecated `hvac` method is called, hvac overrides the caller's warning filter configuration *globally*. This means that warnings from other, entirely unrelated libraries, change behavior.

For example, I want to use `error::DeprecationWarning` to turn DeprecationWarnings into errors in CI builds. But if my code happens to touch a deprecated `hvac` method, the behavior gets overridden to `default::DeprecationWarning` for warnings coming from hvac as well as Django!

Introducing behavior changes to other libraries is sort of a [heisenbug](https://en.wikipedia.org/wiki/Heisenbug). I think it's best to avoid setting warning filters and let the user control it.

### DIFF
--- a/hvac/api/auth_methods/__init__.py
+++ b/hvac/api/auth_methods/__init__.py
@@ -77,12 +77,10 @@ class AuthMethods(VaultApiCategory):
             method_name="login",
             module_name="adapters.Request",
         )
-        warnings.simplefilter("always", DeprecationWarning)
         warnings.warn(
             message=deprecation_message,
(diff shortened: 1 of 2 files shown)


intgr:
@rajmaradia and @colin-pm as aspiring new maintainers may want to chime in here. :)

colin-pm (approved):
Yeah, this makes sense to me.  Setting the warning level should be left to the client of the library.

PR #823, opened 2022-06-22
Readding 3.6 support

colin-pm:
Pulling back in 3.6 support for the rest of the year.  Afterwards HVAC will strive to track Python version support against CPython EOL dates.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -17,6 +17,7 @@ jobs:
     strategy:
       matrix:
         python-version:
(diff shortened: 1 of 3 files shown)


colin-pm:
@briantist Does this look good to you?

briantist (approved):
thanks @colin-pm !

briantist:
Even though the versions will follow CPython EOL dates, it'd be nice to have pinned issues or discussions to specifically announce the versions being dropped, even if there are no deviations in the timelines. wdyt?

PR #827, opened 2022-07-06
Update kubernetes authentication example

eugene-davis:
Update kubernetes authentication example to use the current API call rather than the deprecated one.


colin-pm (approved):
LGTM, thanks for the change!

PR #836, opened 2022-07-18
Add stock version-resolver cfg for release-drafter

jeffwecan:
cc: @colin-pm 

As a method for getting release-drafter aligned with non-patch releases. I.e., this config addition + the `major` label on https://github.com/hvac/hvac/pull/829 will get the draft GitHub release set with the intended tag and name.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -21,6 +21,17 @@ categories:
     labels:
       - 'misc'
 change-template: '- $TITLE. GH-$NUMBER'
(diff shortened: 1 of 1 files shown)


jeffwecan:
Ah, and _also_ this older config needed to be updated as it was set to use `$NEXT_PATCH_VERSION` for the release regardless: https://github.com/hvac/hvac/commit/f23567b475acee38c067f0ece15295a12ef912ea

Issue #802, opened 2022-03-15
Documentation outdated.


ygbourhis:
The documentation here says to use the `auth_kubernetes` method : https://hvac.readthedocs.io/en/stable/usage/auth_methods/kubernetes.html
But it is deprecated and gives a deprecation warning : 
```
DeprecationWarning: Call to deprecated function 'auth_kubernetes'. This method will be removed in version '1.0.0' Please use the 'login' method on the 'hvac.api.auth_methods.kubernetes' class moving forward.
```
The documentation should show the new undeprecated method.

ygbourhis:
I also can't find any documentation on how to use `hvac.api.auth_methods.kubernetes` properly.
Only `auth_kubernetes` works like a charm (but with a deprecation warning).
Can the warning be more helpful (with a code sample) before the deprecated code gets removed ?

Thanks in advance.

theo-ardouin:
It looks like it's working this way:
```python
from hvac import Client, exceptions
from hvac.api.auth_methods import Kubernetes

client = Client(url=url, verify=certificate_path)
Kubernetes(client.adapter).login(role=role, jwt=jwt)
```

Cheers.

ygbourhis:
> It looks like it's working this way:
> 
> ```python
> from hvac import Client
> from hvac.api.auth_methods import Kubernetes
> 
> client = Client(url=url, verify=certificate_path)
> Kubernetes(client.adapter).login(role=role, jwt=jwt)
> ```
> 
> Cheers.

It works.

The documentation should clearly be updated.
Thanks

colin-pm:
Closing this issue now that #837 is merged.

PR #837, opened 2022-07-19
docs(auth-methods): update kubernetes

theo-ardouin:
Add documentation to login use kube with latest version.

Relates to https://github.com/hvac/hvac/issues/802

### DIFF
--- a/docs/usage/auth_methods/kubernetes.rst
+++ b/docs/usage/auth_methods/kubernetes.rst
@@ -6,7 +6,12 @@ Authentication
 
 .. code:: python
 
+    from hvac import Client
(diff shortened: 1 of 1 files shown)


ygbourhis:
I confirm this documentation works.

The deprecation warning does not properly explain how to do it, and the docstring in the code was of strictly no help at all because the doctring is literally a "senior contributor/developer memo" and not a "library end-user help" at all.
This documentation totally solves the problem. I tested and it works perfectly fine.
And the short example even makes the code of the library itself (with it's doctring) far more understandable when you are new on this library because you understand the mechanism and the logic used to code this library in just a glance.

All good for me.

PR #839, opened 2022-07-26
Fix issue #838

### LINKED ISSUE #838: Recursion error in `VaultApiCategory.__getattr__()`
We are seeing a lot of errors like this, in two of our environments.
```
  File "/usr/local/lib/python3.10/site-packages/hvac/api/vault_api_category.py", line 39, in __getattr__
    if item in self.implemented_class_names:
  File "/usr/local/lib/python3.10/site-packages/hvac/api/vault_api_category.py", line 39, in __getattr__
    if item in self.implemented_class_names:
  File "/usr/local/lib/python3.10/site-packages/hvac/api/vault_api_category.py", line 39, in __getattr__
    if item in self.implemented_class_names:
  [Previous line repeated 997 more times]
RecursionError: maximum recursion depth exceeded
```
The runtime context is pods running in a kubernetes context. The outcome does not seem deterministic though as some pods with the same config will fail, while others won't. Could it be related to the fact that some module aren't always loaded in the same order? Or API not listed in the same order?

Just by [reading the code](https://github.com/hvac/hvac/blob/develop/hvac/api/vault_api_category.py#L41), I would say that the only way this happens is that an api category does not have the expected private attribute define (some API impl. not respecting the implementation contract), and hence `getattr()` falls in a recursion pit.

Not sure how/where those api categories are defined yet though.

cansjt:
Fixes #838 

Check attribute exist before calling `getattr()` within `__getattr__()` to prevent infinite recursion loops.

### DIFF
--- a/hvac/api/vault_api_category.py
+++ b/hvac/api/vault_api_category.py
@@ -34,8 +34,8 @@ class VaultApiCategory(VaultApiBase, metaclass=ABCMeta):
         :return: The requested class instance where available.
         :rtype: hvac.api.VaultApiBase
         """
-        if item in self.implemented_class_names:
-            private_attr_name = self.get_private_attr_name(item)
+        private_attr_name = self.get_private_attr_name(item)
+        if item in self.implemented_class_names and hasattr(self, private_attr_name):
             return getattr(self, private_attr_name)
         if item in [u.lower() for u in self.unimplemented_classes]:
             raise NotImplementedError(


hvac/api/vault_api_category.py
```diff
@@ -36,8 +36,8 @@ def __getattr__(self, item):
         :return: The requested class instance where available.
         :rtype: hvac.api.VaultApiBase
         """
-        if item in self.implemented_class_names:
-            private_attr_name = self.get_private_attr_name(item)
+        private_attr_name = self.get_private_attr_name(item)
```
cansjt, line 39: That does not seem to do the trick. With further testing it seem the call to `self.get_private_attr_name()` also causes an infinite recursion.

briantist:
Hi @cansjt ,

That being said, there's been additional research on the issue itself since this PR, which may be fixed in https://github.com/hvac/hvac/pull/982 . If you have a chance and could review that it would be really helpful!

PR #842, opened 2022-07-28
Re-add arguments to create_or_update_role() from old API

intgr:
The previous, deprecated `create_token_role()` method has arguments `period` and `explicit_max_ttl` that were dropped in the new one for some reason (in #734).

Re-added these two arguments to the new `create_or_update_role()` method (using new spelling `token_period`, `token_explicit_max_ttl`).

There are plenty more that are supported by Vault but missing here: https://www.vaultproject.io/api-docs/auth/token#create-update-token-role

### DIFF
--- a/hvac/api/auth_methods/token.py
+++ b/hvac/api/auth_methods/token.py
@@ -555,6 +555,8 @@ class Token(VaultApiBase):
         renewable=True,
         path_suffix=None,
         allowed_entity_aliases=None,
+        period=None,
+        explicit_max_ttl=None,
         mount_point=DEFAULT_MOUNT_POINT,
     ):
         """Create (or replace) the named role.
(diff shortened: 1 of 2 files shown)


briantist:
Hi @intgr ,

On another note, I'm sorry for how long it has taken to get review on this PR and I hope to attack the backlog more as I work toward make some structural changes in the project.

hvac/api/auth_methods/token.py
```diff
@@ -555,6 +555,8 @@ def create_or_update_role(
         renewable=True,
         path_suffix=None,
         allowed_entity_aliases=None,
+        token_period=None,
+        token_explicit_max_ttl=None,
```
briantist, line 559: Although it looks a little weird, let's move these new parameters down to the end, just to accommodate folks who might be setting parameters positionally.

hvac/api/auth_methods/token.py
```diff
@@ -586,6 +588,10 @@ def create_or_update_role(
         :type path_suffix: str
         :param allowed_entity_aliases: not case sensitive.
         :type allowed_entity_aliases: str
+        :param token_period: the token will have no maximum TTL, every renewal will use the given period.
+        :type token_period: str
+        :param token_explicit_max_ttl: the token cannot be renewed past this TTL value.
+        :type token_explicit_max_ttl: str
```
briantist, line 594: and then these entries would move too

briantist (commented):
Hi @intgr , thanks for submitting this! Would you be able to take a look at the tests for token auth as well, and add some coverage for these parameters?

briantist (approved):
@intgr thanks again for this PR, I've gone ahead an switched the parameter order and added some unit tests for token auth after rebasing your branch. Cheers!

PR #844, opened 2022-07-28
build-test: Add test

colin-pm:
Adds build test to ensure hvac is always in a buildable state and can be installed on Mac, Ubuntu, and Windows.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,0 +1,38 @@
+name: Test build and installation
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


colin-pm:
Made a small mistake on the initial change with Python versions requiring strings so Python version 3.10 is interpreted as "3.10" and not 3.1...

Made this change. As expected and pointed out in #843, the tests fail for Windows.  I will still merge this as the failed tests reflect the current state of the repository. I would like to merge this change beforehand so the new tests can run for #843.

PR #846, opened 2022-08-03
Allow configuring the possible salt lengths for RSA PSS signatures

trishankatdatadog:
The Golang crypto/rsa package [allows](https://pkg.go.dev/crypto/rsa#pkg-constants) for using at least two possible [salt lengths](https://crypto.stackexchange.com/questions/1217/rsa-pss-salt-size) (auto and hash for want better terminology) when creating/verifying RSA PSS signatures. However, these options are currently not exposed via the Vault Transit Secrets Engine and, hence, HVAC. These options would allow for cross-platform signature verification without modifying downstream clients (e.g., the https://github.com/secure-systems-lab/securesystemslib/pull/262 package, used by the [python-tuf](https://github.com/theupdateframework/python-tuf) and [python-in-toto](https://github.com/in-toto/in-toto) supply chain security frameworks, which currently assumes a different salt length by default than Vault does).

This PR is related to Vault PR https://github.com/hashicorp/vault/pull/16549. Assuming that that PR is merged, this PR would update HVAC correspondingly.

### DIFF
--- a/hvac/api/secrets_engines/transit.py
+++ b/hvac/api/secrets_engines/transit.py
@@ -753,6 +753,7 @@ class Transit(VaultApiBase):
         prehashed=None,
         signature_algorithm=None,
         marshaling_algorithm=None,
+        salt_length=None,
         mount_point=DEFAULT_MOUNT_POINT,
     ):
         """Return the cryptographic signature of the given data using the named key and the specified hash algorithm.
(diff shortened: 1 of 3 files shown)


ofek (dismissed):

colin-pm:
Hi @trishankatdatadog thank you for this contribution adding support for a new feature you've added to Vault!

I noticed you have some commits where you modified the CI jobs and then effectively reverted the changes in successive commits.

If you don't mind, to keep our git history a little cleaner, do you mind rebasing and cleaning the git commits on your feature branch? I would be happy to review the changes once this is done.

trishankatdatadog:
> If you don't mind, to keep our git history a little cleaner, do you mind rebasing and cleaning the git commits on your feature branch? I would be happy to review the changes once this is done.

In the interest of time, is it ok to just squash it all into one commit and force-push?

colin-pm:
> In the interest of time, is it ok to just squash it all into one commit and force-push?

Yes, I think that would be fine.

trishankatdatadog:
> Yes, I think that would be fine.

Done, PTAL, thx!

trishankatdatadog:
_bump_ 🙂

yzhan289 (approved):

trishankatdatadog:
> @trishankatdatadog I've reviewed the changes and the code looks good to me, I just have a few questions.
> 
> Looking at the Vault repo, this PR adds support for a Vault feature that is unreleased, correct? I'm not 100% sure how Vault reacts to unsupported parameters. Will this change be backwards compatible with older versions of Vault?

I don't know, TBH. Don't the unit tests test for this right now? They seem to be testing against rather old versions of Vault Enterprise.

colin-pm (approved):
As long as the integration tests pass, this LGTM

PR #847, opened 2022-08-06
Handle HTTP 204 empty responses in JSON adapter

### LINKED ISSUE #293: "No JSON object could be decoded" when writing to KV v2
# Versions

- hvac==0.6.4
- Vault v0.11.1 ('8575f8fedcf8f5a6eb2b4701cb527b99574b5286')

# Problem

I'm currently running into a weird issue where I am attempting to write secrets to the kv v2 api, and it is returning an error.

This is an example of what I'm doing:

``` python
import os
import hvac

vault_url = os.environ.get('VAULT_ADDR', 'http://127.0.0.1:8200')
vault_token = os.environ.get('VAULT_TOKEN')
vault_secret_prefix = 'test/demoapp/'

client = hvac.Client(url=vault_url, token=vault_token)

kv = client.kv.v2
kv.create_or_update_secret(
    path="{}secret2".format(vault_secret_prefix),
    secret=dict(foo='bar'))

```

And this is the error:

```
  File "hvac-test.py", line 31, in <module>
    secret=dict(foo='bar'))
  File "/usr/local/lib/python2.7/site-packages/hvac/api/secrets_engines/kv_v2.py", line 124, in create_or_update_secret
    return response.json()
  File "/usr/local/lib/python2.7/site-packages/requests/models.py", line 896, in json
  File "/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 339, in loads
  File "/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 364, in decode
  File "/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 382, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

https://github.com/hvac/hvac/blob/master/hvac/api/secrets_engines/kv_v2.py#L124 looks like it always returns `response.json()` but after adding some debug statements to that function, the response from vault is a `204` and there's no data to parse.

**Sidenote**:  The above error is from running vault in server mode.   I tested using dev mode, and I do not get the same error.  I haven't investigated why, but vault in dev mode apparently returns a response (with a 200 instead of 204) when it does not without dev mode.  I might open a ticket on the vault repo for this if the docs don't mention it.

### LINKED ISSUE #385: kv2 create_or_update_secret response.json() failse
https://github.com/hvac/hvac/blob/develop/hvac/api/secrets_engines/kv_v2.py#L124

Blows up because the server returns and empty body and 204s me.  It would be nice if a response wasn't required on update.

### LINKED ISSUE #660: lookup_group returning Response objects
In a possible regression of #331 it seems lookup_group, and probably other functions as well, are returning Response objects when they get a 204 instead of None as the documentation claims it should. I have only verified this with lookup_group, as that happened to be the call where I encountered the problem.

### LINKED ISSUE #797: The JSON adapter should return an empty response on 204 rather than the response object
https://github.com/hvac/hvac/blob/develop/hvac/adapters.py#L365-L371

On a 204 response, the entire response object is returned, but I think it should be an empty dict instead (or maybe `None` would be fine as well), to be more consistent, since Vault is known to return 204 responses when it has nothing to say, and that result is more appropriate to deal with as a client.

As an example, if I want to test for this case, I need to `import requests` and test the the type of the response, or I can try checking for a `status_code` field, but if the vault request returned some data with a `status_code` field, then I can't rely on that.

There's a number of workarounds I can think of to handle this, but it'd be better for this common response type to be handled by the adapter in a way that's friendlier to deal with on the client side.

briantist:
Closes #797 
Closes #293 
Closes #385 
Closes #660

Most detail is in #797 . Discussed this a bit on the discord, and we came to the conclusion of returning an empty dict on 204, but looking at #660 and the [documentation it references](https://hvac.readthedocs.io/en/v0.10.10/source/hvac_api_secrets_engines.html#hvac.api.secrets_engines.Identity.lookup_group), there's at least one method that is explicitly saying it will return `None` on no response, so I'm torn again.

It's possible to update the methods to either change their documentation, or to make them interpret the empty dict and return `None` if that's what best for them, or we can default the adapter to return `None`.

Putting up what I have for now with the empty dict return and we can figure out what would be best.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -366,6 +366,8 @@ class JSONAdapter(RawAdapter):
                 return response.json()
             except ValueError:
                 pass
+        elif response.status_code == 204:
+            return {}
 
         return response
 
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ docs/_build/
 *.iml
 venv/
 .envrc
+.vscode/
--- a/tests/unit_tests/test_adapters.py
+++ b/tests/unit_tests/test_adapters.py
@@ -3,6 +3,7 @@ import logging
 from unittest import TestCase
 
 import requests_mock
+from requests import Response
 from parameterized import parameterized, param
 from hvac.constants.client import DEFAULT_URL
 from hvac import adapters
@@ -105,3 +106,100 @@ class TestRequest(TestCase):
             second=response.status_code,
         )
         self.assertEqual(first=mock_response, second=response.json())
+
+    @parameterized.expand(
+        [
+            param(
+                "standard Vault address",
+                url="https://localhost:8200",
+            ),
+            param(
+                "Vault address with route",
+                url="https://example.com/vault",
+            ),
+            param(
+                "regression test for hvac issue #51",
+                url="https://localhost:8200",
+                path="keyring/http://some.url/sub/entry",
+            ),
+            param(
+                "redirect with location header for issue #343",
+                url="https://localhost:8200",
+                path="secret/some-secret",
+                redirect_url="https://some-other-place.com/secret/some-secret",
+            ),
+            param(
+                "204 empty response",
+                url="https://localhost:8200",
+                response_status_code=204,
+                mock_json=None,
+            ),
+            param(
+                "non-json response",
+                url="https://localhost:8200",
+                mock_text="<html></html>",
+                parseable=False,
+            )
+        ]
+    )
+    def test_json_get(
+        self, label, url, path="v1/sys/health", redirect_url=None, parseable=True, mock_text=None,
+        response_status_code=200, mock_json={"auth": None, "warnings": None, "data": {"key1": "val1"}}
+    ):
+        path = path.replace("//", "/")
+        expected_status_code = response_status_code
+        mock_url = f"{url}/{path}"
+        expected_request_urls = [mock_url]
+        adapter = adapters.JSONAdapter(base_uri=url)
+        response_headers = {}
+
+        mock_content = {}
+        if mock_text is not None:
+            mock_content['text'] = mock_text
+        elif mock_json is not None:
+            mock_content['json'] = mock_json
+
+        if redirect_url is not None:
+            response_headers["Location"] = redirect_url
+            response_status_code = 301
+        with requests_mock.mock() as requests_mocker:
+            logging.debug('Registering "mock_url": %s' % mock_url)
+            requests_mocker.register_uri(
+                method="GET",
+                url=mock_url,
+                headers=response_headers,
+                status_code=response_status_code,
+                **mock_content,
+            )
+            if redirect_url is not None:
+                expected_request_urls.append(redirect_url)
+                logging.debug('Registering "redirect_url": %s' % redirect_url)
+                requests_mocker.register_uri(
+                    method="GET",
+                    url=redirect_url,
+                    **mock_content,
+                )
+
+            response = adapter.get(
+                url=path,
+            )
+
+        # Assert all our expected uri(s) were requested
+        for request_num, expected_request_url in enumerate(expected_request_urls):
+            self.assertEqual(
+                first=expected_request_url,
+                second=requests_mocker.request_history[request_num].url,
+            )
+
+
+        if parseable:
+            expected_response = mock_json
+            if expected_status_code == 204:
+                expected_response = {}
+
+            self.assertEqual(
+                first=response,
+                second=expected_response,
+            )
+        else:
+            self.assertIsInstance(response, Response)


hvac/adapters.py
```diff
@@ -366,6 +366,8 @@ def request(self, *args, **kwargs):
                 return response.json()
             except ValueError:
                 pass
+        elif response.status_code == 204:
+            return {}
 
         return response
```
adammike, line 372: We should probably never return the response object. Do we also return an empty dictionary here? Raise an exception?

briantist: Hm, I tend to agree that returning the response is not the best idea, though it has been in the adapter since it was introduced: https://github.com/hvac/hvac/pull/537/files#diff-db919c05d0fa65b24b7b140bab3fd3530b3530735e4e9e20a843881557158936

There is a `RawAdapter` that can be used when the full response object is desired, and that's the superclass for this already. 

I suppose the way we can look at this is that this is, specifically, the `JSONAdapter`, so a response that's not valid or parseable as JSON could (should?) be an exception. 

I don't think it should be an empty dictionary in this case, because it implies that everything went ok but really there was a response there that's now hidden.

The main thing we need to be careful of is that this is the default adapter, but then again, if anyone was using this full response object I'd guess it was mostly to deal with 204s, since 4xx and 5xx codes would already be raising exceptions.

colin-pm:
The only thing I don't like about returning an empty dict is that it casts to bool as False. So a condition that depends on the response of an API call that returns 204 will evaluate as False.

I'm not super sure if there are Vault API endpoints that sometimes return a 204 and sometimes returns 200, because that could make downstream code buggy. If the Vault API is consistent where endpoints exclusively return 200 or 204, I think I'm a little less concerned.

The other idea I thought of is to create a hvac.response object that we could return that's similar to a requests.response or a dict. It would just have the return code and dict attributes.  It could also be a dict with a return_code attribute. That way, if the return code is 204, the object would have an empty dict and response code 204, but could still cast True.

One small nitpick I have is the docstrings should be updated for this change.

adammike:
> The other idea I thought of is to create a hvac.response object that we could return that's similar to a requests.response or a dict. It would just have the return code and dict attributes. It could also be a dict with a return_code attribute. That way, if the return code is 204, the object would have an empty dict and response code 204, but could still cast True.

I was thinking this also, but I was concerned it would be a *major* breaking change. Every app that uses this library would need to adapt to the new output.  It's a pain, but honestly I think this is the "right" fix.

colin-pm:
> I was thinking this also, but I was concerned it would be a _major_ breaking change. Every app that uses this library would need to adapt to the new output. It's a pain, but honestly I think this is the "right" fix.

One idea is we could implement `__getitem__` and `get` for hvac.response, so the API change wouldn't be drastic and any code handling a 200 response would still work as-is. It probably would be better long-term to access the dict with `hvac.response.value`, so we could add a warning when using `hvac.response.get()` or `hvac.response.__getitem__()` that it'll be deprecated in the next major release.  This could provide and easier transition to hvac.response.

adammike:
> > I was thinking this also, but I was concerned it would be a _major_ breaking change. Every app that uses this library would need to adapt to the new output. It's a pain, but honestly I think this is the "right" fix.
> 
> One idea is we could implement `__getitem__` and `get` for hvac.response, so the API change wouldn't be drastic and any code handling a 200 response would still work as-is.

That's really elegant. I'd never have thought of that. Great idea.

briantist:
> The only thing I don't like about returning an empty dict is that it casts to bool as False. So a condition that depends on the response of an API call that returns 204 will evaluate as False.

Although it does evaluate to `False`, and I can't be 100% certain that no one is relying on the existing behavior, I think it's unlikely. I raised some of that in the discord discussion, but the main reason I think it's unlikely is that there isn't really a situation now where such a conditional would ever return `False`, so there's little point for the conditional to exist.

Current behavior is that 200 and 204 would evaluate to `True`, but all of the other responses that could evaluate to `False` will raise exceptions instead.

I still consider this a breaking change, but it is hard to know what the level of breakage is.

> I'm not super sure if there are Vault API endpoints that sometimes return a 204 and sometimes returns 200, because that could make downstream code buggy. If the Vault API is consistent where endpoints exclusively return 200 or 204, I think I'm a little less concerned.

Unfortunately it is not consistent. The reason is that although methods are well-documented as returning a 204 response, if there are any warnings to report, the response will be 200 so that it can include the `warnings` field. This is documented here: https://www.vaultproject.io/api-docs#http-status-codes

> The other idea I thought of is to create a hvac.response object that we could return that's similar to a requests.response or a dict. It would just have the return code and dict attributes. It could also be a dict with a return_code attribute. That way, if the return code is 204, the object would have an empty dict and response code 204, but could still cast True.

We could definitely do this; but as @adammike mentioned it is a massive change (not in terms of implementation, but for clients). IMO the way we should do something like this is to make a new Adapter, call it the `HVACAdapter` or whatever, independent of the `JSONAdapter`, but don't make it the default. We can plan to make it the default in the next major release (`2.0.0`), and announce that change in `1.0.0`. This allows folks to opt in to it early, get used to it, change their project, etc., *_or_* to make their code explicitly use `JSONAdapter` (which won't be going away) to preserve existing behavior indefinitely in their projects.

To gain adoption, we can _recommend_ use of the new one. What would be more effective is updating examples to show usages with both the `JSONAdapter` and the new one, and to also start showing client instantiation with explicit adapter choice despite defaults.

While this is a longer road, it won't break anyone suddenly, and it gives folks a long runway to prepare for the change of default; even after that, they can still have the old behavior with a very small change (explicitly ask for `JSONAdapter`), that can be made any time (even right now). It also has the advantage that we don't need to implement shims or convenience methods like overriding `__getitem__` that could be difficult to ever remove.

In fact, I have been mulling over having better built-in retry support, by using urllib3's `Retry` class. This is already possible by providing a custom `Session` with any adapter, and I do this in my project. But I think we can provide solid defaults for something like that, either in all Adapters (still opt-in), or as a new Adapter. Anyway, topic for another time...

> One small nitpick I have is the docstrings should be updated for this change.

Absolutely, not a nit, I just wanted to get us settled on a course of action first. 

I also wanted to ask how or whether we have, a way to document changed and their level/impact, at the PR level. If we only update the changelog at release time, it seems likely we will be missing a lot of stuff.

Ansible projects for example, have a system of writing "changelog fragments" and these are combined with a tool at release time to generate the changelog. We could use that system directly ([it doesn't have to be specific to Ansible](https://github.com/ansible-community/antsibull-changelog/blob/main/docs/other-projects.rst)) but I don't know a lot about what else it out there and what people use.

colin-pm:
After some internal discussion with the @hvac/hvac-maintainers, we decided that we will go the route of creating a new adapter that returns an hvac.response instead of a requests.response. This hvac.response will handle 204 responses better and add a more consistent response object for clients. This will be opt-in initially while the current json response adapter is deprecated and eventually removed as default in a later major release.
Because there will be a smooth transition that will not create any breaking changes in the near future, this issue will be descoped from the 1.0.0 release.

briantist:
As @colin-pm said, we have a more ambitious but better long-term plan for this, so I'll close this out now, and we'll be back with something new!

PR #854, opened 2022-09-02
Add Poetry

colin-pm:
These commits add Poetry to the project as a new means of dependency management, handling the virtual environment for developers, and aiding in other small tasks like bumping versions, building HVAC, or uploading to PyPi.

Also collapsed documentation requirements into development requirements.


colin-pm:
It looks like something is wrong with the Install Poetry action and Python 3.6.  I'm not sure if the action still supports Python 3.6, but this may have to wait until we drop 3.6 support at the end of the year...

.github/workflows/build-test.yml
```diff
@@ -26,38 +29,25 @@ jobs:
...
+        key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
+
     - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install setuptools wheel twine
+      run: poetry install --no-interaction --no-root
     
-    - name: Build and install wheel
+    - name: Build and install wheel for *nix
```
briantist, line 50: nit: this is building cross-platform isn't it? I'd also be fine with removing the for/on portion from the name completely
```suggestion
    - name: Build and install wheel on ${{ matrix.os }}
```

colin-pm:
Turns out there was an issue with the newest version of Poetry, which does not support Python 3.6.

After this gets merged, I will create an issue to update the supported Poetry version in CI and regenerate the lock file with Poetry 2.x.

Also, there's a couple dependencies I had to pin to get everything working with 3.6 since a lot of dependencies have dropped 3.6 in their latest versions. I will also create another issue to clean up dependencies when we drop 3.6 support.

It would probably be best to create a small project to track the work that can be done to drop 3.6 support at the end of December.

All tests appear to be passing, so I'd enjoy getting some feedback on this proposed change!

pyproject.toml
```diff
@@ -0,0 +1,56 @@
+[tool.poetry]
+name = "hvac"
+version = "1.0.0"
+description = "HashiCorp Vault API client"
+readme = "README.md"
+authors = ["Ian Unruh <ianunruh@gmail.com>, Jeffrey Hogan <jeff.hogan1@gmail.com>"]
+maintainers = ["Brian Scholer <brian@briantist.com>", "Colin McAllister <colinmca242@gmail.com>", "Mike Adams <Mike.Adams@kyndryl.com>"]
```
briantist, line 7: Would it be ok to remove my email address and use my GitHub ID `@briantist`? Or maybe the committer email I use if an email address is necessary? `1260690+briantist@users.noreply.github.com`

.github/workflows/build-test.yml
colin-pm: Whoops, I left this in when I was still using different workflows for *nix and Windows. To use the poetry installer I have to use Bash for both, which allowed me to merge the workflows. I just removed "for *nix".

pyproject.toml
colin-pm: That's fine with me, I added the committer email.

briantist (dismissed):

colin-pm:
Removed emails for @briantist and @adammike as requested. Pending code approval, this change is now ready to merge.

briantist (approved):

yan12125:
I noticed pyproject.toml specifies `requests = "2.27.1"` instead of `requests = "^2.27.1"` like other dependencies. I think newer requests are also compatible?

PR #869, opened 2022-09-12
Token: Add create orphaned token method

### LINKED ISSUE #758: Does auth.token.create support orphan tokens?
We are receiving this deprecation warning when calling `create_token`:

```
DeprecationWarning: Call to deprecated function 'create_token'. This method will be removed in version '1.0.0' Please use the 'create' method on the 'hvac.api.auth_methods.token' class moving forward.
```

However, we are creating these tokens as orphans by setting `orphan=True` when calling `create_token`.  Also, the client that creates the tokens is authenticated with sudo privileges, not a root token.

In looking at the new method `auth.token.create` there is no `orphan` argument, but there is a `no_parent` one which is passed as a parameter to the API endpoint.

However, to create an orphan token via sudo privileges the endpoint used must be `/auth/token/create-orphan`:

> If used via the /auth/token/create-orphan endpoint, a root token is not required to create an orphan token (otherwise set with the no_parent option).

In the deprecated `Client.create_token` method if the `orphan=True` argument is passed it uses the `create-orphan` endpoint:

```
if orphan:
    return self._adapter.post(
        "/v1/auth/token/create-orphan", json=params, wrap_ttl=wrap_ttl
    )
```

In the new `auth.token.create` method there appears to be no way to call the `create-orphan` endpoint.  It is always the `create` one.  It appears this could be similarly handled by adapting the endpoint based on the `no_parent` argument, something like:

```
if no_parent:
    create_endpoint = 'create-orphan'
else:
    create_endpoint = 'create'

if role_name is not None:
    api_path = "/v1/auth/{mount_point}/{create_endpoint}/{role_name}".format(
        mount_point=mount_point,
        create_endpoint=create_endpoint,
        role_name=role_name,
    )
    return self._adapter.post(
        url=api_path,
        json=params,
    )

api_path = "/v1/auth/{mount_point}/{create_endpoint}".format(
    mount_point=mount_point,
    create_endpoint=create_endpoint
)
return self._adapter.post(
    url=api_path,
    json=params,
    wrap_ttl=wrap_ttl,
)
```

Maybe I'm missing something and there is already a way to access the `create-orphan` endpoint using non-deprecated paths, but if not is this a good way to handle it?  If so we could test and submit a pull request.

Thanks.

colin-pm:
The Vault API allows orphaned tokens to be created without root privileges using the create-orphan endpoint, a separate method should exist to use that endpoint. This allows implementers the choice of using the create endpoint with the no_parent parameter set to true or with the new create_orphan method.

This closes #758

### DIFF
--- a/hvac/api/auth_methods/token.py
+++ b/hvac/api/auth_methods/token.py
@@ -128,6 +128,102 @@ class Token(VaultApiBase):
             wrap_ttl=wrap_ttl,
         )
 
+    def create_orphan(
+        self,
+        id=None,
+        policies=None,
+        meta=None,
+        no_default_policy=False,
+        renewable=True,
+        ttl=None,
+        type=None,
+        explicit_max_ttl=None,
+        display_name="token",
+        num_uses=0,
+        period=None,
+        entity_alias=None,
+        wrap_ttl=None,
+        mount_point=DEFAULT_MOUNT_POINT,
+    ):
+        """Create a new orphaned token.
+
+        Creates a token via the /auth/token/create-orphan endpoint. A root token
+        is not required to create an orphan token with this endpoint (otherwise
+        an orphaned token can be set with the `create` no_parent option).
+
+
+        :param id: The ID of the client token. Can only be specified by a root token.
+            The ID provided may not contain a `.` character. Otherwise, the
+            token ID is a randomly generated value.
+        :type id: str
+        :param policies: A list of policies for the token. This must be a
+            subset of the policies belonging to the token making the request, unless root.
+            If not specified, defaults to all the policies of the calling token.
+        :type policies: list
+        :param meta: A map of string to string valued metadata. This is
+            passed through to the audit devices.
+        :type meta: map
+        :param no_default_policy: If `True` the default policy will not be contained in this token's policy set.
+        :type no_default_policy: bool
+        :param renewable:  Set to false to disable the ability of the token to be renewed past its initial TTL.
+            Setting the value to true will allow the token to be renewable up to the system/mount maximum TTL.
+        :type renewable: bool
+        :param ttl: The TTL period of the token, provided as "1h", where hour is the largest suffix. If not provided,
+            the token is valid for the default lease TTL, or indefinitely if the root policy is used.
+        :type ttl: str
+        :param type: The token type. Can be "batch" or "service". Defaults to the type
+            specified by the role configuration named by role_name.
+        :type type: str
+        :param explicit_max_ttl: If set, the token will have an explicit max TTL set upon it.
+            This maximum token TTL cannot be changed later, and unlike with normal tokens, updates to the system/mount
+            max TTL value will have no effect at renewal time -- the token will never be able to be renewed or used past
+            the value set at issue time.
+        :type explicit_max_ttl: str
+        :param display_name: The display name of the token.
+        :type display_name: str
+        :param num_uses: The maximum uses for the given token. This can be
+            used to create a one-time-token or limited use token. The value of 0 has no
+            limit to the number of uses.
+        :type num_uses: int
+        :param period: If specified, the token will be periodic; it will have
+            no maximum TTL (unless an "explicit-max-ttl" is also set) but every renewal
+            will use the given period. Requires a root token or one with the sudo capability.
+        :type period: str
+        :param entity_alias: Name of the entity alias to associate with during token creation.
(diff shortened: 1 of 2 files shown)


hvac/api/auth_methods/token.py
```diff
@@ -128,6 +128,102 @@ def create(
...
+        period=None,
+        entity_alias=None,
+        wrap_ttl=None,
+        mount_point=DEFAULT_MOUNT_POINT,
+    ):
+        """Create a new orphaned token.
+
+        Creates a token via the /auth/token/create-orphan endpoint. A root token
+        is not required to create an orphan token with this endpoint (otherwise
+        an orphaned token can be set with the `create` no_parent option).
```
briantist, line 152: nit
```suggestion
        an orphaned token can be set with the `create` method's `no_parent` option).
```

hvac/api/auth_methods/token.py
```diff
@@ -128,6 +128,102 @@ def create(
...
+        :type policies: list
+        :param meta: A map of string to string valued metadata. This is
+            passed through to the audit devices.
+        :type meta: map
+        :param no_default_policy: If `True` the default policy will not be contained in this token's policy set.
+        :type no_default_policy: bool
+        :param renewable:  Set to false to disable the ability of the token to be renewed past its initial TTL.
+            Setting the value to true will allow the token to be renewable up to the system/mount maximum TTL.
+        :type renewable: bool
+        :param ttl: The TTL period of the token, provided as "1h", where hour is the largest suffix. If not provided,
```
briantist, line 171: nit
```suggestion
        :param ttl: The TTL period of the token, provided as `1h`, where hour is the largest suffix. If not provided,
```

hvac/api/auth_methods/token.py
```diff
@@ -128,6 +128,102 @@ def create(
...
+            the token is valid for the default lease TTL, or indefinitely if the root policy is used.
+        :type ttl: str
+        :param type: The token type. Can be "batch" or "service". Defaults to the type
```
briantist, line 174: nit
```suggestion
        :param type: The token type. Can be `batch` or `service`. Defaults to the type
```

hvac/api/auth_methods/token.py
```diff
@@ -128,6 +128,102 @@ def create(
...
+        :type type: str
+        :param explicit_max_ttl: If set, the token will have an explicit max TTL set upon it.
+            This maximum token TTL cannot be changed later, and unlike with normal tokens, updates to the system/mount
+            max TTL value will have no effect at renewal time -- the token will never be able to be renewed or used past
+            the value set at issue time.
+        :type explicit_max_ttl: str
+        :param display_name: The display name of the token.
+        :type display_name: str
+        :param num_uses: The maximum uses for the given token. This can be
+            used to create a one-time-token or limited use token. The value of 0 has no
```
briantist, line 185: nit
```suggestion
            used to create a one-time-token or limited use token. The value of `0` has no
```

hvac/api/auth_methods/token.py
```diff
@@ -128,6 +128,102 @@ def create(
...
+            limit to the number of uses.
+        :type num_uses: int
+        :param period: If specified, the token will be periodic; it will have
+            no maximum TTL (unless an "explicit-max-ttl" is also set) but every renewal
```
briantist, line 189: nit
```suggestion
            no maximum TTL (unless an `explicit-max-ttl` is also set) but every renewal
```

hvac/api/auth_methods/token.py
```diff
@@ -128,6 +128,102 @@ def create(
...
+            will use the given period. Requires a root token or one with the sudo capability.
+        :type period: str
+        :param entity_alias: Name of the entity alias to associate with during token creation.
+            Only works in combination with role_name argument and used entity alias must be listed in
+            `allowed_entity_aliases`. If this has been specified, the entity will not be inherited from the parent.
+        :type entity_alias: str
+        :param wrap_ttl: Specifies response wrapping token creation with duration. IE: '15s', '20m', '25h'.
```
briantist, line 196: nit
```suggestion
        :param wrap_ttl: Specifies response wrapping token creation with duration. IE: `15s`, `20m`, `25h`.
```

briantist (approved):
All my suggestions are non-blocking documentation formatting preferences, thanks for this!

briantist (approved):

PR #886, opened 2022-09-14
Cert: Fix role certificate parameter

### LINKED ISSUE #841: create_ca_certificate_role() fails with 'File name too long' on macOS
The `Client.auth.cert.create_ca_certificate_role()` method fails on macOS when I pass a certificate PEM string as the `certificate=` argument.

The method tries to be helpful and open the passed-in certificate argument as a file path:
```python
        try:
            with open(certificate, 'r') as f_cert:
                cert = f_cert.read()
        except FileNotFoundError:
            cert = certificate
```

But instead of `FileNotFoundError`, an `OSError` gets thrown:

> OSError: [Errno 63] File name too long: '-----BEGIN CERTIFICATE-----\nMII[...snip...]\n-----END CERTIFICATE-----\n'

I'm not sure what the proper fix for this would be. I assume it's not a good idea to break backwards compatibility with the path option of the `certificate` argument?

Maybe a new argument `certificate_str` that always gets used as-is? Or should we do some crazy detection logic like check if `\n` exists in the string, and avoid trying to open it?

colin-pm:
The `create_ca_certificate_role` currently accepts a `certificate` parameter that can either be a PEM-formatted certificate or a path to a PEM-fromatted certificate file. The parameter is differentiated by trying to open a file with what's passed to `certificate`. For some systems, the lengh of a certificate is long enough that it will cause a OSError because of the path length.

The conditional was updated to actually inspect what is passed to contents. To avoid these issues moving forward, `certificate` will only accept a certificate string. Paths should be passed to the new `certificate_file` parameter. For now, passing a certificate file to `certificate` will generate a warning.

Closes #841


briantist (commented):
Just a few questions:
* Have we decided in which version `certificate` will stop accepting files altogether?
* Can we report that version in the warning?
* Are we tracking that in GitHub anywhere? (we could create an issue for removing it, assign it to a `v2.0.0` milestone for example)
* I don't see tests updated, are we missing tests for this method currently?

colin-pm:
> Just a few questions:
> 
> * Have we decided in which version `certificate` will stop accepting files altogether?

Let's do v2.0.0

> * Can we report that version in the warning?

I did add documentation in the API, which will propagate into the docs. Should we also add the version support will be removed in the warning message?

> * Are we tracking that in GitHub anywhere? (we could create an issue for removing it, assign it to a `v2.0.0` milestone for example)

Created #914!

> * I don't see tests updated, are we missing tests for this method currently?

Just added tests 😄, forgot about doing that when I initially created this PR.

briantist:
> I did add documentation in the API, which will propagate into the docs. Should we also add the version support will be removed in the warning message?

imo, yes, it's helpful to see in output, we could also add a link to #914 maybe?

> > * Are we tracking that in GitHub anywhere? (we could create an issue for removing it, assign it to a `v2.0.0` milestone for example)
> 
> Created #914!

Thanks! I updated the title and pinned the issue to make it more visible (maybe we can come up with a different way of surfacing upcoming breaking changes but should be ok for now).

hvac/api/auth_methods/cert.py
```diff
@@ -44,8 +47,13 @@ def create_ca_certificate_role(
         »Parameters
         :param name: The name of the certificate role.
         :type name: str
-        :param certificate: The PEM-format CA certificate.
+        :param certificate: The PEM-format CA certificate. Either certificate or certificate_file is required.
+            NOTE: Passing a certificate file path with the certificate argument is deprecated and will be dropped in
+            version 2.0.0
```
briantist, line 52:
```suggestion
            version 3.0.0
```

briantist (approved):

PR #891, opened 2022-09-16
Add `role_name` parameter to `auth.token.create_orphan`

briantist:
#869 added the `create_orphan` method, but we were missing `role_name`.

This also fixes some headings in the changelog.


hvac/api/auth_methods/token.py
```diff
@@ -156,6 +157,8 @@ def create_orphan(
             The ID provided may not contain a `.` character. Otherwise, the
             token ID is a randomly generated value.
         :type id: str
+        :param role_name: he name of the token role.
```
colin-pm, line 160: nitpick: typo

```suggestion
        :param role_name: The name of the token role.
```

colin-pm (changes requested):
One small doc issue in token.py, but otherwise lgtm!

Thanks for fixing the docs!

colin-pm (changes requested):
If you're planning on immediately merging this into master to create the 1.0.1 patch release, you'll need to bump the version in pyproject.toml. The easiest way to do it is with this command.

```bash
poetry version patch
git add pyproject.toml
...
```

briantist:
> If you're planning on immediately merging this into master to create the 1.0.1 patch release

Thanks! Per our offline discussion, let's do the release in a separate PR, and we'll ensure #894 is included in the release

colin-pm (approved):
LGTM

PR #898, opened 2022-09-18
[WIP] Add `AdapterResponse` types, and `HvacAdapter`

### LINKED ISSUE #293: "No JSON object could be decoded" when writing to KV v2

### LINKED ISSUE #385: kv2 create_or_update_secret response.json() failse

### LINKED ISSUE #660: lookup_group returning Response objects

### LINKED ISSUE #797: The JSON adapter should return an empty response on 204 rather than the response object

briantist:
Closes #797 
Closes #293 
Closes #385 
Closes #660

This started after discussion about how we might handle 204 responses better in the `JSONAdapter` (see also #847), but led to us thinking about it in a new way that's a bit more comprehensive.

The basic idea is to return a specialized response class instead of raw data in some form or another.

The first members of this class are `raw`, `status`, and `value`, where `status` is meant to represent the HTTP status code returned from the request, `raw` is meant to be the raw response object (whatever it may be) or possibly `None` (this is Adapter-dependent), and `value` is meant to be the "usable" data/result of the request (also Adapter-dependent).

This is implemented here first as an abstract base class for adapter responses, an abstract subclass that is tuned for `requests.Response`-based response classes, and finally the `HvacAdapterResponse` which is first concrete response class.

`HvacAdapterResponse` is nominally very simple: it inherits from the `RequestsAdapterResponse` and only needs to override `value`; it takes similar steps to the `JSONAdapter` of trying to interpret the HTTP response as JSON, with an extra step for handling empty 204 responses. On a 204, an empty dict `{}` is returned.

If the response in not parseable as JSON, and the response code is not 204, then it returns `None`. If the caller knows how to handle the particular response, they can use the `.raw` member to get access to the response object and handle it however they wish.

`HvacAdapterResponse` includes an additional set of deprecated behaviors: it overrides `__getattr__`, `__getitem__`, `__setitem__`, and `__delitem__`, and proxies them to the `value` member instead. The purpose of this is so that code like `resp = client.secrets.kv2.read_secret_version(...)` which assumes that `resp` is a dict, can still work if `resp` is a `HvacAdapterResponse`; essentially letting consumers still treat it like a dict for a while. 

Each time this happens, a deprecation warning will be shown. Consuming code should change their calls like `resp.get('key', None)` or `resp['key']` to `resp.value.get('key', None)` or `resp.value['key']`, etc.

---

The next step is introducing a new Adapter, called the `HvacAdapter`. It operates rather transparently, based on `RawAdapter`, and simply returns an `HvacAdatperResponse` based on the `requests.Response` from the `RawAdapter`.

---

### v1.x
- Introduce these response types and new Adapter
- Update documentation and examples to show/recommend its use
- When a client is created with no explicit Adapter, **still default to JSONAdapter** but show a warning that the default will become `HvacAdapter` in v2.0.0
- Clients who explicitly want to retain `JSONAdatper` functionality can now explicitly specify it, and the warning will go away. `JSONAdapter` is not deprecated and their code will work indefinitely.

### v2.0.0
- Change the default Adapter when none is chosen to `HvacAdapter`. Responses are now of type `HvacAdapterResponse`
- Clients who ignored the warnings/recommendations about the default change _should still work_ due to the machinery in the response that allows to simulate a dict, but _all of those accesses should be spitting out deprecation warnings now_.
- To make the warnings go away, clients may update their code to use the `.value` member, or may choose to explicitly set the `JSONAdapter` (or another one of their choosing).

### v3.0.0
- Remove all of the fallback overrides that allows for treating the response as a dict.

---

This gives a lot of advance warning of the changes coming, and provides a lot of overlapping functionality so that downstream projects have time to support both ways so that their consumers can still use a wide range of `hvac` versions.

Opting into `HvacAdatper` is available from the beginning; opting out to stick with `JSONAdapter` is also available from the beginning.

---

**This is a draft, still WIP, and is subject to change.**


klukashka:
@briantist, why should `HvacAdapterResponse` behaviour be changed in the future versions? I suppose, it is more convenient to call by key than each time to mention a `value` in the same statement. Why should we deprecate it?

hvac/adapters.py
```diff
@@ -381,3 +383,173 @@ def request(self, *args, **kwargs):
...
+            f"Directly using `{repr(__o)} in response`",
+            replacement=f"{repr(__o)} in response.value",
+        )
+        return self.value.__contains__(__o)
+
+
+class HvacAdapter(RawAdapter):
+    """
+    The HvacAdapter adapter class.
+    This adapter interprets JSON responses similarly to the JSONAdapter, but it returns an HvacAdapterResponse object.
```
klukashka, line 531: Due to the fact that we inherit from `RawAdapter` and in `request` method we return `HvacAdapterResponse`, `JSONAdapter` should be the same in that method. In this case there will be no difference between `JSONAdapter` and `HvacAdapter`.
As we want to stay backward-compatible, we have to create something different from `RawAdapter`.

hvac/adapters.py
```diff
@@ -381,3 +383,173 @@ def request(self, *args, **kwargs):
...
+        return getattr(self.value, __name)
+
+    def __getitem__(self, __key: object) -> Any:
+        self._deprecate(
+            f"Directly accessing `response[{repr(__key)}]`",
+            replacement=f"response.value[{repr(__key)}]",
+        )
+        return self.value.__getitem__(__key)
+
+    def __setitem__(self, __key: object, __value: object) -> None:
```
klukashka, line 499: As our `HvacAdapterResponse` supposed to be mutable, shouldn't `raw` member of the class also be changed here? Cannot we get out-of-date information when calling to `raw`.

hvac/adapters.py
```diff
@@ -381,3 +383,173 @@ def request(self, *args, **kwargs):
...
+        if __name == "_value":
+            raise AttributeError
+
+        self._deprecate(
+            f"Directly accessing `response.{__name}`",
+            replacement=f"response.value.{__name}",
+        )
+        return getattr(self.value, __name)
+
+    def __getitem__(self, __key: object) -> Any:
```
klukashka, line 492: Isn't that risky to call this method on a `value` which can be `None`? We do that in many parts of code. Unfortunately, we should check that it isn't `None` every time with such an implementation.

PR #903, opened 2022-10-18
Add new field `auto_rotate_period` on transit key management

### LINKED ISSUE #902: Implement the new field `auto_rotate_period` on transit function
Ref documentation: https://developer.hashicorp.com/vault/api-docs/secret/transit#auto_rotate_period

Can we have the implementation of the new `auto_rotate_period` field related to the documentation?

To be implemented on the `create_key` and `update _key`

sebglon:
Implement new `auto_rorate_period` on transit `create_key` and `update_key`

Resolves #902

### DIFF
--- a/hvac/api/secrets_engines/transit.py
+++ b/hvac/api/secrets_engines/transit.py
@@ -60,7 +60,7 @@ class Transit(VaultApiBase):
         :type key_type: str | unicode
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
-        :param auto_rotate_period: The period at which this key should be rotated automatically.
+        :param auto_rotate_period: The period at which this key should be rotated automatically. Requires Vault 1.10.x or higher.
         :type auto_rotate_period: str | unicode
         :return: The response of the request.
         :rtype: requests.Response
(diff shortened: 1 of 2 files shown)


colin-pm:
Change overall LGTM, thanks for the contribution!

Can you run black and amend your commit?

@adammike @briantist This field appears to be added in v1.10.x. Thankfully it looks like using the parameter doesn't break the integration tests for <1.10.x, however should it be on us to mark when these fields were added?

briantist:
> This field appears to be added in v1.10.x. Thankfully it looks like using the parameter doesn't break the integration tests for <1.10.x, however should it be on us to mark when these fields were added?

Yes, we don't have to be perfect about it but when possible we should mark when a specific documented field was added to Vault, especially if it does not pre-date the versions of Vault we officially support (but we should still seek to do it even if it does).

sebglon:
@colin-pm Format is fixed
Do you need more change on my side?

briantist (dismissed):

hvac/api/secrets_engines/transit.py
```diff
@@ -59,6 +60,8 @@ def create_key(
         :type key_type: str | unicode
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
+        :param auto_rotate_period: The period at which this key should be rotated automatically.
```
briantist, line 63:
```suggestion
        :param auto_rotate_period: The period at which this key should be rotated automatically. Requires Vault 1.10.x.
```

hvac/api/secrets_engines/transit.py
```diff
@@ -201,6 +206,8 @@ def update_key_configuration(
         :type allow_plaintext_backup: bool
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
+        :param auto_rotate_period: The period at which this key should be rotated automatically.
```
briantist, line 209:
```suggestion
        :param auto_rotate_period: The period at which this key should be rotated automatically. Requires Vault 1.10.x.
```

briantist (approved):

briantist:
@sebglon thanks very much for this contribution!

PR #904, opened 2022-10-21
Replace Azure docs occurence to Kubernetes

gmsantos:
Update mentions of Azure mount path to Kubernetes and the signed JWT to Kubernetes service account


adammike (approved):
This change makes sense to me

PR #907, opened 2022-10-31
Allow for reading deleted secret versions (kv2) without an exception

### LINKED ISSUE #906: Reading a KV v2 secret version that has been deleted raises an `InvalidPath` exception
When a secret version is "deleted" KV v2, its data is soft-deleted. An attempt at retrieving the secret will return metadata, for example:

```console
# vault kv get -format=json kv2/versioned/twover
```
```json
{
  "request_id": "bf7bb5df-33ae-3c20-3479-93b3c4079fa3",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": {
    "data": null,
    "metadata": {
      "created_time": "2022-10-31T00:38:19.232734Z",
      "custom_metadata": null,
      "deletion_time": "2022-10-31T00:38:19.6945887Z",
      "destroyed": false,
      "version": 3
    }
  },
  "warnings": null
}
```

Here we can see that the version returned was version `3`, its `data` is `null`, and its `deletion_time` is set.

However, Vault returns this information as a 404 response body:
```console
$ curl -v --header "X-Vault-Token: 47542cbc-6bf8-4fba-8eda-02e0a0d29a0a" http://localhost:8200/v1/kv2/data/versioned/twover
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8200 (#0)
> GET /v1/kv2/data/versioned/twover HTTP/1.1
> Host: localhost:8200
> User-Agent: curl/7.58.0
> Accept: */*
> X-Vault-Token: 47542cbc-6bf8-4fba-8eda-02e0a0d29a0a
>
< HTTP/1.1 404 Not Found
< Cache-Control: no-store
< Content-Type: application/json
< Strict-Transport-Security: max-age=31536000; includeSubDomains
< Date: Mon, 31 Oct 2022 01:25:02 GMT
< Content-Length: 328
<
* Connection #0 to host localhost left intact
{"request_id":"b46d128d-947d-60c5-3a48-a4ced36c3240","lease_id":"","renewable":false,"lease_duration":0,"data":{"data":null,"metadata":{"created_time":"2022-10-31T00:38:19.232734Z","custom_metadata":null,"deletion_time":"2022-10-31T00:38:19.6945887Z","destroyed":false,"version":3}},"wrap_info":null,"warnings":null,"auth":null}
```

Since this is not an `ok` response, it's triggering the adapter's error handling and raising an exception:
https://github.com/hvac/hvac/blob/163956eb7a271d825d5a45abdfcf52fb0d6743ca/hvac/adapters.py#L313-L324

Because an exception is raised, the user has to catch the exception, try to convert its text from JSON, and if that succeeds, then check to see if it was a deleted version.

The user can make a specific custom request to the adapter with `raise_exception=False` but that option is generally not exposed from the secrets engines, so for example, calling `client.kv.v2.read_secret_version` does not give such an option.

Another alternative is to create an Adapter instance that uses `ignore_exceptions=True`, but since that is usually undesirable, it's kind of a pain to use since it requires instantiating a separate `Client` with that separate adapter, and using it with only certain calls.

Both of the above require additional work to handle actual exceptions then.

---

We should do something about this, but as of yet I'm unsure what. 

I may look into updating the kv2 `read_secret_version` function to make its call with `raise_exception=False`, and then have it process its own exceptions.

The main thing I'm concerned about there is that it'd be a breaking change, technically.

To address that, I might then propose an extra option that controls the handling; so something like `raise_on_deleted` which defaults to `True` (current behavior), but can be set to `False` to allow a "successful" return the way the CLI does, that way the user can determine how to handle the return of a deleted secret.

We can then optionally change the default in a newer version of `hvac` as a breaking change, if we want to.

briantist:
Fixes #906 

### Here's the final state of this PR
- `VaultError` (the base class for all the Vault exceptions) has been updated as follows:
  - constructor accepts optional `text` and `json` arguments and populates them as fields
  - [nicety] a new class method `from_status` is added to return an instance of a specific Vault exception, based on the given status code; this replaces the logic in `raise_for_error` to something a little cleaner
- `utils.raise_for_error` now accepts `text` and `json` fields to pass along to the exception classes, and is reduced to just calling `VaultError.from_status`
- `RawAdapter` (the only current place where `raise_for_status` is used) has been updated to populate the `text` and `json` fields as best it can from its `requests.Response`

All of the above now enables methods to remain Adapter-agnostic, but gain some additional information in the exceptions raised which can allow for handling of some conditions. With that:
- kv_v2's `read_secret_version` can catch the `InvalidPath` exception, inspect the `json` field to try to determine if the cause was due to a deleted secret version, and if so, can respond by returning the data instead of allowing the exception to be raised.
- that functionality is a (slightly) breaking change though, so it's controlled by a new parameter `raise_on_deleted_version`
  - Currently this parameter defaults to `None`
  - A value of `None` will default to `True` **(preserving the old behavior)** with a warning explaining that the default will change to `False` in `v3.0.0`
  - In `v3.0.0` we'll change the actual default to `False` and remove the warning code and the test for it

Since clients also gain from the new exception fields, sophisticated clients can decide how to handle a deleted version result in their application (whether by setting the new parameter, or catching exceptions), including:
* using additional calls to determine the next newest non-deleted secret (if they have permission to make those calls)
* "brute-forcing" the finding of a non-deleted version, by using the version number _V_ returned in the metadata, and requesting version _V-1_ until a non-deleted version is encountered or they get to `0`
* reporting a more appropriate status relevant to their use case

At this time, I don't think we should look to implement these workarounds in `hvac` itself.

Tests have also been added or improved, and all of the changes here have 100% test coverage.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -309,16 +309,31 @@ class RawAdapter(Adapter):
         )
 
         if not response.ok and (raise_exception and not self.ignore_exceptions):
-            text = errors = None
+            msg = json = text = errors = None
+            try:
+                text = response.text
+            except Exception:
+                pass
+
             if response.headers.get("Content-Type") == "application/json":
                 try:
-                    errors = response.json().get("errors")
+                    json = response.json()
                 except Exception:
                     pass
+                else:
+                    errors = json.get("errors")
+
             if errors is None:
-                text = response.text
+                msg = response.text
+
             utils.raise_for_error(
-                method, url, response.status_code, text, errors=errors
+                method,
+                url,
+                response.status_code,
+                msg,
+                errors=errors,
+                text=text,
+                json=json,
             )
 
         return response
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -324,7 +324,7 @@ class RawAdapter(Adapter):
                     errors = json.get("errors")
 
             if errors is None:
-                msg = response.text
+                msg = text
 
             utils.raise_for_error(
                 method,
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -237,6 +237,34 @@ class RawAdapter(Adapter):
     but always returns Response objects for requests.
     """
 
+    def _raise_for_error(self, method: str, url: str, response: requests.Response):
+        msg = json = text = errors = None
+        try:
+            text = response.text
+        except Exception:
+            pass
+
+        if response.headers.get("Content-Type") == "application/json":
+            try:
+                json = response.json()
+            except Exception:
+                pass
+            else:
+                errors = json.get("errors")
+
+        if errors is None:
+            msg = text
+
+        utils.raise_for_error(
+            method,
+            url,
+            response.status_code,
+            msg,
+            errors=errors,
+            text=text,
+            json=json,
+        )
+
     def get_login_token(self, response):
         """Extracts the client token from a login response.
 
@@ -309,32 +337,7 @@ class RawAdapter(Adapter):
         )
 
         if not response.ok and (raise_exception and not self.ignore_exceptions):
-            msg = json = text = errors = None
-            try:
-                text = response.text
-            except Exception:
-                pass
-
-            if response.headers.get("Content-Type") == "application/json":
-                try:
-                    json = response.json()
-                except Exception:
-                    pass
-                else:
-                    errors = json.get("errors")
-
-            if errors is None:
-                msg = text
-
-            utils.raise_for_error(
-                method,
-                url,
-                response.status_code,
-                msg,
-                errors=errors,
-                text=text,
-                json=json,
-            )
+            self._raise_for_error(method, url, response)
 
         return response
 
--- a/hvac/api/secrets_engines/kv_v2.py
+++ b/hvac/api/secrets_engines/kv_v2.py
@@ -2,7 +2,6 @@
 """KvV2 methods module."""
 from hvac import exceptions, utils
 from hvac.api.vault_api_base import VaultApiBase
-from requests import Response
 
 DEFAULT_MOUNT_POINT = "secret"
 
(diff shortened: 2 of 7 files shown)


briantist:
Just to update this with some behind-the-scenes discussion and thoughts I had: the fix currently proposed here won't be viable. The main reason is that it assumes the response from the adapter is a specific known type, so it assumes use of the library's default adapter (`JSONAdapter`), but would break with with any custom adapter.

In short: it breaks the encapsulation of the Adapter system.

This is a bigger issue than just for this one PR: it means that all of these methods, which defer to the configured adapter for requests, cannot really assume anything about the response.

The changes in #898 would add a defined set of types for Adapters to return, with a minimal set of properties that could be used in situations like this, but methods would still have to check that the response type subclasses from the base class to be able to assume those things, which I think is an ok compromise, but it could also lead to some unexpected behavior.

Another thing I'm looking at is updating the exceptions thrown so that they include the data of the response (which they currently lack). That will not allow for methods to infer anything about the response when no exceptions are raised, but it would allow them to handle raised exceptions with a little better insight, and should be usable for this kind of thing.

I think I will refactor a bit to make the current `raise_for_error()` function native to the Adapters where it's used (since it's `requests`-specific).

briantist:
The current state of the PR implements the changes I mentioned in my last comment, and a few more:

- `VaultError` (the base class for all the Vault exceptions) has been updated as follows:
  - constructor accepts optional `text` and `json` arguments and populates them as fields
  - [nicety] a new class method `from_status` is added to return an instance of a specific Vault exception, based on the given status code; this replaces the logic in `raise_for_error` to something a little cleaner
- `utils.raise_for_error` now accepts `text` and `json` fields to pass along to the exception classes, and is reduced to just calling `VaultError.from_status`
- `RawAdapter` (the only current place where `raise_for_status` is used) has been updated to populate the `text` and `json` fields as best it can from its `requests.Response`

Finally all of the above now enables methods to remain Adapter-agnostic, but gain some additional information in the exceptions raised which can allow for handling of some conditions. With that:
- kv_v2's `read_secret_version` can catch the `InvalidPath` exception, inspect the `json` field to try to determine if the cause was due to a deleted secret version, and if so, can respond by returning the data instead of allowing the exception to be raised.

---

## What's left

The current behavior is a breaking change: clients who expected an `InvalidPath` exception would now receive a response, albeit one with only metadata, and this could cause breakage.

I am considering adding a new parameter to `read_secret_version` (and probably `read_secret` as well), something like `raise_on_deleted_version` which can default to `True` to preserve the current behavior. Clients can set it to `False` to get the metadata back instead of an exception.

Since clients also gain from the new exception fields, sophisticated clients can decide how to handle a deleted version result in their application, including:
* using additional calls to determine the next newest non-deleted secret (if they have permission to make those calls)
* "brute-forcing" the finding of a non-deleted version, by using the version number _V_ returned in the metadata, and requesting version _V-1_ until a non-deleted version is encountered or they get to `0`
* reporting a more appropriate status relevant to their use case

At this time, I don't think we should look to implement these workarounds in `hvac` itself.

adammike (approved):
Really clean, looks good!

briantist:
@blatnoy7447 set `raise_on_deleted_version` to any explicit value and the warning will go away. If you want to preserve the current default behavior, set it to `True`, if you want the future default behavior, set it to `False`. See the original post for details.
If you have further questions please open a new [Q&A discussion](https://github.com/hvac/hvac/discussions/categories/q-a).

PR #909, opened 2022-11-02
Remove deprecated python syntax

dhuckins:
this PR removes older/deprecated python syntax from the code base
specifically:
- updates metaclass https://docs.python.org/3/library/2to3.html#to3fixer-metaclass
- remove old `abc.abstractproperty` decorator https://docs.python.org/3/library/abc.html#abc.abstractproperty

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -11,10 +11,8 @@ from hvac import utils
 from hvac.constants.client import DEFAULT_URL
 
 
-class Adapter:
+class Adapter(metaclass=ABCMeta):
     """Abstract base class used when constructing adapters for use with the Client class."""
-
-    __metaclass__ = ABCMeta
 
     def __init__(
         self,
--- a/hvac/api/system_backend/system_backend_mixin.py
+++ b/hvac/api/system_backend/system_backend_mixin.py
@@ -7,7 +7,5 @@ from hvac.api.vault_api_base import VaultApiBase
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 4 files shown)


briantist:
There's a problem with the way we do caching of the virtual environment in our workflows. I'm going to put up a PR for that and then I'd like this to be re-based so we can run the tests again.

briantist:
@dhuckins with #913 merged, would you mind rebasing from `develop` and pushing again?

dhuckins:
should be all set! (not sure what's happening with codecov here... 🤷 )

colin-pm:
Small nitpick, do you mind rebasing your two commits on top of develop? I'd like to avoid having any merge commits included in our development branch. Otherwise change lgtm, thanks for the PR!

briantist:
> Small nitpick, do you mind rebasing your two commits on top of develop? I'd like to avoid having any merge commits included in our development branch. Otherwise change lgtm, thanks for the PR!

I've disabled the merge commit method for this repository (hope that's ok) so that the default is now "Squash and merge" which will eliminate any merge commits in `develop`.

> should be all set! (not sure what's happening with codecov here... 🤷 )

I've been trying to figure out the coverage issues too. All I can tell so far is that the inexplicable coverage drops all seem to be in files related to enterprise features. I don't see why that would be happening though since the enterprise tests were turned off a while ago so those coverage drops should already be the baseline. Maybe when we re-enable those tests this issue will go away, but I can't be sure yet.

colin-pm (approved):

PR #913, opened 2022-11-13
update workflows

briantist:
Our venv caching was not quite doing the right thing. It was intended to include the python version in the cache key, but we incorrectly referenced the setup-python step when trying to get its outputs because we didn't assign that step an ID.

Figured it might also be a good time to switch to the action's built-in caching.

- update versions of some actions
- remove manual caching
- use setup-python action's built-in caching for poetry

Updating the versions also removes all warnings from the GHA runs.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -25,19 +25,19 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Set up Python ${{ matrix.python-version }}
-      id: python
-      uses: actions/setup-python@v4
-      with:
-        python-version: ${{ matrix.python-version }}
-        cache: poetry
-
     - name: Install Poetry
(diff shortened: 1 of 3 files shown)


briantist:
Codecov did not comment; I'm not certain why, but I've seen this on other repos. Lately they have been having problems with hitting GitHub API limits which has caused issues uploading reports. In this case, the reports were uploaded, but maybe a similar issue was hit when trying to post the comment.

In any case, coverage is not changed: https://app.codecov.io/gh/hvac/hvac/pull/913

---

Interestingly enough, perhaps the caching issues (restoring a cache from who knows which version of Python and using it with every version) are somehow what's been causing the weird coverage we've been seeing on other PRs (where it shows decreasing coverage on files not included).

adammike (approved):
LG2M

PR #915, opened 2022-11-19
CI: Add 3.11 support

colin-pm:
Adds 3.11 tests for build-test and lint-and-test pipelines.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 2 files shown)


dhuckins (dismissed):
lgtm!
can probably drop 3.6
https://endoflife.date/python

colin-pm:
> lgtm!
> can probably drop 3.6
> https://endoflife.date/python

We are aware of still supporting 3.6 while it's EOL. Based on prior discussions between @briantist and @adammike, we wanted to provide a 6-month window after announcing we will drop 3.6. That window ends this month and I plan to push up a change soon to drop 3.6 support.

It seems like running 3.6 on Ubuntu 22.04 is not supported, so I think it's safe to remove 3.6 from CI.

PR #916, opened 2022-11-20
don't cache on py3.6 windows combo

briantist:
Follow up to 
- #913 

For some reason this cache restores corrupted somehow and we keep having to delete it to get successful runs. Python 3.6 is getting removed soon, for now we'll just skip the cache on this one combo.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -37,7 +37,7 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


colin-pm (commented):
LGTM, but I would avoid merging this unless we are sure all the commits are squashed together.

colin-pm (approved):

briantist:
> LGTM, but I would avoid merging this unless we are sure all the commits are squashed together.

@colin-pm I disabled merge commits for PRs in this repo; the default is now "Squash and merge" and the only way to have a PR result in multiple commits in the base branch is to click the drop-down and choose "Rebase and merge" which I expect we'll never use. 

That should avoid us having to manually squash or accidentally merging intermediate commits :)

Issue #930, opened 2023-01-03
Unable to generate_credentials() for an AWS secret engine STS endpoint


sunchill06:
I get the following error when I try to get credentials from an sts endpoint. Below is the error.

Response

```json
{
  "errorMessage": "Error assuming role: InvalidParameter: 1 validation error(s) found.\n- minimum field value of 900, AssumeRoleInput.DurationSeconds.\n, on put https://vault.endpoint.com/v1/aws/sts/vault-policy",
  "errorType": "InvalidRequest",
  "stackTrace": [
    "  File \"/var/task/lambda_function.py\", line 54, in lambda_handler\n    print(hvac.api.secrets_engines.Aws.generate_credentials(client, 'lambda_china_staging', role_arn='arn:aws:iam::1234567890:role/VaultAssumeRoleDummy', endpoint='sts', ttl='3600s'))\n",
    "  File \"/var/task/hvac/api/secrets_engines/aws.py\", line 397, in generate_credentials\n    params=params,\n",
    "  File \"/var/task/hvac/adapters.py\", line 139, in put\n    return self.request(\"put\", url, **kwargs)\n",
    "  File \"/var/task/hvac/adapters.py\", line 369, in request\n    response = super().request(*args, **kwargs)\n",
    "  File \"/var/task/hvac/adapters.py\", line 336, in request\n    method, url, response.status_code, text, errors=errors\n",
    "  File \"/var/task/hvac/utils.py\", line 36, in raise_for_error\n    raise exceptions.InvalidRequest(message, errors=errors, method=method, url=url)\n"
  ]
}
```
Request: 
```python
hvac.api.secrets_engines.Aws.generate_credentials(client, 'lambda_china_staging', role_arn='arn:aws:iam::1234567890:role/VaultAssumeRoleDummy', endpoint='sts', ttl='3600s')
```

briantist:
Hi @sunchill06 , I'd like to clarify what's being attempted here and what's going wrong.

It seems like you're calling `hvac.api.secrets_engines.Aws.generate_credentials`, and passing in the `ttl` parameter, but the `ttl` is not being passed correctly to the endpoint in Vault, is that correct?

sunchill06:
[...]

PR #932, opened 2023-01-06
AWS secret engine generate credentials with ttl

sunchill06:
Signed-off-by: Sunil Aggarwal <sunil.aggarwal@commercetools.com>

Resolves https://github.com/hvac/hvac/issues/930

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -285,6 +285,11 @@ class RawAdapter(Adapter):
         if self.namespace:
             headers["X-Vault-Namespace"] = self.namespace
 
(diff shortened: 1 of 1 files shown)


briantist (changes requested):
Hi @sunchill06 thanks for opening this. Unfortunately this is not the correct place to make this change.

I suspect the correct place is [here](https://github.com/hvac/hvac/blob/develop/hvac/api/secrets_engines/aws.py#L327) but I'm going to comment on your bug report separately to make sure I understand the issue entirely.

PR #933, opened 2023-01-08
pin to ubuntu 20.04 until 3.6 is dropped

briantist:
Alternative option to #927 

I think we will only use one or the other, so: Closes #927

Fixes CI until we formally drop Python 3.6, at which point we can revert this.


adammike (approved):
LG2M

PR #939, opened 2023-01-16
Propagate client's adapter to API categories

WilkenSteiner:
Propagate the adapter used by the client to all handlers registered on the client.

Previously, setting the adapter (and thus session/connection pool) on the client after initialization would only replace the adapter owned by the client, leaving the adapters of all API categories untouched. 

### Notes

- If the current behaviour was by design, I will extend the docstrings instead. The new behaviour seems more intuitive to me, also more in line with other methods (e.g. the URL setter on the client).
- Added a regression test.

### DIFF
--- a/hvac/api/system_backend/__init__.py
+++ b/hvac/api/system_backend/__init__.py
@@ -75,9 +75,3 @@ class SystemBackend(
         Wrapping,
     ]
     unimplemented_classes = []
-
-    def __init__(self, adapter):
-        self._adapter = adapter
-
-    def __getattr__(self, item):
-        raise AttributeError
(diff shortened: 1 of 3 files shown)


briantist:
Hi @WilkenSteiner ! Thanks for this submission, I will have to take some time to go over it in a little more detail.

While the original designers are not active here anymore, off the top of my head, I would say that it was probably expected that the adapter would not b re-assigned after `Client` instantiation, though the fact that the `.adapter` setter was defined would make that confusing.

So I am inclined to agree that this is more intuitive, but let me think about it a bit and discuss with the other maintainers.

Also **huge thanks** for adding a regression test!

briantist:
@WilkenSteiner by the way, I've just created a poll and discussion topic about the Adapter system, and I'd appreciate a vote and/or some feedback if you have a few moments:
https://github.com/hvac/hvac/discussions/940

WilkenSteiner:
@briantist Thank you for considering and putting a more structural spin on it! I left a comment in the poll with my use case.

briantist:
@WilkenSteiner would you rebase this? The CI should be fixed now

briantist (approved):

PR #942, opened 2023-01-24
doc: update reference to removed method

### LINKED ISSUE #900: 'Client.renew_self_token' method is lost
https://github.com/hvac/hvac/blob/v1.0.2/hvac/v1/__init__.py#L360
was changed to work only with `/auth/token/renew` Vault endpoint and **renew_self_token** is mentioned there to be used instead. But there is no such method.

Maybe it was replaced by https://github.com/hvac/hvac/blob/v1.0.2/hvac/api/auth_methods/token.py#L348
But it means that at least docstring is wrong and should be updated.

bendem:
Resolves #900

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -375,7 +375,7 @@ class Client:
         :return:
         :rtype:
 
-        For calls expecting to hit the renew-self endpoint please use the "renew_self_token" method instead
+        For calls expecting to hit the renew-self endpoint please use the "renew_self" method on "hvac_client.auth.token" instead
         """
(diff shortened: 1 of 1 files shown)


adammike:
This looks appropriate to me, @briantist do you have any thoughts?

briantist:
Also looks good to me! 
@bendem could I trouble you to rebase? The CI should be fixed in #933

bendem:
Sorry for the delay, was off the last few days. PR is rebased now. ;)

briantist (approved):

PR #943, opened 2023-02-11
fix vault client certificates loaded from envirnoment variables

### LINKED ISSUE #787: Client certs are incorrectly handled
#735 broke use of client certs by overriding the passed in value, specifically https://github.com/hvac/hvac/blob/3e3ae1c9ff3cf73e4debb024cbe5f65e7a436086/hvac/v1/__init__.py#L76-L82

The code should be:

```
        if cert is None and VAULT_CLIENT_CERT:
            cert = [VAULT_CLIENT_CERT, VAULT_CLIENT_KEY]
```

This concatenated string used will never work with the requests library. See https://github.com/psf/requests/blob/39d0fdd9096f7dceccbc8f82e1eda7dd64717a8e/requests/adapters.py#L242-L254

### LINKED ISSUE #938: VAULT_CLIENT_CERT will always override cert param
```
        if cert is not None and VAULT_CLIENT_CERT:
            cert = "\n".join(
                [
                    VAULT_CLIENT_CERT,
                    VAULT_CLIENT_KEY,
                ]
            )
```

I expect the param `cert` code to take priority over the environment variable `VAULT_CLIENT_CERT`. 
This is correctly followed for the `verify` param. `verify` param will take priority over `VAULT_CAPATH`

At least this logic should be consistent for both.

BrandonHoffman:
It seems like the hvac library does not handle the VAULT_CLIENT_CERT and VAULT_CLIENT_KEY variables properly. 

- public/private key-pairs in requests need to be passed as a tuple 
- environment variables are overriding the values directly passed in

Fixes #938 
Fixes #787

### DIFF
--- a/tests/utils/__init__.py
+++ b/tests/utils/__init__.py
@@ -49,7 +49,6 @@ def is_enterprise():
 
 def if_vault_version(supported_version, comparison=operator.lt):
     current_version = get_installed_vault_version()
-    print(current_version)
     return comparison(Version(current_version), Version(supported_version))
 
 
--- a/tests/utils/__init__.py
+++ b/tests/utils/__init__.py
@@ -49,6 +49,7 @@ def is_enterprise():
 
 def if_vault_version(supported_version, comparison=operator.lt):
     current_version = get_installed_vault_version()
+    print(current_version)
     return comparison(Version(current_version), Version(supported_version))
 
 
@@ -96,15 +97,13 @@ def create_client(url="https://localhost:8200", use_env=False, **kwargs):
     client_key_path = get_config_file_path("client-key.pem")
     server_cert_path = get_config_file_path("server-cert.pem")
     if use_env:
-        with (
-            mock.patch("hvac.v1.VAULT_CAPATH", server_cert_path),
-            mock.patch("hvac.v1.VAULT_CLIENT_CERT", client_cert_path),
-            mock.patch("hvac.v1.VAULT_CLIENT_KEY", client_key_path),
-        ):
-            client = Client(
-                url=url,
-                **kwargs,
-            )
+        with mock.patch("hvac.v1.VAULT_CAPATH", server_cert_path):
+            with mock.patch("hvac.v1.VAULT_CLIENT_CERT", client_cert_path):
+                with mock.patch("hvac.v1.VAULT_CLIENT_KEY", client_key_path):
+                    client = Client(
+                        url=url,
+                        **kwargs,
+                    )
     else:
         client = Client(
             url=url,
--- a/tests/utils/__init__.py
+++ b/tests/utils/__init__.py
@@ -8,10 +8,10 @@ import re
 import socket
 import subprocess
 from distutils.spawn import find_executable
-from packaging.version import Version
-from unittest import SkipTest
+from unittest import SkipTest, mock
 
 from hvac import Client
+from packaging.version import Version
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


BrandonHoffman:
https://github.com/hvac/hvac/issues/938

and 

https://github.com/hvac/hvac/issues/787

are both related

briantist:
@BrandonHoffman welcome and thanks for submitting this! I've been away traveling but I intend to take a look at the backlog some more after I get back and get caught up on other things, so it might take a little while but I will review this!

In the meantime run `petry run black .` and `poetry run flake8 .` to check for some formatting lint errors.

BrandonHoffman:
@briantist - no worries, just know we were hitting issues with this in my org and figured it would effect others. The fmt/lint issue is fixed.

BrandonHoffman:
also fixed an integration test issue related to integration tests running in python 3.7 which does not support the nested with syntax I initially used.

adammike (changes requested):
I can't imagine how that original code worked, with the list split into a string like that. @briantist  I've reviewed this. The code looks good, and I've validated that what's being passed to the adapter is what requests is expecting. 

Rather than change all tests to use the environment, we should add new tests for this use case specifically - mTLS authentication with the certs passed to the client as parameters, and mtls with the certs passed in the environment.

Another issue I have, and this is not aimed at this committer, we are implicitly passing authentication parameters from the adapter to the requests session. While functional and technically not a bug, Its best practice to be implicit with authentication and authorization parameters.

adammike:
The issue with the implicit auth parameters does not need to be fixed in this PR, just pointing out the issue. It should probably be tracked as a security enhancement bug.

BrandonHoffman:
Hey @adammike,

toggling everything to use the environment variables was actually a mistake...... accidentally checked in the global toggle to true that I used locally just to validate everythin was working.

I actually only intended for this one subclass to be the test case for using the environment variables since it seemed like it would be enough to validate this flow.

https://github.com/hvac/hvac/pull/943/files#diff-f12f973ead75ad35d6886ad63a746f8a3c2f540b4ea538845ac6346353a65a74R158

adammike:
Looks good to me!

tests/utils/__init__.py
```diff
@@ -79,26 +79,38 @@ def get_generate_root_otp():
...
-    return Client(
-        url=url,
-        cert=(client_cert_path, client_key_path),
-        verify=server_cert_path,
-        **kwargs,
-    )
+    if use_env:
+        with mock.patch("hvac.v1.VAULT_CAPATH", server_cert_path):
+            with mock.patch("hvac.v1.VAULT_CLIENT_CERT", client_cert_path):
+                with mock.patch("hvac.v1.VAULT_CLIENT_KEY", client_key_path):
```
briantist, line 101: I might consider using `with mock.patch.dict` here against `os.env`, that way you can provide a dict of env vars to override.

BrandonHoffman: @briantist i don’t think that will work since the vault environment variables are read at import time of the constants module before this test case patch would be applied. Also if it did work it would end up leaving the global variables in that state after the test completes.

briantist: Thanks that  makes sense. Opened https://github.com/hvac/hvac/issues/945 to consider whether this should change in the future.

briantist (approved):
This all looks good! Thanks for this fix @BrandonHoffman .

The only thing I want to consider before merging is whether we're able to release this in a minor version or whether we should wait until a major out of an abundance of caution.

It's technically a breaking change. It should only hit someone who:
* is passing values for these both directly and in environment variables
* is passing the desired values in environment variables
* is passing undesired values directly

While the remedy is clear for such a use case, the change is still unexpected for that user.

In any case we'll get this out as soon as we can!

briantist:
@adammike is there anything else you'd like to review here? 
If not, could you approve or dismiss your previous review?

Any thoughts on the breaking change consideration I posted above?

briantist:
Fixes #938 
Fixes #787 

@BrandonHoffman @adammike I don't see any reason why merging this shouldn't close the issues above, do you agree?

BrandonHoffman:
I think both those issues should be addressed with this.

briantist:
> from what I have seen any case where the environment variables are set results in a traceback so I don’t believe that anyone using the library as it is now could use the environment variable without having A trace back occur. That leads me to believe it should be safe, but maybe there is some edge case I’m not considering.

Ah that's right, I forgot that the env code as is wouldn't have worked anyway, so I think you're right, and that we can safely call this a bugfix, thanks again!

BrandonHoffman:
Yeah no worries thank you for being so welcoming.

tests/utils/__init__.py
```diff
@@ -8,10 +8,10 @@
 import socket
 import subprocess
 from distutils.spawn import find_executable
-from packaging.version import Version
-from unittest import SkipTest
+from unittest import SkipTest, mock
 
 from hvac import Client
+from packaging.version import Version
```
dhuckins, line 14: `packaging` is a dep for various packages but should still add it as a dev dep

briantist: good catch! Added in [54543a4](https://github.com/hvac/hvac/pull/943/commits/54543a41f02cad790530bfbd394016d3e1990b9d).

briantist (approved):

briantist (approved):

adammike (approved):
LG2M

PR #949, opened 2023-02-19
docs - sphinx - fail on warnings

### LINKED ISSUE #946: Documentation tests should fail on sphinx warnings
For example:
- #775 
- https://github.com/hvac/hvac/actions/runs/4213669243/jobs/7313540329#step:8:88

In this run, the submitted `userpass` documentation was not added to the toctree, so it will not actually show up in documentation, but the tests pass.

There are other warnings that we probably need to address in order to fail on warnings, but there aren't too many... 

It would also be great if we had rendered documentation previews on PRs. I've done this in the Ansible world, maybe I can put together something similar here 🤔

briantist:
Resolves #946 

The docs build tests don't fail on warnings, and it means we can miss some pretty egregious stuff. 

This PR will attempt to fail on warnings, and then fix any outstanding warnings we have so that we're in a good state to catch these things on future PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@ Breakfix release to revert some unintended post-1.0 requirements changes.
 
 - **Note**: This is _actually and truly_ (😝)  intended to by the last hvac release supporting Python 2.7.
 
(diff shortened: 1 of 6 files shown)


briantist:
@dhuckins thanks for reviewing this!

Issue #958, opened 2023-03-08
Provide a way to opt-out of implicit token loading


briantist:
When a Vault client is created with a token, it will try to find a token via the `VAULT_TOKEN` environment variable or the token sink file, much like the Vault CLI.

This can cause unexpected results:
- https://github.com/ansible-collections/community.hashi_vault/issues/13#issuecomment-740192580
- https://github.com/hvac/hvac/issues/937

This can be worked around by performing a `.logout()` call from the client after it's created.

We should at a minimum provide a more explicit way to opt out of this behavior on Client creation.

I might go so far as to say we should consider changing the default behavior in a future version, and make this opt-in.

Issue #937, opened 2023-01-13
auth/userpass is expecting a namespace in the token


d-bourdon:
Hello,

I have a bug when I try to auth with the auth/userpass endpoint.

```Traceback (most recent call last):
  File "/home/dbourdon/Documents/vault_client.py", line 16, in <module>
    vault_client.auth.userpass.login("simvault-dev", password)
  File "/home/dbourdon/.local/lib/python3.10/site-packages/hvac/api/auth_methods/userpass.py", line 141, in login
    return self._adapter.post(
  File "/home/dbourdon/.local/lib/python3.10/site-packages/hvac/adapters.py", line 110, in post
    return self.request('post', url, **kwargs)
  File "/home/dbourdon/.local/lib/python3.10/site-packages/hvac/adapters.py", line 353, in request
    response = super(JSONAdapter, self).request(*args, **kwargs)
  File "/home/dbourdon/.local/lib/python3.10/site-packages/hvac/adapters.py", line 315, in request
    utils.raise_for_error(
  File "/home/dbourdon/.local/lib/python3.10/site-packages/hvac/utils.py", line 47, in raise_for_error
    raise exceptions.InternalServerError(message, errors=errors, method=method, url=url)
hvac.exceptions.InternalServerError: error performing token check: failed to look up namespace from the token: no namespace, on post https://vault.*********/v1/auth/userpass/login/*******```
```
I've confirmed the credential but using directly the HTTP API with curl, and the vault server is returning a token that works.

## To reproduce
Hvac=v1.0.2, Vault server=1.10.1
```python
import hvac

user="****"
password="****"

vault_client = hvac.Client(url="https://vault.*******", namespace=None) #tried also to not specify the namespace
vault_client.auth.userpass.login(user, password)
```

## Current Workaround
To mitigate that for now, I send the same credential via the REST api, parse the response and init hvac with the given token
```python

import json
import os
import hvac
from urllib import request

vault_client = hvac.Client(url="https://vault.************", namespace=None)

password="****"
user="****"

payload = {"password": password}
url = f"https://vault.*******/v1/auth/userpass/login/{user}"
req = request.Request(url, json.dumps(payload).encode("utf-8"))
req.add_header('Content-Type', 'application/json')

with request.urlopen(req) as response:
    body = response.read()

decode_body = json.loads(body.decode("utf-8"))
tmp_token = decode_body["auth"]["client_token"]

vault_client.token = tmp_token
[...]

PR #960, opened 2023-03-09
Feature: Add VAULT_SKIP_VERIFY environment variable

fe80:
Hello,

Actually the environment variable [VAULT_SKIP_VERIFY](https://developer.hashicorp.com/vault/docs/commands#vault_skip_verify) is not supported by the library:

```python
Python 3.10.9 (main, Dec 19 2022, 17:35:49) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getenv("VAULT_SKIP_VERIFY")
'true'
>>> import hvac
>>> hvac.Client().is_authenticated()
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/connectionpool.py", line 386, in _make_request
    self._validate_conn(conn)
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/connectionpool.py", line 1042, in _validate_conn
    conn.connect()
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/connection.py", line 414, in connect
    self.sock = ssl_wrap_socket(
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/util/ssl_.py", line 453, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(sock, context, tls_in_tls)
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/util/ssl_.py", line 495, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock)
  File "/usr/lib/python3.10/ssl.py", line 513, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/lib/python3.10/ssl.py", line 1071, in _create
    self.do_handshake()
  File "/usr/lib/python3.10/ssl.py", line 1342, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)

During handling of the above exception, another exception occurred:

  File "/home/steffy/.pip/lib/python3.10/site-packages/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    retries = retries.increment(
  File "/home/steffy/.pip/lib/python3.10/site-packages/urllib3/util/retry.py", line 592, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='127.0.0.1', port=8200): Max retries exceeded with url: /v1/auth/token/lookup-self (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)')))

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/home/steffy/.pip/lib/python3.10/site-packages/hvac/v1/__init__.py", line 408, in is_authenticated
    self.lookup_token()
  File "/home/steffy/.pip/lib/python3.10/site-packages/hvac/v1/__init__.py", line 329, in lookup_token
    return self._adapter.get(path, wrap_ttl=wrap_ttl)
  File "/home/steffy/.pip/lib/python3.10/site-packages/hvac/adapters.py", line 112, in get
    return self.request("get", url, **kwargs)
  File "/home/steffy/.pip/lib/python3.10/site-packages/hvac/adapters.py", line 356, in request
    response = super().request(*args, **kwargs)
  File "/home/steffy/.pip/lib/python3.10/site-packages/hvac/adapters.py", line 305, in request
    response = self.session.request(
  File "/home/steffy/.pip/lib/python3.10/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/steffy/.pip/lib/python3.10/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/home/steffy/.pip/lib/python3.10/site-packages/requests/adapters.py", line 563, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: HTTPSConnectionPool(host='127.0.0.1', port=8200): Max retries exceeded with url: /v1/auth/token/lookup-self (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)')))
>>> 
```

This feature add the support of this one

```python
╰─➤ python
Python 3.10.9 (main, Dec 19 2022, 17:35:49) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getenv("VAULT_SKIP_VERIFY")
'true'
>>> import hvac
>>> hvac.Client().is_authenticated()
/home/steffy/.pip/lib/python3.10/site-packages/urllib3/connectionpool.py:1045: InsecureRequestWarning: Unverified HTTPS request is being made to host '127.0.0.1'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  warnings.warn(
True
```

Sorry I've open and close a bad request before (#959)

Regards,


briantist:
Hi @fe80 ! Thanks for opening this.

I still need to discuss with the other maintainers but I think we will stop adding new Vault CLI environment variables to the library, and may look to scale back or remove some of the existing ones.

`hvac` is a library used by other software and not a CLI, so the use of values from the environment implicitly in the library can end up causing unexpected behavior in software that uses `hvac`.

For a recent example, see:
- #958 
- #937 

In most cases, it's pretty easy for software authors to decide if they want to use environment variables themselves and pass them in, that way we're not making that decision for them.

In this case, passing it in during Client construction:

```python
import os
from hvac import Client

client = Client(verify=os.getenv('VAULT_SKIP_VERIFY"))
```

---

I'll leave this open for now, because I haven't yet written up a complete set of thoughts on this nor discussed with other maintainers.

fe80:
Hello,

Thank for your reply, it's too bad thank doesn't work like ruby library as [default value](https://github.com/hashicorp/vault-ruby/blob/master/lib/vault/defaults.rb).

>  hvac is a library used by other software and not a CLI, so the use of values from the environment implicitly in the library can end up causing unexpected behavior in software that uses hvac.

All software are suppose to be run on a specific context with limited environment variable no ?

I'm not using `hvac` like a CLI but with multiple scripts running on different network context. Maybe I should create a new library for initialize `hvac.Client` with all environment variable support ?

Regards,

briantist:
> Hello,
> 
> Thank for your reply, it's too bad thank doesn't work like ruby library as [default value](https://github.com/hashicorp/vault-ruby/blob/master/lib/vault/defaults.rb).
> 
> > hvac is a library used by other software and not a CLI, so the use of values from the environment implicitly in the library can end up causing unexpected behavior in software that uses hvac.
> 
> All software are suppose to be run on a specific context with limited environment variable no ?

Sure, I think the difference is when you as a user run software X which uses library Y, you expect to know the behavior based on what X tells you; typically you should not have to be figuring out how X uses Y and then reading the code or documentation for Y to understand how X is going to behave in your environment.

If the author of X is not even aware, they do not report what could be causing behavior changes for you, and it makes their job somewhat harder.

In this case, if `hvac` uses environment variables by default, it's harder for software X to control its behavior and offer a consistent experience, but it is straightforward for X to read the environment variables it wants to respond to, and pass them into `hvac`.

In short: the author(s) of X should make those decisions, and the authors of Y should mostly leave that alone.

Just my opinion, having now been on both sides of this library and gotten bit by this before.

> I'm not using `hvac` like a CLI but with multiple scripts running on different network context. Maybe I should create a new library for initialize `hvac.Client` with all environment variable support ?

You may create such a library, yes! Or subclass the `Client`, or however you might like to implement it. Doing so would ensure you get a consistent experience for what you need it for. I think that's a better approach, personally.

If we were going to add any support for new ones, I would probably make them opt-in, with the default being not to load them, while I also look to at least add opt-out capability to some of the older ones.

Tylerlhess:
While I don't think pulling in all environment variables by default is what HVAC wants I could get behind an override kwarg in the client initialization that gratuitously gets such envvars. Just thinking out loud for a workable solution that allows everyone to remain working off the same branch.

briantist:
> While I don't think pulling in all environment variables by default is what HVAC wants I could get behind an override kwarg in the client initialization that gratuitously gets such envvars. Just thinking out loud for a workable solution that allows everyone to remain working off the same branch.

@Tylerlhess 
Could you give an example of what that would look like? Would it be much different than the caller supplying the direct value by way of retrieving it from env vars themselves, like:

```python
client = Client(verify=os.getenv('VAULT_SKIP_VERIFY"))
```

Tylerlhess:
@briantist I would do something like

`client = Client(useEnvVars=true)`

This would allow those using envvars to have it get any and all envvars that are set.

briantist:
Hello! The `develop` branch has recently had some changes, and your branch will need to be rebased and pushed again. Thanks and sorry for the inconvenience.

briantist:
Hi @fe80 ,

Re: the PR itself, I'm still leaning toward not introducing (and possibly removing some) environment variable support. But I will leave this open until there's a definitive policy/decision in the project.

PR #966, opened 2023-03-26
fix wrapped response for `auth.token.create`

### LINKED ISSUE #965: `auth.token.create` does not use `wrap_ttl` parameter if a `role_name` is specified
https://github.com/hvac/hvac/blob/develop/hvac/api/auth_methods/token.py#L119-L122

When `role_name` is specified, the API path must be modified, but in this case the `wrap_ttl` parameter is not passed, unlike the code path where `role_name` is not specified:

https://github.com/hvac/hvac/blob/develop/hvac/api/auth_methods/token.py#L125-L129

This leaves us with no way to get a wrapped token that's created with a role.

briantist:
Fixes #965 

- Adds tests for wrapped responses
- Ensures role-based tokens can be wrapped

(first run expected to fail: test added to catch the condition, fix not applied yet)

### DIFF
--- a/hvac/api/auth_methods/token.py
+++ b/hvac/api/auth_methods/token.py
@@ -111,17 +111,11 @@ class Token(VaultApiBase):
             }
         )
 
-        if role_name is not None:
-            api_path = "/v1/auth/{mount_point}/create/{role_name}".format(
-                mount_point=mount_point,
-                role_name=role_name,
-            )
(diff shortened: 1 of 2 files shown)


blmhemu (approved):

adammike (approved):
LG2M, I also think the code looks much cleaner. Great job.

PR #972, opened 2023-04-11
feat: add support for `disable_local_ca_jwt` in Kubernetes auth method

fad3t:
Hi,

First of all thanks for developing this great project :)
I would like to propose my first contribution, about adding the support for the `disable_local_ca_jwt` in the Kubernetes auth method (cf. https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#disable_local_ca_jwt).

This has been tested and it looks good; please let me know if anything else is needed/missing.

Thanks!
Fred


hvac/api/auth_methods/kubernetes.py
```diff
@@ -24,6 +24,7 @@ def configure(
         token_reviewer_jwt=None,
         pem_keys=None,
         issuer=None,
+        disable_local_ca_jwt=False,
         mount_point=DEFAULT_MOUNT_POINT,
```
briantist, line 28: we should switch this order, so that we don't break positional argument calls

```suggestion
        mount_point=DEFAULT_MOUNT_POINT,
        disable_local_ca_jwt=False,
```

briantist (commented):
Hi @fad3t welcome and thanks for putting this up.

The change looks pretty straightforward to me. I know you mentioned testing it, would be able to update the automated tests in this PR to include this functionality?

fad3t:
> Hi @fad3t welcome and thanks for putting this up.
> 
> The change looks pretty straightforward to me. I know you mentioned testing it, would be able to update the automated tests in this PR to include this functionality?

Done, hopefully this is good enough 😄 Thanks for the review!

briantist:
Looks like the tests are failing; you may want to get set up so you can run the tests locally, that way you don't rely on the CI (since you're a new contributor, I have to approve each push before it runs). This is an ok starting point to get set up: https://hvac.readthedocs.io/en/stable/contributing.html

Let me know if you have issues running the tests locally.

fad3t:
> Looks like the tests are failing; you may want to get set up so you can run the tests locally, that way you don't rely on the CI (since you're a new contributor, I have to approve each push before it runs). This is an ok starting point to get set up: https://hvac.readthedocs.io/en/stable/contributing.html
> 
> Let me know if you have issues running the tests locally.

I ran the tests locally before, I'm getting some kind of timeouts during the LDAP tests. Anyway I explicitly targeted the tests related to the k8s auth methods and they were OK.
The problem I had was related to the error message, which changed in Vault 1.9.x. My tests were running OK locally because I'm using Vault 1.11.x - which is inline with the error message I had in my test. But in the pipeline which starts with Vault 1.7.x I see the message is a bit different, so I adapted it to cover those too.

fad3t:
@briantist can you advise on the failing test? It doesn't seem related to the proposed change (dep installation on Windows env). Thx!

briantist:
I agree 100%, it's likely not related to this PR at all, but I haven't had a chance to dig into yet. The reason it fails on Windows 3.6 and _not_ others is probably because dependency caching in CI is disabled on that combination due to a bug (#916), so my guess is that there are probably failures on either everything, or perhaps just 3.6 (but not just Windows). I'll have to delete the caches to find out.

Anyway this is a repo problem and I'll take a look at it as soon as I can. Thanks!

briantist:
@fad3t please rebase and push again to incorporate the change in #979 which fixes CI

fad3t:
> @fad3t please rebase and push again to incorporate the change in #979 which fixes CI

Done, thanks!

briantist:
@fad3t The `develop` branch has recently had some changes, and your branch will need to be rebased and pushed again. In this case, a merge may not work correctly, so please try to do it with a rebase. Thanks and sorry for the inconvenience.

fad3t:
> @fad3t The `develop` branch has recently had some changes, and your branch will need to be rebased and pushed again. In this case, a merge may not work correctly, so please try to do it with a rebase. Thanks and sorry for the inconvenience.

Hi @briantist, done -- I hope it's OK.

briantist:
> > @fad3t The `develop` branch has recently had some changes, and your branch will need to be rebased and pushed again. In this case, a merge may not work correctly, so please try to do it with a rebase. Thanks and sorry for the inconvenience.
> 
> Hi @briantist, done -- I hope it's OK.

Unfortunately, the wrong commits are still included, see: https://github.com/hvac/hvac/pull/972/commits

As I mentioned, doing a merge will not work in this case, it must be a rebase.

And now that a merge has been included, you will likely have to interactive rebase and remove the offending commits:
```
# save the current state of the branch in a new branch
git checkout -b save-branch
# switch back to your original
git checkout -
# interactive rebase
git rebase -i develop
```

If you're not sure, I can pull your branch down and do this, but I'm not sure when I'll have time so it might be a while.

fad3t:
Hi @briantist, I just gave it a try but I don't think it worked -- I'm a bit confused 😄 
Maybe it would be easier if I start over from a new branch, and redo a new pull request?

briantist:
> I just gave it a try but I don't think it worked -- I'm a bit confused 😄 Maybe it would be easier if I start over from a new branch, and redo a new pull request?

I don't think we need to go that far yet. When I get a chance (might be a few weeks out) I'll see if I can pull this down and fix the branch.

Issue #989, opened 2023-06-01
transit sign_data doesn't take the hash_algorithm in consideration


deidax:
When running the following script:
`
sign_data_response = client.secrets.transit.sign_data(
    name='key-name',
    mount_point=app-transit',
    hash_algorithm='sha2-512',
    hash_input=hash_input=base64ify('hi its me hvac')
)
`

I get the following error

`hvac.exceptions.Forbidden: 1 error occurred: *permission denied on post http://mywebsite.com/v1/app-transit/sign/key-name
`

briantist:
The description for this parameter says that it can _also_ be specified in the URL, not that it must be. The method seems to already submit it in parameters, so I'm wondering if maybe something else is going on here, or if the description of the parameter in HashiCorp's documentation is incorrect: https://developer.hashicorp.com/vault/api-docs/secret/transit#sign-data

Do you have any more info that might be helpful in reproducing this error?

deidax:
When I try using the sign data function with the sha2-512 algorithm like this:
```
import hvac
client = hvac.Client(url='https://127.0.0.1:8200')

key_name = 'hvac-signing-key'
sign_data_response = client.secrets.transit.sign_data(
    name=key_name,
    mount_point=my_mount-point,
    hash_algorithm='sha2-512',
    hash_input=base64ify('hi its me hvac'),
)
signature = sign_data_response['data']['signature']
print('Signature is: {signature}'.format(signature=signature))
```
I get the following error
```
Forbidden: 1 error occurred:
    * permission denied
    on post http://127.0.0.1:8200/v1/key-name/sign/my_mount-point
```

briantist:
@deidax thanks for the additional info. I've set up a Vault dev server to test this out, and what I've found is that `hvac`'s way of setting `hash_algorithm` does work.

I was able to confirm that it worked with both the path and parameter-based methods, and confirm that if the hash algorithm is set in both the path and as a parameter, the parameter is ignored. This was most easily tested by setting invalid algorithm names like `sha9`

```
$ vault write transit/sign/key2/sha1 plaintext=$(echo "abc" | base64) hash_algorithm=sha9
Key          Value
---          -----
[...]

PR #990, opened 2023-06-01
adding batch_input to transit.sign_data #988

### LINKED ISSUE #988: batch_input attribute is not available for sign_data
it seems that **Transit.sign_data** doesn't accept a batch_input attribute like in the **Transit.encrypt_data** even though the Vault API accepts batch_input for **/transit/sign/:name(/:hash_algorithm)**

deidax:
Resolves #988 by adding the `batch_input` to the `transit sign_data method` and updating the params.

~Resolve the issue #989 by checking if the value of hash_algorithm is None `hash_algorithm = '' if not hash_algorithm else hash_algorithm` and updating the api_path `"/v1/{mount_point}/sign/{name}/{hash_algorithm}"`~


Pipfile
briantist, line 1: This file doesn't appear related to this change, could you explain what it is or remove it?

hvac/api/secrets_engines/transit.py
```diff
@@ -757,6 +757,7 @@ def sign_data(
         hash_input,
         key_version=None,
         hash_algorithm=None,
+        batch_input=None,
```
briantist, line 760: Please move this to the end of the argument list, otherwise it will break folks using positional arguments.

hvac/api/secrets_engines/transit.py
```diff
@@ -873,13 +874,18 @@ def sign_data(
                     "signature_algorithm": signature_algorithm,
                     "marshaling_algorithm": marshaling_algorithm,
                     "salt_length": salt_length,
+                    "batch_input": batch_input
```
briantist, line 877:
```suggestion
                    "batch_input": batch_input,
```

hvac/api/secrets_engines/transit.py
```diff
@@ -873,13 +874,18 @@ def sign_data(
...
         )
+        
+        hash_algorithm = '' if not hash_algorithm else hash_algorithm
+        
         api_path = utils.format_url(
-            "/v1/{mount_point}/sign/{name}",
+            "/v1/{mount_point}/sign/{name}/{hash_algorithm}",
             mount_point=mount_point,
             name=name,
+            hash_algorithm=hash_algorithm
```
briantist, line 888:
```suggestion
            hash_algorithm=hash_algorithm,
```

briantist (commented):
thanks for submitting this! I've left a few small comments, but I still need to look at the underlying issues more deeply. I'm a little overloaded at the moment, but I will get back to it at some point.

Pipfile
deidax: thank you for pointing this out, the Pipfile is not related to this change

briantist:
Hello @deidax ! 

I will come back to review the actual content as soon as I can, thanks for your patience!

hvac/api/secrets_engines/transit.py
```diff
@@ -763,6 +763,7 @@ def sign_data(
         marshaling_algorithm=None,
         salt_length=None,
         mount_point=DEFAULT_MOUNT_POINT,
+        batch_input=None,
```
briantist, line 766: This needs a docstring entry below so that it will be documented properly. The documentation for the API should help with the description: https://developer.hashicorp.com/vault/api-docs/secret/transit#batch_input-4

hvac/api/secrets_engines/transit.py
```diff
@@ -873,13 +874,18 @@ def sign_data(
...
+                    "batch_input": batch_input,
                 }
             )
         )
+
+        hash_algorithm = "" if not hash_algorithm else hash_algorithm
+
         api_path = utils.format_url(
-            "/v1/{mount_point}/sign/{name}",
+            "/v1/{mount_point}/sign/{name}/{hash_algorithm}",
```
briantist, line 885: The description for this parameter says that it can _also_ be specified in the URL, not that it must be. The method seems to already submit it in parameters, so I'm not certain why we would require it to be in the URL. Let's continue this in the issue report: 
- https://github.com/hvac/hvac/issues/989

If you can update the tests in a way that demonstrates the issue, it would also be very helpful.

briantist (commented):
Thanks for submitting this! As commented below, I'd like to get more information on the `hash_algorithm` issue.

For `batch_input`, it needs docs, and I'd really like to see the tests updated for this, is that something you could take a look at?

hvac/api/secrets_engines/transit.py
deidax: > The description for this parameter says that it can _also_ be specified in the URL, not that it must be. The method seems to already submit it in parameters, so I'm not certain why we would require it to be in the URL. Let's continue this in the issue report:
> 
> * [transit sign_data doesn't take the hash_algorithm in consideration #989](https://github.com/hvac/hvac/issues/989)
> 
> If you can update the tests in a way that demonstrates the issue, it would also be very helpful.

hi @briantist , that's true but I got this issue when I was working on a project that needed the hash algorithm to be passed in the URL and after looking at Vault documentation here https://developer.hashicorp.com/vault/api-docs/secret/transit#hash-data I noticed that the hash algorithm is passed in the encryption URL. The changes I made here worked for my case.

briantist:
@deidax could you please revert the merge of the `develop` branch into the PR? 
Both of these commits should be reverted or removed via interactive rebase:
- https://github.com/hvac/hvac/pull/990/commits/38378d695cb7fcd3a887929024fba1b8e68b03c8
- https://github.com/hvac/hvac/pull/990/commits/eca6563cbab62c3c77207d42027c4a285706d217

As well as the 4 commits shown here that have only me as the author:
- https://github.com/hvac/hvac/pull/990/commits

See here for why: https://github.com/hvac/hvac/pull/990#issuecomment-1595858305

hvac/api/secrets_engines/transit.py
briantist: Let's continue this in the issue report. If you prefer, you can also change this PR to only add `batch_input` and then a separate PR could be used for the other thing once we figure out what's going on, or we can work that out first, whatever works for you.

deidax:
> @deidax could you please revert the merge of the `develop` branch into the PR? Both of these commits should be reverted or removed via interactive rebase:
> 
> * [38378d6](https://github.com/hvac/hvac/commit/38378d695cb7fcd3a887929024fba1b8e68b03c8)
> * [eca6563](https://github.com/hvac/hvac/commit/eca6563cbab62c3c77207d42027c4a285706d217)
> 
> As well as the 4 commits shown here that have only me as the author:
> 
> * https://github.com/hvac/hvac/pull/990/commits
> 
> See here for why: [#990 (comment)](https://github.com/hvac/hvac/pull/990#issuecomment-1595858305)
> 
> If you need help, I can try to do this and force push your branch.

@briantist Yes I would appreciate it if you can do it, because I'm a little lost here 😅

hvac/api/secrets_engines/transit.py
deidax: @briantist let's change this PR to `batch_input` for now.

deidax:
> Thanks for submitting this! As commented below, I'd like to get more information on the `hash_algorithm` issue.
> 
> For `batch_input`, it needs docs, and I'd really like to see the tests updated for this, is that something you could take a look at?

@briantist Before running any tests, please consider the following: According to the vault documentation for the `transit.sign_data` function, if `batch_input` is set, `hash_input` should be ignored. Currently, the function does not handle this case properly, as `hash_input` is a required parameter. To use `batch_input` instead, the user should set `hash_input=None`.

hvac/api/secrets_engines/transit.py
```diff
@@ -861,6 +868,9 @@ def sign_data(
                     allowed_types=transit_constants.ALLOWED_SALT_LENGTHS.pattern,
                 )
             )
+        
+        hash_input = None if batch_input is not None else hash_input
```
deidax, line 872: To ensure compatibility, I included the following line to set the value of `hash_input` to `None` when `batch_input` is set, as both values cannot be simultaneously used.

briantist:
> > @deidax could you please revert the merge of the `develop` branch into the PR? Both of these commits should be reverted or removed via interactive rebase:
> > 
> > * [38378d6](https://github.com/hvac/hvac/commit/38378d695cb7fcd3a887929024fba1b8e68b03c8)
> > * [eca6563](https://github.com/hvac/hvac/commit/eca6563cbab62c3c77207d42027c4a285706d217)
> > 
> > As well as the 4 commits shown here that have only me as the author:
> > 
> > * https://github.com/hvac/hvac/pull/990/commits
> > 
> > See here for why: [#990 (comment)](https://github.com/hvac/hvac/pull/990#issuecomment-1595858305)
> > If you need help, I can try to do this and force push your branch.
> 
> Yes I would appreciate it if you can do it

Sure, I'm going to attempt this now. Since this operation can lose data, you may want to create a new branch off of your current one locally in case you need to recover commits later.

briantist:
> > > @deidax could you please revert the merge of the `develop` branch into the PR? Both of these commits should be reverted or removed via interactive rebase:
> > > 
> > > * [38378d6](https://github.com/hvac/hvac/commit/38378d695cb7fcd3a887929024fba1b8e68b03c8)
> > > * [eca6563](https://github.com/hvac/hvac/commit/eca6563cbab62c3c77207d42027c4a285706d217)
> > > 
> > > As well as the 4 commits shown here that have only me as the author:
> > > 
> > > * https://github.com/hvac/hvac/pull/990/commits
> > > 
> > > See here for why: [#990 (comment)](https://github.com/hvac/hvac/pull/990#issuecomment-1595858305)
> > > If you need help, I can try to do this and force push your branch.
> > 
> > 
> > Yes I would appreciate it if you can do it
> 
> Sure, I'm going to attempt this now. Since this operation can lose data, you may want to create a new branch off of your current one locally in case you need to recover commits later.

@deidax I believe this was successful! I also pushed an additional commit where I ran black to fix the formatting/lint error.

**For this PR, please do not use the GitHub UI update branch button. You can ignore the message telling you that your branch is behind.**

What's happening is that you have a `develop` branch which tracks our `develop` branch, but we're not using that one anymore. I previously _rebased_ your `develop` branch to actually be based on our `main` branch.

I think this will not be a problem for your future PRs (I hope there will be more 😁), which should be based off the `main` branch to begin with.

I'm very sorry for the confusion about this, I know it's not a good experience especially for a first-time contributor to a project, it was just bad timing on my part.

hvac/api/secrets_engines/transit.py
briantist: Since we are changing this PR to be for `batch_input` only, please undo these `hash_algorithm` changes for now

hvac/api/secrets_engines/transit.py
briantist: Thanks for pointing this out. I am thinking about whether we should continue to have `hash_input` be a required parameter (argument) in the function, or we should default it to `None`, and then have code to raise an exception if `hash_input is None and batch_input is None`.

Basically, I am thinking it might be an awkward user experience when using `batch_input` if you are forced to explicitly set `hash_input=None`.

Since you would presumably be using this exact scenario, what do you think about that?

deidax: I think it would be better set both parameters to None by default. In case the user fails to provide a value for any of the parameters, an exception can be raised with message guiding the user to set the necessary values.

hvac/api/secrets_engines/transit.py
```diff
@@ -805,6 +806,13 @@ def sign_data(
         :type salt_length: str | unicode
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
+        :param batch_input: Specifies a list of items for processing.
+            When this parameter is set, if the parameter 'hash_input' is also set, it will be ignored.
+            Responses are returned in the 'batch_results' array component of the 'data' element of the response.
+            Any batch output will preserve the order of the batch input.
+            If the input data value of an item is invalid, the corresponding item in the 'batch_results'
+            will have the key 'error' with a value describing the error.
```
briantist, line 814:
```suggestion
            This parameter is mutually exclusive with the ``hash_input`` parameter, but one of them must be supplied. 
            If both are set, or neither are set, an exception will be raised.
            Responses are returned in the ``batch_results`` array component of the ``data`` element of the response.
            Any batch output will preserve the order of the batch input.
            If the input data value of an item is invalid, the corresponding item in the ``batch_results``
            will have the key ``error`` with a value describing the error.
```

Updating to match new behavior. Please add something similar to the first two lines to the description for `hash_input` as wel.

briantist (commented):
Just some wording and formatting suggestions

hvac/api/secrets_engines/transit.py
```diff
@@ -773,7 +774,9 @@ def sign_data(
 
         :param name: Specifies the name of the encryption key to use for signing. This is specified as part of the URL.
         :type name: str | unicode
-        :param hash_input: Specifies the base64 encoded input data.
+        :param hash_input: This parameter is mutually exclusive with the ``batch_results`` parameter, but one of them must be supplied. 
+            If both are set, or neither are set, an exception will be raised.
+            ``hash_input`` Specifies the base64 encoded input data.
```
briantist, line 779: reordering a little to put the actual description first
```suggestion
        :param hash_input: Specifies the base64 encoded input data.
            This parameter is mutually exclusive with the ``batch_results`` parameter, but one of them must be supplied. 
            If both are set, or neither are set, an exception will be raised.
```

hvac/api/secrets_engines/transit.py
```diff
@@ -805,6 +808,14 @@ def sign_data(
         :type salt_length: str | unicode
         :param mount_point: The "path" the method/backend was mounted on.
         :type mount_point: str | unicode
+        :param batch_input: Specifies a list of items for processing.
+            This parameter is mutually exclusive with the ``hash_input`` parameter, but one of them must be supplied. 
+            If both are set, or neither are set, an exception will be raised.
+            Responses are returned in the ``batch_results`` array component of the ``data`` element of the response.
+            Any batch output will preserve the order of the batch input.
+            If the input data value of an item is invalid, the corresponding item in the ``batch_results``
+            will have the key ``error`` with a value describing the error.
```
briantist, line 817: some more re-ordering to put description first and align with the other param.
```suggestion
        :param batch_input: Specifies a list of items for processing.
            Any batch output will preserve the order of the batch input.
            If the input data value of an item is invalid, the corresponding item in the ``batch_results``
            will have the key ``error`` with a value describing the error.
            This parameter is mutually exclusive with the ``hash_input`` parameter, but one of them must be supplied. 
            If both are set, or neither are set, an exception will be raised.
            Responses are returned in the ``batch_results`` array component of the ``data`` element of the response.
```

briantist (commented):
a few more docs re-ordering, I'll go ahead and commit these, and if you want to change anything with the edits, feel free to push more commits

briantist (approved):
This looks good to me! Thanks so much for submitting and sticking through the edits and suggestions.

Since I was the last committer, I'll wait for you to look at those changes before merging. Let me know if it all looks good to you.

deidax:
> This looks good to me! Thanks so much for submitting and sticking through the edits and suggestions.
> 
> Since I was the last committer, I'll wait for you to look at those changes before merging. Let me know if it all looks good to you.

@briantist yeah it looks good to me. thank you

PR #998, opened 2023-06-16
support "user_claim_json_pointer" in create_role() for JWT auth method

ferenc-hechler:
When creating a role for the JWT auth method, the optional parameter "user_claim_json_pointer" is missing.
See documentation here: 

https://developer.hashicorp.com/vault/api-docs/auth/jwt#user_claim_json_pointer

The parameter is a bool value which defaults to false.
This pull request adds the missing parameter. I tested it for JWT auth and it works.
The parameter can be bool or a str containing "true" or "false".
If a wrong parameter is given, the server returns a corresponding error message.

The OIDC auth mehod inherits create_role() from JWT. 
So, I also added this parameter to ODIC.create_role(). 
The OIDC use case was not tested.

Documentation was updated.

### DIFF
--- a/hvac/api/auth_methods/jwt.py
+++ b/hvac/api/auth_methods/jwt.py
@@ -179,7 +179,7 @@ class JWT(VaultApiBase):
         :type bound_audiences: list
         :param user_claim: The claim to use to uniquely identify the user; this will be used as the name for the
             Identity entity alias created due to a successful login. The interpretation of the user claim
-            is configured with user_claim_json_pointer. If set to True, user_claim supports JSON pointer syntax 
+            is configured with user_claim_json_pointer. If set to True, user_claim supports JSON pointer syntax
             for referencing a claim. The claim value must be a string.
         :type user_claim: str | unicode
         :param clock_skew_leeway: Only applicable with "jwt" roles.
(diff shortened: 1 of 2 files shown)


briantist:
Hi @ferenc-hechler ! Welcome and thanks for submitting this. I hope to look this over a more thoroughly as soon as I get a little time.

In the meantime, please ensure you set up your local environment and run the lint/format before pushing:
```
poetry run black .
poetry run flake8 .
```

Let me know if you need a hand with that.

---

I am curious, why make the parameter accept strings in addition to bool? The API documentation seems to expect a bool only, and since we are passing through the value directly, we aren't doing any conversion or checking of it.

My instinct is to make this bool only. _If_ the API is documented as accepting a string, then I could see us accepting either, though I might still lean toward it being bool only in `hvac`.

I don't use this API currently so if I'm missing something please let me know.

Thanks again!

ferenc-hechler:
Hi @briantist ,
thanks for your feedback. 

I formatted the code as you proposed.

About the additional "str" type: 
I removed this. I saw, that there is "int | str" for other parameters and thought this is a pattern, to also allow "str". But looking at "token_no_default_policy" is also a "bool" without "str" alternative.

briantist:
Thank you @ferenc-hechler that makes sense.

ferenc-hechler:
looks like my idea to switch the branch names from main and develop was not successful.
I will create a new pull request from the main branch.

Issue #877, opened 2022-09-13
Drop EOL Python versions


colin-pm:
At the end of 2022, EOL Python versions will be dropped. Starting in 2023, HVAC will track with the CPython Release lifecycle. This decision was made to track with dependencies that track the CPython Release lifecycle. For reference, the next dates are listed below.

- Python 3.6 will be dropped 31 Dec 2022
- Python 3.7 will be dropped 27 Jun 2023
- Python 3.8 will be dropped 14 Oct 2024

Further EOL information can be found here: https://endoflife.date/python

briantist:
It's been a while since we commented on this issue. I will add that **we will remove Python versions in major version releases of `hvac` only** unless there is a compelling reason we need to do so earlier.

_New versions of Python_ may become supported in minor releases.

## Update on schedule
Since we're behind schedule and the 3.7 EoL date is approaching, **we will drop support for both Python 3.6 and Python 3.7** in `v2.0.0`. 

Python 3.11 support is expected to be added in the same release (we cannot add it earlier due to dependency conflicts with Python 3.6)
- #915 

The `v2.0.0` milestone can be followed here:
- https://github.com/hvac/hvac/milestone/34

PR #1005, opened 2023-06-17
Add support for Python 3.11

briantist:
Resolves #915 

- add Python 3.11 to the test matrix

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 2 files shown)


briantist:
The failure in CI for pytest (specifically, it's the coverage part) in 3.11 is due to needing `coverage` 6.3 or higher. This comment helped me find that: https://github.com/pytest-dev/pytest-cov/issues/564#issuecomment-1418329497

I was able to confirm in a local environment that that worked.

We currently use `coverage` 6.2, and we can't jump to 6.3 yet because it doesn't support python 3.6.

**Effectively, we can't add python 3.11 support until we drop 3.6** which _is_ planned for the next major version: 
- https://github.com/hvac/hvac/issues/877

briantist:
Closing in favor of #1048

PR #1009, opened 2023-06-20
Add a method to read static roles in the database engine

mimato:
Currently, the database backend does not provide a way to read static roles, only dynamic roles. This PR adds a method to read static roles as well. 

Vault API doc: https://developer.hashicorp.com/vault/api-docs/secret/databases#read-static-role

### DIFF
--- a/hvac/api/secrets_engines/database.py
+++ b/hvac/api/secrets_engines/database.py
@@ -261,6 +261,25 @@ class Database(VaultApiBase):
             url=api_path,
         )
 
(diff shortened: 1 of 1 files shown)


briantist (approved):
Hi @mimato , welcome and thanks very much for this contribution! I was going to ask if you might be able to add tests for the new method... and then realized we don't have tests for this engine at all 😔 so I can't reasonably ask for that as a pre-requisite on this PR!

The code looks great though, if you're interested [we have lots more we can use help with](https://github.com/hvac/hvac/discussions/994), thanks again!

PR #1010, opened 2023-06-20
docs: make OIDC Authorization URL Request example work again.

robbat2:
- Import `urllib.parse` so parse works.
- Document how this interacts with OIDC allowed_redirect_uris & roles.
- Python used a `return` outside of any function, add a `main()` function to keep the code clean.
- Fix `odic` mis-spelling.
- Add a nicer HTML redirect page

Signed-off-by: Robin H. Johnson <rjohnson@coreweave.com>

### DIFF
--- a/docs/usage/auth_methods/jwt-oidc.rst
+++ b/docs/usage/auth_methods/jwt-oidc.rst
@@ -142,36 +142,81 @@ OIDC Authorization URL Request
 
 :py:meth:`hvac.api.auth_methods.JWT.oidc_authorization_url_request`
 
+This is equivalent to `vault login -method=oidc`
+
 .. code:: python
 
     import webbrowser
     import http.server
     import hvac
-    client = hvac.Client()
+    import urllib.parse
 
-    auth_url_response = client.auth.oidc.oidc_authorization_url_request(
-        role='hvac',
-        redirect_uri='http://localhost:8250/oidc/callback'
-    )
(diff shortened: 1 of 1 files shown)


docs/usage/auth_methods/jwt-oidc.rst
```diff
@@ -142,36 +142,76 @@ OIDC Authorization URL Request
...
+        token = login_oidc_get_token()
+
+        auth_result = client.auth.oidc.oidc_callback(
+            code=token,
+            path='oidc',
+            nonce=auth_url_nonce,
+            state=auth_url_state,
+        )
+
+        print('Client token returned: %s' % auth_result['auth']['client_token'])
```
briantist, line 211: Since we're updating this example, and using an f-string already on line 160, what do you think about updating this line to use an f-string as well?

briantist (approved):
Hi @robbat2 , welcome! Thanks for submitting this improvement to our docs. I've left an optional inline comment but this otherwise looks great!

docs/usage/auth_methods/jwt-oidc.rst
robbat2: Fixed to f-string.

briantist (approved):
@robbat2 thanks again for updating these docs! 

If you haven't seen this yet, we're actively looking for more contributors; we'd love to see more if you're interested and able:
- https://github.com/hvac/hvac/discussions/994

PR #1015, opened 2023-06-25
add tests and docs to sdist, improve build testing

### LINKED ISSUE #1011: Add tests to pypi tarball
Hello team,
while working on updating hvac version FreeBSD port [1], we noticed that tarball downloaded from pypi.org does not have test suite, which is instead available on this github repo.

Ideally, we'd like to have that included also in pypi file: is this something you can consider adding?

[1] https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272060

Let us know if you need anything else.
Thank you

briantist:
- update build tests
  - test both wheel and sdist installation
  - upload the build artifacts to CI run so they can be downloaded by interested parties
- set sdist to include tests and docs, resolves #1011

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
-        os: [ubuntu-20.04, macOS-latest, windows-latest]
+        os: [ubuntu-20.04, macos-latest, windows-latest]
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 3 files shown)


briantist:
Hi @asagratini this PR adds the tests (and also docs) to the sdist tarball. I also updated our CI so that it uploads build artifacts for all the build tests, so you can actually download the dist files for those CI runs now and ensure that they will work for you packaging needs.

For example, here's the latest run as of this writing, and you should be able to download any of the artifacts listed (they should all be the same even though they are generated on different platforms):
- https://github.com/hvac/hvac/actions/runs/5367407986

If you could let me know that it meets your needs I would appreciate it. The 1.2.0 release is probably going to be the last before 2.0.0 and I'd like to get this in before the next major version.

Tylerlhess (approved):

briantist:
Hi @asagratini , just checking in one more time to see if you can confirm this will meet your packaging needs.

asagrati (approved):

asagrati:
> Hi @asagratini , just checking in one more time to see if you can confirm this will meet your packaging needs.

Hi @briantist, yes, it looks good, thank you!

PR #1016, opened 2023-06-25
Versioning updates

briantist:
When we switched to poetry in #854 , we replaced instructions for using `bumpversion` (`bump2version`) with `poetry version`, however, we left around the config file for `bumpversion`, and the poetry command did not update versions in all the places that were needed.

As a result, our docsite version has been "stuck" on `v1.0.2` for a while in some places (not the content, just the title and maybe other areas).

In this PR, I'm leaning back on `bumpversion` for now:
- set `bump2version` as an explicit dependency
- update the config to: 
  - update version in `pyproject.toml`
  - don't do a commit or tag automatically
- manually update the wrong versions in docs config so they'll be updated correctly next time
- update the release instructions

---

In the future, I'd like to consider [`poetry-bumpversion`](https://pypi.org/project/poetry-bumpversion/) as a replacement. This is a plugin to `poetry version` that can be used to make updates in other files.

I think this might fit the poetry ecosystem better, and it also has nicer options for pre- and post- releases, something I might want to look into.

But, we can't bring this package in right now while we're still on Python 3.6, so it would have to be a post-2.0 change.


Tylerlhess (approved):

briantist:
thanks for reviewing @Tylerlhess !

PR #1020, opened 2023-06-29
Fix for issue 991

Tylerlhess:
Adding a fix for session variables not overriding the default adapter settings.


Tylerlhess:
It looks like this section of code isn't in the tests already. I will try to find time to add that but not sure when I will have the time to provide a good use case and test situation.

Issue #991, opened 2023-06-07
Custom certificate verification fails with requests>=2.28.0


satwell:
With versions of requests starting with 2.28.0, I get a requests.exceptions.SSLError exception when trying to connect to a Vault instance that uses a custom SSL certificate.  Example code:

```
import hvac
import requests

s = requests.Session()
s.verify = "mycert.crt"
c = hvac.Client(session=s)
print(c.sys.read_health_status())
```

This works correctly with requests 2.27.1.  But with 2.28.0, I get the following stack trace:

```
Traceback (most recent call last):
  File "/home/satwell/tmp/hvac/status.py", line 7, in <module>
    print(c.sys.read_health_status())
  File "/home/satwell/tmp/hvac/venv/lib/python3.9/site-packages/hvac/api/system_backend/health.py", line 68, in read_health_status
    return self._adapter.head(
  File "/home/satwell/tmp/hvac/venv/lib/python3.9/site-packages/hvac/adapters.py", line 175, in head
    return self.request("head", url, **kwargs)
  File "/home/satwell/tmp/hvac/venv/lib/python3.9/site-packages/hvac/adapters.py", line 372, in request
    response = super().request(*args, **kwargs)
  File "/home/satwell/tmp/hvac/venv/lib/python3.9/site-packages/hvac/adapters.py", line 331, in request
    response = self.session.request(
  File "/home/satwell/tmp/hvac/venv/lib/python3.9/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/satwell/tmp/hvac/venv/lib/python3.9/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/home/satwell/tmp/hvac/venv/lib/python3.9/site-packages/requests/adapters.py", line 517, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: HTTPSConnectionPool(host='vault.example.com', port=8200): Max retries exceeded with url: /v1/sys/health (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1123)')))
```

This doesn't seem like a bug in requests.  I can use the same requests.Session with `verify` set to successfully make a request to the same server.

I bisected changes to requests, and it looks like this problem starts with psf/requests@79c4a01.

satwell:
[...]

PR #1021, opened 2023-06-30
adapters: if `session` is user-supplied, do not overwrite session options with `Client`/`Adapter` options

Tylerlhess:
Fixes: #991

# Breaking change
For more details see also:
- https://github.com/hvac/hvac/issues/1040

Before this change:
* if you do not use a custom `session`, a session is created for you with values for the `verify`, `cert`, and `proxies` settings that are set in the adapter, either explicitly or by default
* if you _do_ use a custom `session`:
  * the values in the `session` will always be overwritten with the adapter values, whether you specified them or not

After this change:
* if you do not use a custom `session`, a session is created for you with values for the `verify`, `cert`, and `proxies` settings that are set in the adapter, either explicitly or by default
* if you _do_ use a custom `session`:
  * if any value is set for the `verify`, `cert`, or `proxies` properties in your custom `session`, then that value will be kept and the adapter's value will be ignored, even if you set it on the adapter explicitly
  * if any of those properties _are not set_ then they will be set based on the adapter's value for those

The only case where this should cause breakage of existing code is where all of the following are true:
* you use a custom session
* you set any of the affected properties on the custom session
* the value(s) you set on those properties are actually invalid for your use case, and so your code relied on the adapter defaults overriding your values

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -90,6 +90,15 @@ class Adapter(metaclass=ABCMeta):
         if not session:
             session = requests.Session()
             session.cert, session.verify, session.proxies = cert, verify, proxies
+        # fix for issue 991 using session verify if set
+        else:
+            if session.verify:
+                # need to set the variable and not assign it to self so it is propperly passed in kwargs
+                verify = session.verify
+            if session.cert:
+                cert = session.cert
+            if session.proxies:
+                proxies = session.proxies
 
         self.base_uri = base_uri
         self.token = token
--- a/tests/unit_tests/test_adapters.py
+++ b/tests/unit_tests/test_adapters.py
@@ -46,6 +46,11 @@ class TestAdapters:
         ],
     )
     def test_from_adapter(self, conargs):
+        session = requests.Session()
+        session.cert = conargs["cert"]
+        session.verify = conargs["verify"]
+        session.proxies = conargs["proxies"]
+        conargs["session"] = session
         expected = conargs.copy()
         for internal_kwarg in self.INTERNAL_KWARGS:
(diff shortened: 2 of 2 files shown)


briantist:
Hey @Tylerlhess thanks for this. I haven't looked at it fully yet, but I wanted to mention I opened #1023 to try to deal with our out-of-date Vault versions in the integration tests.

My hope is we can get all the issues ironed out in there so that you don't have to add these unrelated changes into your PR. I've already borrowed your azure fix (with attribution) in https://github.com/hvac/hvac/pull/1023/commits/b01e796550de3288e36ffb686f65475318ff1ad7 though since we're also not pinning to `X.Y.0` in that PR anymore, it seems like some changes may have made it into patch versions of Vault, so we may need to be more nuanced in the version compares now.

I really appreciate you bringing these issues to our attention!

briantist:
With #1023 & #1024 merged, those should fix the Azure failures, SSH failures you saw locally, and the coverage issues, so you can rebase against or merge from the latest `main` branch HEAD, and remove your azure changes.

I was going to do this for you and push up the branch but for some reason it looks a little weird to me, when I try to rebase your branch all of the files are "conflicts" even the ones you didn't touch. Trying to merge `main` into your branch refuses due to "unrelated histories" which is probably the same reason for rebase weirdness.

I think I can resolve that anyway and force push, so if you want me to go ahead with that please let me know, otherwise I'll leave it you.

---

The other changes in the PR look good to me, the only other thing I'd like to do is add warnings for users who are affected, telling them that in a future major version we'll stop applying any overrides from the affected parameters when a session is supplied.

I can also add that stuff in alongside the work you've done already.

briantist (approved):

PR #1023, opened 2023-07-02
expand Vault CI matrix, announce deprecation of Vault dynamic SSH keys

briantist:
# Vault Dynamic SSH Keys

This feature was deprecated in Vault and was removed in Vault 1.13.0. Although we have not decided a version for removal yet, the support for it in `hvac` will be removed at some point; likely at the same time we decide to drop support for older versions of Vault. 

Once we have a decision and timeline for supported versions of Vault, a more specific announcement can be made.

---

Thanks to @Tylerlhess for pointing out that our Vault integration matrix is quite a bit behind.

In this PR I'm attempting to expand it.

- use a wildcard for the patch version component
- adding 1.12.x-1.14x
- turns off fail-fast on lint, unit, and integration tests (we want to know if failures only exist in a certain combination)
- turn on colorized output for pytest in CI
- fixing breakages due to backports of changes into new minor versions
- skipping SSH tests for features removed in later Vault versions
- some minor coverage changes following on the heels of #1024 

~I will also look at expanding the enterprise versions, since those are really old...~
Using newer enterprise versions is a more complicated matter which I'll try to address in a different PR.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -165,14 +165,17 @@ jobs:
         python-version:
           - 3.7
         vault-version:
-          - "vault-enterprise=1.6.5+ent"
-          - "vault-enterprise=1.7.2+ent"
-          - "vault=1.6.5"
-          - "vault=1.7.2"
-          - "vault=1.8.0"
-          - "vault=1.9.0"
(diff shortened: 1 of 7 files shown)


adammike (approved):
Looks Great, Thanks!

PR #1028, opened 2023-07-18
add `Client.write_data` method

### LINKED ISSUE #133: Cannot write to keys named "path"
Discovered this trying to query the [`sys/capabilities`](https://www.vaultproject.io/api/system/capabilities.html) endpoint, which takes both `path` and `token` as it's required arguments. 

We may wish to use something that isn't `**kwargs` in `Client.write`

```
In [9]: obj = {'path': 'sys/leader', 'token': client.token}

In [10]: client.write('sys/capabilities', **obj)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-10-b2eab5c7a994> in <module>()
----> 1 client.write('sys/capabilities', **obj)

TypeError: write() got multiple values for keyword argument 'path'
```

M0NsTeRRR:
## If you are here to understand the changes as a user, please see:
- https://github.com/hvac/hvac/issues/1034

---

Hello,

This PR fix #133.

I suggest this implementation (name can be changed but i've avoided `write_v2` to not lead confusion with kv v1 and v2) to handle backward compatibility.

V2 : We add this function
V3 : We replace the old one and delete this function
or
V2 : We add this function
V3 : We replace the old one and redirect this function (to keep both for a certain time, for example until v4)
VX: We remove this function

Regards,

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -265,6 +265,23 @@ class Client:
         """
         return self._adapter.post(f"/v1/{path}", json=kwargs, wrap_ttl=wrap_ttl)
 
+    def write_fix_path(self, path, data={}, wrap_ttl=None):
+        """Same as write() method but fix issue #133 (you can pass `path` key in POST data)
+        
+        Supported methods:
+            POST /<path>
+
+        :param path:
+        :type path:
+        :param data:
+        :type dict:
+        :param wrap_ttl:
+        :type wrap_ttl:
+        :return:
+        :rtype:
+        """
+        return self._adapter.post(f"/v1/{path}", json=data, wrap_ttl=wrap_ttl)
+
     def delete(self, path):
         """DELETE /<path>
 
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -256,7 +256,7 @@ class Client:
 
         Write data to a path. Because this method uses kwargs for the data to write, "path" and "wrap_ttl" data keys cannot be used.
         If these names are needed, or if the key names are not known at design time, consider using the write_data method.
-        
+
         :param path:
         :type path:
         :param wrap_ttl:
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -267,7 +267,7 @@ class Client:
 
     def write_fix_path(self, path, data={}, wrap_ttl=None):
         """Same as write() method but fix issue #133 (you can pass `path` key in POST data)
-        
+
         Supported methods:
             POST /<path>
 
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -254,6 +254,9 @@ class Client:
     def write(self, path, wrap_ttl=None, **kwargs):
         """POST /<path>
 
+        Write data to a path. Because this method uses kwargs for the data to write, "path" and "wrap_ttl" data keys cannot be used.
+        If these names are needed, or if the key names are not known at design time, consider using the write_data method.
+        
         :param path:
         :type path:
         :param wrap_ttl:
@@ -265,8 +268,8 @@ class Client:
         """
         return self._adapter.post(f"/v1/{path}", json=kwargs, wrap_ttl=wrap_ttl)
 
-    def write_fix_path(self, path, data={}, wrap_ttl=None):
-        """Same as write() method but fix issue #133 (you can pass `path` key in POST data)
+    def write_data(self, path, *, data={}, wrap_ttl=None):
+        """Write data to a path. Similar to write() without restrictions on data keys.
 
         Supported methods:
             POST /<path>
--- a/tests/integration_tests/v1/test_integration.py
+++ b/tests/integration_tests/v1/test_integration.py
@@ -143,18 +143,12 @@ class IntegrationTest(HvacIntegrationTestCase, TestCase):
         self.client.sys.disable_auth_method("userpass")
 
     def test_write_fix_path(self):
-        self.client.write_fix_path(
-            "secret/foo",
-            data={
-                "path": "foo1",
-                "foo": "foo2"
-            }
-        )
+        self.client.write_fix_path("secret/foo", data={"path": "foo1", "foo": "foo2"})
         result = self.client.read("secret/foo")
 
         assert result["data"]["path"] == "foo1"
         assert result["data"]["foo"] == "foo2"
-        
+
         self.client.delete("secret/foo")
 
     def test_missing_token(self):
--- a/tests/integration_tests/v1/test_integration.py
+++ b/tests/integration_tests/v1/test_integration.py
@@ -142,8 +142,8 @@ class IntegrationTest(HvacIntegrationTestCase, TestCase):
         self.client.token = self.manager.root_token
         self.client.sys.disable_auth_method("userpass")
 
-    def test_write_fix_path(self):
-        self.client.write_fix_path("secret/foo", data={"path": "foo1", "foo": "foo2"})
+    def test_write_data(self):
+        self.client.write_data("secret/foo", data={"path": "foo1", "foo": "foo2"})
         result = self.client.read("secret/foo")
 
         assert result["data"]["path"] == "foo1"
--- a/tests/integration_tests/v1/test_integration.py
+++ b/tests/integration_tests/v1/test_integration.py
@@ -142,6 +142,21 @@ class IntegrationTest(HvacIntegrationTestCase, TestCase):
         self.client.token = self.manager.root_token
         self.client.sys.disable_auth_method("userpass")
 
+    def test_write_fix_path(self):
+        self.client.write_fix_path(
+            "secret/foo",
+            data={
+                "path": "foo1",
+                "foo": "foo2"
+            }
+        )
+        result = self.client.read("secret/foo")
+
+        assert result["data"]["path"] == "foo1"
+        assert result["data"]["foo"] == "foo2"
+        
+        self.client.delete("secret/foo")
+
     def test_missing_token(self):
         client = utils.create_client()
         assert not client.is_authenticated()


briantist:
Hi @M0NsTeRRR welcome! Thanks very much for submitting this.

I like the idea of adding a new function to avoid the issue described in https://github.com/hvac/hvac/issues/133#issuecomment-1612322787 where we would lose access to a second named kwarg without a transition period.

I think I would like to modify the idea and schedule as follows:
* `v1.2.0` -- add the new function (let's call it `write_data`) [**THIS PR**]
* `v2.0.0` -- update `write` as follows:
  * remove the (explicit) arguments `path` and `wrap_ttl`, add `*args`
  * set value of `path` and `wrap_ttl` based on `*args` (for positional) and `**kwargs`
  * if `path` or `wrap_ttl` values were gotten from `**kwargs`, they won't be set as data (this preserves previous behavior), and we'll issue a warning suggesting the use of `write_data` or setting these positionally
  * if `data` exists in `kwargs` we'll also warn about its impending use (see next items) and suggest `write_data`
* `v3.0.0` -- "activate" a `data` parameter in `write`:
  * if `kwargs` contains `data` and no other non-argument keys, then this call will operate the same as `write_data` (forward compatible)
  * if `kwargs` contains both `data` and other data keys, raise an exception
  * otherwise, it will operate the same as `write` does now (backward compatible), with a deprecation warning that `kwargs` is to be removed in `v4.0.0`, suggesting using the `data` dictionary going forward
* `v4.0.0` -- update `write` to its Final Form™:
  * remove `*args` and `**kwargs`, and return to defined arguments for `path`, `wrap_ttl`, and `data`
  * no more warnings
  * at this point, `write` basically becomes `write_data`
* `v4.0.0` -- mark `write_data` as an alias for `write` in documentation
  * deprecate its name in favor of `write`, but because some folks may have migrated to this name, and because it has almost no maintenance burden, let's make it a long deprecation, to be removed in say, `v6` or `v7`? Exact version can be determined later.

---

So, here's my reasoning for the more complex replacement:
* I would like to avoid a situation where people have to migrate `write` calls to `write_data` calls, and then migrate back again in a short time
  * this can be really annoying for projects using this library that want to support a range of versions
* I would like to avoid warnings that don't have a (good) way to make them go away

So the idea is that in `v1.2.0` we introduce the new implementation as a new function name, so it's opt-in. `write` doesn't change at all: if you're hitting a use case where you need to use `write_data`, it's there for you.

In `v2.0.0`, we change the `write` implementation so that we can detect if our parameters were passed positionally or by keyword. If they were set positionally, we do nothing, `write` operates identically, and we avoid warnings.

If the arguments were passed by keyword, `write` still does the same operation that it used to (those values cannot be part of the data written) but a warning is issued to suggest alternatives:
1. Set `path` and/or `wrap_ttl` positionally (you can keep using the `write` function, but make the warning go away) -- this is decent if you know that you'll never need to be setting those names as actual data
2. Use `write_data` if those names might need to be part of data (explicitly, or if you're writing user-/run-time- supplied data) -- eventually down the line you'll have to change back to `write`
(or, the warning can be swallowed/ignored)
In this version, we'll also warn if `kwargs` contains a `data` key since behavior of that will change in the next version: suggest switching to `write_data` to avoid breakage.

In `v3.0.0`, we'll add a way to use a `data` parameter in `write`. This will be an actual breaking change for any call that tried to write a `data` key, but at this point, `write` continues to work the same for anyone who wasn't using a `data` key before
Adding this allows a caller who wanted to stick with the `write` function name to update their implementation to use the `data` dictionary instead, which will let them be forward compatible with `v4.0.0`, or allows a caller who changed to `write_data` to change back now ahead of `write_data`'s eventual removal.

In `v4.0.0`, `write` becomes `write_data`, setting via `kwargs` no longer works, and we can keep the `write_data` name around as an alias for several versions to give a lot of time for folks to migrate back to `write`.

---

I will try to write a "guide" for callers on what change(s) to make, in what version(s), depending on their needs.

---

Sorry that was quite long! What do you think?

hvac/v1/__init__.py
```diff
@@ -265,6 +265,23 @@ def write(self, path, wrap_ttl=None, **kwargs):
         """
         return self._adapter.post(f"/v1/{path}", json=kwargs, wrap_ttl=wrap_ttl)
 
+    def write_fix_path(self, path, data={}, wrap_ttl=None):
```
briantist, line 268:
```suggestion
    def write_data(self, path, wrap_ttl=None, data={}):
```

hvac/v1/__init__.py
```diff
@@ -265,6 +265,23 @@ def write(self, path, wrap_ttl=None, **kwargs):
...
+        """Same as write() method but fix issue #133 (you can pass `path` key in POST data)
```
briantist, line 269:
```suggestion
        """Write data to a path. Similar to write() without restrictions on data keys.
```

I also suggest that we add a description to `write` that suggests the use of `write_data`. I can't make this a real "suggestion" in GitHub because the line wasn't changed (line 255), but here's a draft:

> Write data to a path. Because this method uses kwargs for the data to write, "path" and "wrap_ttl" data keys cannot be used. If these names are needed, or if the key names are not known at design time, consider using the `write_data` method.

hvac/v1/__init__.py
```diff
@@ -265,6 +265,23 @@ def write(self, path, wrap_ttl=None, **kwargs):
...
+
+        Supported methods:
+            POST /<path>
+
+        :param path:
+        :type path:
+        :param data:
+        :type dict:
+        :param wrap_ttl:
+        :type wrap_ttl:
```
briantist, line 279:
```suggestion
        :param wrap_ttl:
        :type wrap_ttl:
        :param data:
        :type dict:
```

tests/integration_tests/v1/test_integration.py
```diff
@@ -142,6 +142,15 @@ def test_create_userpass(self):
         self.client.token = self.manager.root_token
         self.client.sys.disable_auth_method("userpass")
 
+    def test_write_fix_path(self):
+        self.client.write_fix_path("secret/foo", data={"path": "foo1", "foo": "foo2"})
```
briantist, line 146:
```suggestion
    def test_write_data(self):
        self.client.write_data("secret/foo", data={"path": "foo1", "foo": "foo2"})
```

M0NsTeRRR:
Hello,

No worries :)

I find the plan a bit complicated and it adds a lot of maintenance on several versions. If we follow the semantic version, a major version update can break changes. We also have to think that we will have to explain this change in the changelog and that it must remain quite readable.
It seems complex to me to project ourselves so much into the future with the end of this migration to V4 there may be time to see a complete rewrite.
Also the changes to the function itself seem rather minimal to me for an user of the library.

I would suggest : 
- `v1.2.0` -- add the new function (let's call it write_data) **[THIS PR]**
- `v2.0.0` --  delete `write_data` and add the logic in `write`

Regards,

hvac/v1/__init__.py
M0NsTeRRR: After reading it again maybe it's better to put `data` as mandatory like this `write_data(self, path, data, wrap_ttl=None)` because I don't see a case to write at a path without data. What do you think ?

hvac/v1/__init__.py
M0NsTeRRR: Yes I agree with the idea :)

hvac/v1/__init__.py
briantist: There are calls that need a write but all options are either optional or can be specified in the path, so I think having a default like that is valid. For example: https://developer.hashicorp.com/vault/api-docs/system/tools#generate-random-bytes

However, after reading it again, I actually agree with your original ordering 😅 and I would like to make `wrap_ttl` kw-only.
```suggestion
    def write_data(self, path, data={}, *, wrap_ttl=None):
```

I am also considering whether `data` should be kw-only as well 🤔 thoughts?

```suggestion
    def write_data(self, path, *, data={}, wrap_ttl=None):
```

I am thinking about how the calls might look with `data`:
```python
# no data, works either way
client.write_data("sys/tools/random")

# positional
data = {"format": "hex"}
client.write_data("sys/tools/random", {"format": "hex"})
client.write_data("sys/tools/random", dict(format="hex"))
client.write_data("sys/tools/random", data)

# keyword
client.write_data("sys/tools/random", data={"format": "hex"})
client.write_data("sys/tools/random", data=dict(format="hex")
client.write_data("sys/tools/random", data=data)
```

I'm leaning a little toward making everything except `path` be kw-only.

briantist:
> If we follow the semantic version, a major version update can break changes.

While it's true that we can make a breaking change more quickly than I laid out and still conform to semver, strict adherence isn't the only concern. 

* Giving advance notice where possible helps users prepare for changes.
* Using warnings to do so increases the chances of changes being brought to the attention of users.
* Limiting the scope of warnings to actual calls affected helps reduce warning spam and warning fatigue.

In this case, due to the unique situation of us being unable to add the new parameter the original function first due to exacerbating the effect, we have a new function that will eventually replace the original, and this means some users will have to change the function they use twice.

I want to minimize the number of users who have to make that change and make the path as smooth as possible for everyone.

---

> We also have to think that we will have to explain this change in the changelog and that it must remain quite readable.

I agree; but our changelog at the moment only consists of PR titles and links, so any explanation for these changes need to go into some PR or issue.

I will write a more coherent guide for what changes to make and when they need to be made (from the POV of a user and/or a dependent project maintainer) to help with that.

Writing from this POV should be simple than the explanations above which are all about implementation.

> It seems complex to me to project ourselves so much into the future with the end of this migration to V4 there may be time to see a complete rewrite.

I can definitely understand that it seems like v4 is incredibly far in the future. It look a very long time before v1 was released, and since then, we're quite behind schedule on v2.

This is something I'm actively looking to change. I want us releasing more quickly, with fewer changes per release (mostly these will be minor releases).

[But we've also committed to](https://github.com/hvac/hvac/issues/877):
1. Remove support for Python versions in `hvac` when they stop being supported
2. Only remove support for a Python version in a major version of `hvac`

This means we should be releasing a major version a minimum of once a year, and that assumes no other breaking changes outside that schedule. Realistically, I expect we'll do ~2 major releases a year, as long as there is something breaking to release.

I also expect to release `v2.0.0` very shortly after `v1.2.0` (like within a few weeks at most), and if nothing else breaking is introduced except Python EoLs, that puts `v3` around a year from now, and `v4` around 2 years from now, not terribly long, but we can still move up some of the changes, depending on how the warnings and deprecations go, based on feedback, etc.

I can say with some confidence, that even looking 2 years out, the chance of a complete rewrite is effectively zero. If that were to happen, it's unlikely it would be with the current set of maintainers, and it would likely be a new project altogether.

> Also the changes to the function itself seem rather minimal to me for an user of the library.

They are minimal! Especially if your scope is that you have a script or small project, or an application where you can strictly control your versions.

If you maintain a project that also needs to support a range of library versions (such as one of the projects I maintain), then more overlap is appreciated :)

---

> I find the plan a bit complicated and it adds a lot of maintenance on several versions
I think I can also see that the plan sounds complicated when explained all at once.

I think the maintenance needed _per version_ is quite minimal, and can be planned out in advance (which I will write up specific issues for).

I also want to stress that the responsibility for that maintenance is on us maintainers. While [this PR and any future contributions  are very much appreciated](https://github.com/hvac/hvac/discussions/994), I do not expect that this PR means you must also implement all the other steps. 

The warning stuff in particular is probably not very fun to do, but I'm prepared to implement it for `v2.0.0` because I want to get that release out quickly. And I feel the same about the other milestones.

---

We can continue this discussion (and probably we should move it back to #133), but I think we both at least agree that this PR can be added to `v1.2.0` no matter what, so we do not have to finish the decision before moving forward with this. 🙂

Thanks again for your time and this contribution.

M0NsTeRRR:
Ok I understand your point of view, it's understable. One of my concern was of course the release of v1 who tooks 5 years but as you are the author of the projet you have the clearest view of it.

Your welcome :)

hvac/v1/__init__.py
M0NsTeRRR: I prefer the first suggested change, `wrap_ttl` can be an option that is not required and `data` is required but sometimes It can be an empty dict. Also what makes the most sense compared to the rest of the library ?

briantist: When you say first suggested change, which one do you mean?

> Also what makes the most sense compared to the rest of the library?

One thing I have been seeing is that we get a lot of PRs to add new parameters to existing methods. For a lot of our methods, `mount_point` is the last parameter, and the impulse is to keep it last and add new parameters before it, or to rearrange parameters in alphabetical order when they correspond to the API parameters, etc.

The problem is that since we don't enforce keyword use, changing the order of parameters will break any existing uses that passed parameters positionally.

So as a result, I keep having to suggest to PR authors that they move the new parameters to the end of the method definition, even if that doesn't look right.

What I'd like to start doing in `hvac` is that we think of this intentionally when adding new methods, and going forward, we start a transition toward the (breaking) change of making most parameters of most methods kw-only, so that we can solve that going forward.

---

So that's where I'm coming from with this change. 

For this function, `path` makes sense to be the first parameter in this and many other calls, `data` _might_ make sense as a second positional (though I think that it might be better to force it as keyword), and anything else, I think should definitely be forced as a keyword.

briantist: > That can happens only if hashicorp spec is updated. I mean when the signature of the function is created from the spec, if something change on their side it can breaks compatibility but yeah we need to takes it in consideration.

Sometimes this is because we missed parameters, or sometimes it's because new ones were added to the spec. But not all of the parameters in the python function directly correspond to the API spec, `mount_point` is a good example, since it is used as part of path generation.

> After reading again your example, data can be considered as an optional argument because this function is an abstraction of multiple API rest call that sometimes doesn't have DATA so we could consider it as a keyword argument.
`def write_data(self, path, *, data={}, wrap_ttl=None):` could be the best idea.

Ok, let's go with that then! Thank you.

briantist (approved):
This looks great, thank you! I'm going to merge soon, I want to write up the other issues for future changes, and to provide guidance for users before merging. Very excited to get this released!

M0NsTeRRR:
Sure your welcome, I will fix ansible issue once you release a tag. Will you tag a new release after this PR or I have to wait a bit ?

briantist:
> Sure your welcome, I will fix ansible issue once you release a tag. Will you tag a new release after this PR or I have to wait a bit ?

As it happens, this is the last remaining PR that we're [planning for `v1.2.0`](https://github.com/hvac/hvac/milestone/37) so we'll be doing a new release pretty quickly after merging this one.

Let's chat on [the `community.hashi_vault` issue](https://github.com/ansible-collections/community.hashi_vault/issues/389) as well. Unless we make the decision to raise the minimum supported `hvac`, we're going to need to support the earlier version of `hvac` for a bit on the Ansible collection, and so we need a bit of code to determine which method to use, and tests around that. But I'm happy to help with some of the implementation there too!

PR #1039, opened 2023-09-02
Add alias_name_source for Kubernetes Auth create_role

michael-diggin:
For the Kubernetes auth method, Vault's create role API endpoint accepts `alias_name_source` as part of the payload ([https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#alias_name_source](https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#alias_name_source)). This PR adds that as an optional argument to the `create_role` python method. Defaults to not set so no change to existing behaviour (as this param was only added in Vault 1.9)

### DIFF
--- a/hvac/api/auth_methods/kubernetes.py
+++ b/hvac/api/auth_methods/kubernetes.py
@@ -119,7 +119,7 @@ class Kubernetes(VaultApiBase):
         policies=None,
         token_type="",
         mount_point=DEFAULT_MOUNT_POINT,
-        alias_name_source="serviceaccount_uid",
+        alias_name_source="",
     ):
         """Create a role in the method.
 
@@ -179,8 +179,10 @@ class Kubernetes(VaultApiBase):
             "bound_service_account_namespaces": comma_delimited_to_list(
                 bound_service_account_namespaces
             ),
-            "alias_name_source": alias_name_source,
         }
+        if alias_name_source:
+            params["alias_name_source"] = alias_name_source
+
         params.update(
             utils.remove_nones(
                 {
--- a/hvac/api/auth_methods/kubernetes.py
+++ b/hvac/api/auth_methods/kubernetes.py
@@ -180,7 +180,7 @@ class Kubernetes(VaultApiBase):
                 bound_service_account_namespaces
             ),
         }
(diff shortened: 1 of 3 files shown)


hvac/api/auth_methods/kubernetes.py
```diff
@@ -119,6 +119,7 @@ def create_role(
         policies=None,
         token_type="",
         mount_point=DEFAULT_MOUNT_POINT,
+        alias_name_source="",
```
briantist, line 122:
```suggestion
        alias_name_source=None,
```

hvac/api/auth_methods/kubernetes.py
```diff
@@ -176,6 +180,9 @@ def create_role(
                 bound_service_account_namespaces
             ),
         }
+        if alias_name_source:
```
briantist, line 183:
```suggestion
        if alias_name_source is not None:
```

briantist:
Hi @michael-diggin welcome and thanks for putting this up! 

I made two small suggestions but it otherwise looks good. I'm trying to get `v1.2.0` out imminently and we can probably get this included. If you're ok with the suggestions, you can commit them directly from the GitHub UI, or make them locally, or let me know and I can commit the changes as well.

michael-diggin:
Hi @briantist, thanks for the quick review! I've accepted both suggestions and it would be great to get this into v1.2.0!

tests/integration_tests/api/auth_methods/test_kubernetes.py
```diff
@@ -171,6 +178,7 @@ def test_create_role(
         bound_service_account_names=None,
         bound_service_account_namespaces=None,
         token_type=None,
+        alias_name_source="",
```
briantist, line 181:
```suggestion
        alias_name_source=None,
```

briantist (commented):
I missed one, will commit it for expediency

briantist (approved):
@michael-diggin looks great! thanks very much for adding tests as well. Hope you might stick around with more contributions if you're able, [we're actively seeking help on the project](https://github.com/hvac/hvac/discussions/994). Cheers!

PR #1035, opened 2023-08-27
Release/v1.2.0

briantist:
Release PR for `v1.2.0`

PR #1041, opened 2023-09-02
use anonymous references in changelog conversion

briantist:
We use m2r primarily to convert the markdown changelog to RST, so we don't need named references. Since we may have multiple changelog entries refer to the same GitHub issue, and we use the same text to anchor it (GH-###), it would result in duplicate explicit target names.
- https://github.com/CrossNox/m2r2/issues/30

For an example, see the error from #1035 :
- https://github.com/hvac/hvac/actions/runs/5992125579/job/16251211029?pr=1035#step:8:75

I tested locally to ensure that this fixes the issue.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,6 +33,13 @@ extensions = [
     "autodocsumm",
 ]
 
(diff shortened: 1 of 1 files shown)


kaypeter87 (approved):

briantist:
thanks for the review @kaypeter87 !

PR #1045, opened 2023-09-06
Fix raise_on_deleted_version warning

### LINKED ISSUE #955: The default value of `raise_on_deleted_version` will change from `True` to `False` in `v3.0.0`
kv2's `read_secret_version` and `read_secret` methods will have their new `raise_on_deleted_version` parameter (#907) change from (effectively) `True` to `False` in `v3.0.0`.

This is unlikely to affect most uses, but is a potentially breaking change in behavior.

When the value is `True`, attempting to retrieve a (soft) deleted secret version, including the "latest" version of a secret if that version happens to be deleted, will result in an `InvalidPath` exception. This the behavior that `hvac` had historically.

When the value is `False`, the actual result from Vault will be returned without raising an exception; that result will include no data, but have metadata related to the secret including when it was deleted.

It is also possible as of [`v1.1.0`](https://github.com/hvac/hvac/milestone/33) to get the raw JSON response from Vault on raised exceptions, so you may also choose to keep the exception-raising functionality and check the result in your own code.

**Only the default value is changing in `v3.0.0`.** 

By changing your code now to explicitly choose which functionality you want, you can avoid any changes in behavior when the default value is changed.

For more implementation details, see:
- #907

iTrooz:
The warning mentioned the "raise_on_deleted" parameter, where the parameter is actually "raise_on_deleted_version"

Fixes an issue raised in [#955](https://github.com/hvac/hvac/issues/955#issuecomment-1708213606)

### DIFF
--- a/hvac/api/secrets_engines/kv_v2.py
+++ b/hvac/api/secrets_engines/kv_v2.py
@@ -131,7 +131,7 @@ class KvV2(VaultApiBase):
 
         if raise_on_deleted_version is None:
             msg = (
-                "The raise_on_deleted parameter will change its default value to False in hvac v3.0.0. "
+                "The raise_on_deleted_version parameter will change its default value to False in hvac v3.0.0. "
                 "The current default of True will presere previous behavior. "
                 "To use the old behavior with no warning, explicitly set this value to True. "
                 "See https://github.com/hvac/hvac/pull/907"


briantist (approved):

briantist:
@iTrooz thanks for noticing this and putting in the fix, [this is live in `v1.2.1`](https://github.com/hvac/hvac/releases/tag/v1.2.1). 
If you're, we are [actively looking for more contributions](https://github.com/hvac/hvac/discussions/994)!

PR #1048, opened 2023-09-09
Drop Python 3.6 & 3.7, Add Python 3.11, bump some dependencies

### LINKED ISSUE #999: Add (official) support for Python 3.11
I don't know of any specific blockers to using Python 3.11 right now, but we don't test against it, and the package doesn't explicitly state support for it.

We should add testing against it ASAP, and if there are any failures fix them. Once we're confident in it we can add it to the package specifications in the next minor release.

briantist:
This PR will make a number of changes around python, dependencies, & CI/CD

Resolves #999
Supersedes #1005 

- remove support for python 3.6 & 3.7
- add support for python 3.11
- (failed) attempt to add python 3.12
- bump `poetry` version to `1.6.1`
- update `pyproject.toml` dev dependencies to newer poetry group format
  - with associated CI changes
  - disable all caching on build jobs since they don't install dev dependencies and can poison the cache for other jobs
  - TODO: re-do CI caching and setup
  - TODO: separate out docs dependencies into their own group
- bump versions of various dev dependencies to work on newer python and with each other
  - replace deprecated flask syntax after updating flask
  - add `jinja2` as an explicit dev dependency so we can pin it to deal with old docs dependencies
  - TODO: update docs dependencies to much newer versions
- bump `actions/checkout` to `v4` and disable showing progress
- TODO: improve CI to better surface warnings in tests so we can proactively address them
- TODO: test warnings

(all TODOs are for new issues/PRs)

### About Python 3.12

I figured I'd add it to the matrix if we could (pre-release) just to be prepared. Current blocker (haven't explored all of them) is the `greenlet` package which is a dependency of `sqlalchemy` (via `Flask-SQLAlchemy`) used in our tests.

It seems `greenlet` has fixed those 3.12 issues but they've only been released in pre-release versions so far, so when 3.0.0 is out we should be able to revisit this, as long as the packages we actually use update their dependencies too.
- https://github.com/python-greenlet/greenlet/tags
- https://github.com/python-greenlet/greenlet/issues/365


kaypeter87 (approved):
LGTM!

briantist:
@kaypeter87 thanks for reviewing!

PR #1049, opened 2023-09-09
Make `plaintext` and `ciphertext` optional for batch operations

dosisod:
The Vault API requires setting `plaintext` and `ciphertext` for the encrypt/decrypt data API, even though the fields are ignored when `batch_input` is set. This means that end users have to include an empty argument when doing batch operations, which is a bit annoying.

While this works, it might be more ergonomic to have an `encrypt_data` and a `batch_encrypt_data` function as most of the `encrypt_data` args are ignored in batch mode anyways, which could potentially lead to bugs. This is a bit out of scope for this PR, just thought I would mention it here as an alternative.

Also, I could not test locally due to self-signed certificate errors. Is there a certificate I need to install for testing?

### DIFF
--- a/hvac/api/secrets_engines/transit.py
+++ b/hvac/api/secrets_engines/transit.py
@@ -314,7 +314,7 @@ class Transit(VaultApiBase):
     def encrypt_data(
         self,
         name,
-        plaintext,
+        plaintext=None,
         context=None,
         key_version=None,
         nonce=None,
@@ -335,7 +335,7 @@ class Transit(VaultApiBase):
 
         :param name: Specifies the name of the encryption key to encrypt against. This is specified as part of the URL.
         :type name: str | unicode
-        :param plaintext: Specifies base64 encoded plaintext to be encoded.
+        :param plaintext: Specifies base64 encoded plaintext to be encoded. Ignored if ``batch_input`` is set, otherwise required.
         :type plaintext: str | unicode
         :param context: Specifies the base64 encoded context for key derivation. This is required if key derivation is
             enabled for this key.
@@ -367,6 +367,8 @@ class Transit(VaultApiBase):
         :return: The JSON response of the request.
         :rtype: dict
         """
+        if plaintext is None and batch_input is None:
+            raise ValueError("plaintext must be specified unless batch_input is set")
         params = {
             "plaintext": plaintext,
         }
@@ -395,7 +397,7 @@ class Transit(VaultApiBase):
     def decrypt_data(
         self,
         name,
-        ciphertext,
+        ciphertext=None,
         context=None,
         nonce=None,
         batch_input=None,
@@ -408,7 +410,7 @@ class Transit(VaultApiBase):
 
         :param name: Specifies the name of the encryption key to decrypt against. This is specified as part of the URL.
         :type name: str | unicode
-        :param ciphertext: the ciphertext to decrypt.
+        :param ciphertext: The ciphertext to decrypt. Ignored if ``batch_input`` is set, otherwise required.
         :type ciphertext: str | unicode
         :param context: Specifies the base64 encoded context for key derivation. This is required if key derivation is
             enabled.
@@ -426,6 +428,8 @@ class Transit(VaultApiBase):
         :return: The JSON response of the request.
         :rtype: dict
         """
+        if ciphertext is None and batch_input is None:
+            raise ValueError("ciphertext must be specified unless batch_input is set")
         params = {
             "ciphertext": ciphertext,
         }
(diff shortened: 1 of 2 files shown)


briantist:
Hi @dosisod welcome! I'll look over this in some more detail soon. I will probably ask you to rebase soon (or I'll push up a rebase on your branch) as we're about to change some python versions and dependencies #1048 .

> Also, I could not test locally due to self-signed certificate errors. Is there a certificate I need to install for testing?

Could you show me:
* what you ran to set up your local environment
* the command you're running to execute tests
* the error output you got?

We have a [contributing guide](https://hvac.readthedocs.io/en/main/contributing.html) but I suspect it needs updating and it's helpful to know what kind of issues people run into.

dosisod:
Thanks for the quick response!

Here are the commands I ran to set everything up:

```
$ git clone https://github.com/hvac/hvac.git
$ cd hvac
$ poetry install
$ poetry shell
$ make test
pytest --cov=hvac tests/
make: *** [Makefile:7: test] Error 1

$ pytest tests -x
<snip>
                except Exception as ex:
                    print(ex)
                    if process.poll() is not None:
>                       raise Exception("Vault server terminated before becoming ready")
E                       Exception: Vault server terminated before becoming ready

tests/utils/server_manager.py:74: Exception
---------------------------------------------------------------------------------------------- Captured stdout setup -----------------------------------------------------------------------------------------------
HTTPSConnectionPool(host='localhost', port=8200): Max retries exceeded with url: /v1/sys/init (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1006)')))
HTTPSConnectionPool(host='localhost', port=8200): Max retries exceeded with url: /v1/sys/init (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1006)')))
================================================================================================= warnings summary =================================================================================================
tests/utils/__init__.py:10
  /home/loot/git/hvac/tests/utils/__init__.py:10: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
    from distutils.spawn import find_executable

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

I had to add a `print()` statement to actually see the exception that was causing it to fail.

I looked at the CI workflows, but those don't do any certificate magic or anything, so I'm at a loss as to why it doesn't work on my machine. I'm using HVAC with mTLS in one of my projects with no issues, so I suspect the certificates aren't being loaded during testing. Currently using Python 3.11.5 on Arch Linux, if that helps.

hvac/api/secrets_engines/transit.py
```diff
@@ -335,7 +335,7 @@ def encrypt_data(
 
         :param name: Specifies the name of the encryption key to encrypt against. This is specified as part of the URL.
         :type name: str | unicode
-        :param plaintext: Specifies base64 encoded plaintext to be encoded.
+        :param plaintext: Specifies base64 encoded plaintext to be encoded. Not needed if batch_input is set.
```
briantist, line 338:
```suggestion
        :param plaintext: Specifies base64 encoded plaintext to be encoded. Ignored if ``batch_input`` is set, otherwise required.
```

hvac/api/secrets_engines/transit.py
```diff
@@ -367,6 +367,8 @@ def encrypt_data(
         :return: The JSON response of the request.
         :rtype: dict
         """
+        if not plaintext and not batch_input:
```
briantist, line 370:
```suggestion
        if plaintext is None and batch_input is None:
```

hvac/api/secrets_engines/transit.py
```diff
@@ -408,7 +410,7 @@ def decrypt_data(
 
         :param name: Specifies the name of the encryption key to decrypt against. This is specified as part of the URL.
         :type name: str | unicode
-        :param ciphertext: the ciphertext to decrypt.
+        :param ciphertext: The ciphertext to decrypt. Not needed if batch_input is set.
```
briantist, line 413:
```suggestion
        :param ciphertext: The ciphertext to decrypt. Ignored if ``batch_input`` is set, otherwise required.
```

hvac/api/secrets_engines/transit.py
```diff
@@ -426,6 +428,8 @@ def decrypt_data(
         :return: The JSON response of the request.
         :rtype: dict
         """
+        if not ciphertext and not batch_input:
```
briantist, line 431:
```suggestion
        if ciphertext is None and batch_input is None:
```

briantist (commented):
Thanks again for putting up this PR and especially for adding tests. Could you rebase this to include the latest changes in `main`?

briantist (approved):
Looks great, thanks again @dosisod ! We're [actively looking for contributors](https://github.com/hvac/hvac/discussions/994) so if you're interested please don't be a stranger :)

PR #1051, opened 2023-09-10
dev dependencies: bump pytest version

briantist:
Resolves [CVE-2022-42969](https://github.com/advisories/GHSA-w596-4wvx-j9j6)

Worth noting that this CVE doesn't really affect us, since it only affects SVN, and `pytest` doesn't use the affected code path in `py`, but bumping our version of `pytest` will get rid of the entire transitive dependency which is nice.

Removes:
- `py`
- `attrs`


kaypeter87 (approved):

briantist:
Thanks for reviewing @kaypeter87 !

Issue #984, opened 2023-04-29
Add support for Login MFA


briantist:
The newer MFA style supported in Vault since v1.10.0 requires changes to use with `hvac` and we don't support it right now.

It supports both a one-phase and (more commonly) two-phase approach, the latter of which will require multiple requests to be made to complete a login.

Adding support for _configuring_ Login MFA should be pretty straightforward, with a new class.

However since Login MFA can be used with any auth method, support for using it on login might be a little more involved. We will probbaly want to solve this "centrally" somehow, I'd guess in the `login` method of the adapter(s), and then adding any parameters that need to be passed through to the individual auth method `login` methods.

### Links

* https://developer.hashicorp.com/vault/docs/auth/login-mfa
* https://developer.hashicorp.com/vault/api-docs/secret/identity/mfa
* https://github.com/hashicorp/vault/blob/main/CHANGELOG.md#1100
* https://developer.hashicorp.com/vault/docs/auth/login-mfa/faq

cognifloyd:
Before adding Login MFA support to the various auth login methods, we should add helper methods for the mfa validate endpoint:
- `/sys/mfa/validate`: https://developer.hashicorp.com/vault/api-docs/system/mfa/validate

To use Login MFA with okta auth today, I'm using this:
```python
def handle_vault_login_mfa(vault: hvac.Client, login_response: dict[str, Any]) -> None:
    mfa_requirement = login_response.get("auth", {}).get("mfa_requirement", {})
    if not mfa_requirement or not mfa_requirement.get("mfa_request_id"):
        return

    # We don't care what the constraint is named, so just get the first value
    # (the vault CLI only does interactive MFA only if there is one constraint).
    # Then the mfa method is tucked under an "any" object (who knows why),
    # which is a list that contains one item.
    mfa_method = list(mfa_requirement["mfa_constraints"].values())[0]["any"][0]
    if not mfa_method["use_passcode"]:
        return

    totp = getpass(prompt=f"Please enter {mfa_method['type']} TOTP (will be hideen): ")

    mfa_response = vault.write(
        "sys/mfa/validate",
        mfa_request_id=mfa_requirement["mfa_request_id"],
        mfa_payload={
            mfa_method["id"]: [totp],
        },
    )
    vault.token = mfa_response["auth"]["client_token"]

resp = client.auth.okta.login(
    username=email,
    password=password,
)
handle_vault_login_mfa(client, resp)
```

[...]

PR #1056, opened 2023-09-10
Remove MFA class

### LINKED ISSUE #1026: 2.0.0 - Drop old MFA methods in favor of LegacyMFA
In https://github.com/hvac/hvac/pull/883 we announced that the old methods for legacy MFA (which was dropped in Vault 1.11) would be dropped from the MFA class and remain available in a new `LegacyMFA` class (introduced in that PR).

This issue is the reminder to drop those.

briantist:
Resolves #1026 

As previously announced, the deprecated `MFA` class is being removed. All of its methods were moved to the `LegacyMFA` class.

The `MFA` class may resurface depending on `hvac`'s implementation of Vault's newer "Login MFA":
- #984 

In addition to removing the class, this PR is cleaning up some other things:
- moving the unit tests
- fixing documentation
- fixing references in the client `auth` property
- removing the very deprecated direct `client.mfa` property
- adding tests for `utils` module functions, whose code paths are no longer hit because they aren't currently in use


kaypeter87 (approved):

briantist:
Thanks again for reviewing @kaypeter87 !

PR #1057, opened 2023-09-11
Fix typos, add `typos` linter to CI

dosisod:
[`typos`](https://github.com/crate-ci/typos) is a great tool for finding typos in a varietry of file formats.

I have not tested these changes, though most changes are in the documentation and the docstrings. There are 2 potential API changes: The `saftey_buffer` arguments for `tidy_blacklist_tags` and `tidy_identity_whitelist_entries` have been changed to `safety_buffer`. If we want to maintain compatibility we could use kwargs and check for the new flag, then the old one.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -62,6 +62,9 @@ jobs:
     - name: Lint with `flake8`
       run: poetry run flake8 . --count --statistics
 
+    - name: Check for typos
+      run: poetry run typos --color always
+
   docs-tests:
     name: Documentation Tests
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,7 @@ This is the last expected release before `v2.0.0`.
 ### 🐛 Bug Fixes
 
 - Allow for reading deleted secret versions (kv2) without an exception ([GH-907](https://github.com/hvac/hvac/pull/907))
-- fix vault client certificates loaded from envirnoment variables ([GH-943](https://github.com/hvac/hvac/pull/943))
+- fix vault client certificates loaded from environment variables ([GH-943](https://github.com/hvac/hvac/pull/943))
 - approle - fix metadata for generated secret IDs, re-add `wrap_ttl` ([GH-782](https://github.com/hvac/hvac/pull/782))
 - AWS secret engine - fix `generate_credentials` for STS endpoint ([GH-934](https://github.com/hvac/hvac/pull/934))
 - Propagate client's adapter to API categories ([GH-939](https://github.com/hvac/hvac/pull/939))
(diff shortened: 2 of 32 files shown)


briantist:
Thanks for this @dosisod ! It's uh.. really something to see all the typos here 😅 I'd like to treat this PR a little differently than most.

First, the one functional change you mentioned, re: `saftey_buffer`, thanks for specifically raising that. I want to update that parameter specifically in a new PR, and make the change with an aliased parameter and deprecation of the old. Luckily, we have a pattern for doing just that:
* https://github.com/hvac/hvac/blob/main/hvac/utils.py#L52
* https://github.com/hvac/hvac/blob/main/hvac/api/auth_methods/ldap.py#L15
* https://github.com/hvac/hvac/pull/975/files

Because this will then eventually be a breaking change (in `v3.0.0` when the old one is removed) I want to get this out in `v2.0.0` so folks have time to make updates.

---

The other thing I want to do then is 1) add the `typos` package/workflow in CI, and make the other typo changes themselves.

I want all the typo changes to be in a single commit, so that we can put that commit into `.git-blame-ignore-revs` for helping blame views: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

We need to do this via at least 2 PRs.

The first PR (this one, probably) can both set up the typo workflow, and fix the typos, each in their respective commits, and I will "Rebase and merge" instead of squash and merge (so we will have to rebase/squash locally to get the 2 commits we want). Alternatively, we could do these steps as separate PRs but I guess we'd have to fix all the typos first so the workflow doesn't fail on them.

The last (second or third) PR would be to add the ignore file with the commit hash containing the typo fixes.

---

I consider the new PR for `saftey_buffer` to be high priority and want to get that done soon (if you intend to submit that PR, which is very appreciated, please let me know, if not I will get to it at some point soon), the other steps here... I may have to delay those in favor of doing other things to get `v2.0.0` out. I only mention that in advance because I don't want to seem like I'm ignoring them if you or someone else puts up those changes. I will definitely get to them (and your PRs so far are pretty complete and easy to review so there's that too).

dosisod:
@briantist sorry for the delay on this! I just opened a PR for the `safety_buffer` argument. Once that gets merged I'll rebase and remove the `safety_buffer` changes (as well as the MFA changes), then make the 2 commits for the CI and typo changes.

briantist:
No worries at all, I've been short on time myself

dosisod:
@briantist Ok I rebased and made it so there are 2 commits: The first sets up `typos` and the related workflows/Makefile, and the second commit updates all the typos.

Also, do we need to make a separate PR to update the `.git-blame-ignore-revs` files? Assuming you do a normal merge and not a rebase, the git SHA should be the same. If you rebase though the SHA's will be different though, hence the need for another PR. I'm fine with it either way, just thought I'd ask.

briantist:
> @briantist Ok I rebased and made it so there are 2 commits: The first sets up `typos` and the related workflows/Makefile, and the second commit updates all the typos.

Thanks, I've merged some other things and there are conflicts now, but I'm going to pull this down today and try to resolve all that so we can get this in.

> Also, do we need to make a separate PR to update the `.git-blame-ignore-revs` files? Assuming you do a normal merge and not a rebase, the git SHA should be the same. If you rebase though the SHA's will be different though, hence the need for another PR. I'm fine with it either way, just thought I'd ask.

Regular merge is disabled in the repo and I think we want to keep it that way. From what I could find, GitHub (through the UI) does not support any  method that preserves commit hashes, though it is possible if you merge from the CLI and push instead which I don't really want to do.

I'll take care of adding the ignore revs file once we're ready, there might be at least one other commit I want to add to it.

briantist (approved):
Great job with this, thank you!

PR #874, opened 2022-09-12
Improvement: error_msg changed to f-strings

xunzhou:
As @colin-pm suggested in https://github.com/hvac/hvac/pull/873#discussion_r968735250,
updating `error_msg` to f-string

PR #1059, opened 2023-09-15
Add linting via `ruff`

dosisod:
[Ruff](https://github.com/astral-sh/ruff) is a really great flake8 alternative that includes a very comprehensive set of linter rules. This PR doesn't remove flake8, but since Ruff is meant to be a replacement of flake8 (and implements a superset of the same rules, [click here](https://beta.ruff.rs/docs/rules/) for a full list), we could drop flake8 entirely.

I've configured Ruff to enable all available linter rules, and manually disabled all the ones that don't apply or are too noisy. Since Ruff is actively under development, any new lint rules are likely to be overlooked. By enabling them all (and fixing any errors), it is much easier to keep the codebase up to date.

Each commit is broken down to a single lint rule, or a group of similar lint rules. Feel free to yay or nay the commits you do or don't like, I can easily rebase accordingly.

Probably closes #874 as this PR converts `format()` calls to fstrings.


briantist:
Hello again @dosisod ! First I want to say thanks for the detailed explanation, and especially for splitting the changes per-rule into discrete commits, that's exactly how I would've liked to see that 🤩 

Some of the changes it made look really nice and are definitely ones I agree with, some of the others, I'm a little less sure about or don't understand.

I am going to leave this open but I will have to deprioritize this change a little bit until I can give it more thought (especially per-rule) and try to discuss with the other maintainers. I definitely intend to re-evaluate it, but likely at some point post-`v2.0.0`, hope that's ok!

PR #1060, opened 2023-09-15
Make `pyhcl` optional again

### LINKED ISSUE #1038: Hvac cannot be installed off-architecture
Hvac is pure python and can be successfully used on any processor architecture. However, trying to install `hvac` off-architecture fails because it requires a dependency that is not correctly tagged.

Pip supports various flags that cause it to download and install versions of modules that are different from the architecture from the host running Pip.

This is useful for building programs in situations where the build system architecture does not match where the program will run. A real world example of this is AWS Lambda. I could be building my program on a mac and need to install the linux x86 or linux arm64 version of a package so that it is able to run on AWS Lambda.

If `hvac` is included in my requirements and I try to install them using this approach:

```
pip3 install --platform manylinux2014_x86_64 --python-version 3.9 --only-binary=:all: --target ./outdir hvac==1.1.1
```

The install fails with:

```
Collecting hvac==1.1.1
  Using cached hvac-1.1.1-py3-none-any.whl (144 kB)
ERROR: Could not find a version that satisfies the requirement pyhcl<0.5.0,>=0.4.4 (from hvac) (from versions: none)
ERROR: No matching distribution found for pyhcl<0.5.0,>=0.4.4
```

...because `hvac` depends on this improperly tagged version of `pyhcl`.

I have opened virtuald/pyhcl#87 to fix the tagging on that module, but `hvac` will need to update its dependency to the updated version, when it comes around.

Another option would be to make the dependency on `pyhcl` optional, as it is not needed for most of `hvac`'s functions.

dosisod:
Resolves #1038 
See comment https://github.com/hvac/hvac/issues/1038#issuecomment-1703844682 .

This makes the `pyhcl` dependency installable via `pip install hvac[parser]` again like the README suggests.

The `pyhcl` dependency is also specified as a dev dependency so that it is always installed during testing.

Also, sorry for the PR spam, I've been in contribution mode the past few days!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,8 +34,11 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.8"
-pyhcl = "^0.4.4"
+pyhcl = { version = "^0.4.4", optional = true }
 requests = "^2.27.1"
+
+[tool.poetry.extras]
+parser = ["pyhcl"]
 
 [tool.poetry.group.dev]
 optional = true
@@ -52,6 +55,7 @@ ipaddress = "^1.0.23"
 mock = "^4.0.3"
 nose = "^1.3.7"
 parameterized = "^0.8.1"
+pyhcl = "^0.4.4"
 pytest = "7.0.1"
 pytest-cov = "^3.0.0"
 python-ldap-test = "^0.3.1"


briantist (approved):
Excellent! This was on my short list of things to do for `v2.0.0`, thanks for putting this up!

briantist:
> Also, sorry for the PR spam, I've been in contribution mode the past few days!

haha please do not apologize for that, love to see it, I would usually have tried to look at them sooner, but it's been one of those weeks...

PR #1063, opened 2023-09-17
Changing default values for `sys.initialize` parameters `secret_shares` and `secret_threshold`

### LINKED ISSUE #1030: Vault Initialization fails when using azurekeyvault for auto unsealing
I have a vault server setup on HA using raft storage and azure key vault for auto unsealing. I am trying to automate the process of configuring vault using HVAC

I am trying to initialize vault by using 
```
client.sys.initialize(
                    stored_shares=5, recovery_shares=5,
                    recovery_threshold=3
                    )
```

but i am getting the below error 
`parameters secret_shares,secret_threshold not applicable to seal type azurekeyvault, on put https://<HOST>/v1/sys/init` 

According to the docs when using KMS to unseal vault, recovery_keys should be passed when making the init request instead of secret_shares and secret_threshold 

https://github.com/hashicorp/vault/issues/17551

But in the code i see secret_shares, secret_threshold is added regardless and is always sent with default values when initializing vault 
```
        params = {
            "secret_shares": secret_shares,
            "secret_threshold": secret_threshold,
            "root_token_pgp_key": root_token_pgp_key,
        }
```

With this logic vault init will not work when using KMS for auto unseal or am i missing something

briantist:
Fixes #1030 

For more details see:
- #1030

Because we set default values for `secret_shares` and `secret_threshold` in the `initialize` call, it's not possible to make the API request without values set.

Unfortunately, there are certain cases in the API call where these values must not be set, when using certain auto-unseal configurations, in which only `recovery_shares` and `recovery_threshold` should be set, and Vault will reject the call if the other two are set as well.

# Breaking change

We're changing the default value of these parameters to `None` along with the rest.

In `v2.0.0`, if `secret_shares` is `None`, and `recovery_shares` is also `None`, then we will set `secret_shares` to `5` (which was the default previously), and issue a warning.
Likewise with `secret_threshold` and `recovery_threshold`, where `secret_threshold` will be set to `3` (the previous default) when both of those parameters are `None`.

In `v3.0.0` we will remove the default fallback.

**This is very unlikely to actually be a breaking change in `v2.0.0`** but it could be _if_ all of the following are true:
* there are any versions of the API call where setting both sets of parameters is actually valid (unconfirmed if this is even possible)
* you are making such a call
* you are only setting `recovery_shares` and/or `recovery_threshold` while relying on the defaults for `secret_shares` and/or `secret_threshold`

## Action needed
If you currently call the `initialize` method relying on the previous default values, you should update to using explicit values right away. This can be done right away, even in already released versions.

With `hvac>=2,<3`, relying on defaults will work (with above caveats) but will issue a deprecation warning.

In `hvac>=3` default values will no longer be provided and your calls will probably break unless you specify your values explicitly.

### DIFF
--- a/hvac/api/system_backend/init.py
+++ b/hvac/api/system_backend/init.py
@@ -1,3 +1,4 @@
+import warnings
 from hvac.api.system_backend.system_backend_mixin import SystemBackendMixin
 from hvac.exceptions import ParamValidationError
 
@@ -28,8 +29,8 @@ class Init(SystemBackendMixin):
 
     def initialize(
         self,
-        secret_shares=5,
-        secret_threshold=3,
+        secret_shares=None,
+        secret_threshold=None,
         pgp_keys=None,
         root_token_pgp_key=None,
         stored_shares=None,
@@ -49,7 +50,7 @@ class Init(SystemBackendMixin):
         :type secret_shares: int
         :param secret_threshold: Specifies the number of shares required to reconstruct the master key. This must be
             less than or equal secret_shares. If using Vault HSM with auto-unsealing, this value must be the same as
-            secret_shares.
+            secret_shares, or ommitted, depending on the version of Vault and the seal type.
         :type secret_threshold: int
         :param pgp_keys: List of PGP public keys used to encrypt the output unseal keys.
             Ordering is preserved. The keys must be base64-encoded from their original binary representation.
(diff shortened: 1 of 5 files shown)


kaypeter87 (approved):
I think this is a good compromise. Plus deprecation warnings have been implemented 👍 LGTM

hvac/api/system_backend/init.py
```diff
@@ -96,18 +126,18 @@ def initialize(
...
-        if recovery_pgp_keys is not None:
-            if len(recovery_pgp_keys) != recovery_shares:
-                raise ParamValidationError(
-                    "length of recovery_pgp_keys list argument must equal recovery_shares value"
-                )
-            params["recovery_pgp_keys"] = recovery_pgp_keys
+            if recovery_threshold is not None:
+                if recovery_threshold > recovery_shares:
+                    error_msg = "value for recovery_threshold argument must be less than or equal to recovery_shares argument"
+                    raise ParamValidationError(error_msg)
```
Tylerlhess, line 132: Why use an explicit error_msg variable here when the current file has previously not done so? Its not a breaking instance of code just stylistic.

Tylerlhess (approved):
Code looks good from my side just a single stylistic inconsistency.

hvac/api/system_backend/init.py
briantist: looking at left side in the side-by-side diff, it seems it's because the previous code did in fact use this variable, and I just changed the message minimally but didn't remove use of the var. I don't feel strongly about it, if you want to put in a suggested change in the comment that removes the variable I'll probably accept it.

briantist:
been a little while since CI ran so I'm going to close/re-open to kick off CI again

PR #1074, opened 2023-10-14
Drop support for Vault 1.6-1.10

briantist:
# Breaking changes
- Drop unsupported Vault versions 1.6 through 1.10 in CI (https://endoflife.date/hashicorp-vault)
  - 1.6 and 1.7 remain tested as enterprise versions, mostly because we haven't been able to properly test enterprise versions without a license past those
  - unsupported versions still remain, we will probably become more aggressive about dropping them in the future
  - while dropping these from CI doesn't mean that `hvac` won't work against these versions, it does mean we don't test against them and don't enforce new or edited code to be compatible with them

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -184,15 +184,11 @@ jobs:
         vault-version:
           - "vault-enterprise=1.6.*+ent"
           - "vault-enterprise=1.7.*+ent"
-          - "vault=1.6.*"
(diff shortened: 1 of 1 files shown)


briantist:
It seems there's some change in the behavior of kubernetes support and a test is not getting the result that is expected. I'm going to leave this in draft and it pull it out of the running for 2.0.0 as I don't have time to investigate it further now.

kaypeter87 (approved):

Issue #1034, opened 2023-08-27
Breaking changes coming to `Client.write` method


briantist:
# Summary

[The `Client.write` method](https://hvac.readthedocs.io/en/stable/source/hvac_v1.html#hvac.v1.Client.write) is a generic function for "writing" (`POST`ing) data to Vault. It does not correspond to any specific backend and takes a raw path. This method is useful for being able to perform writes against backends that the library doesn't (yet) have direct support for.

The method signature uses `kwargs` for specifying the data to write. This is problematic because it prevents the use of any of the method's other argument names as keys in the data. See also:
- #133 

# What's changing

The end goal is to change the method so that it accepts a `data` argument which contains the data to write, and to deprecate and remove the ability to set keys via `kwargs`. This change is not only breaking, there is no way to support both on the same method without sacrificing some backwards compatibility, because adding any arguments to the method also makes them unusable in `kwargs`.

## The plan

We will first introduce a new (temporary!) method, `Client.write_data` which can be used as an alternative to `Client.write` and accepts only the `data` argument for setting data.

Over the course of several major versions, we will begin to make targeted breaking changes to the `write` method to bring it closer the `write_data` implementation, with special care taken to warn on usages that could exhibit breakage and to try to avoid breakage on uses of the old method that can be used successfully without changing calls to the temporary method.

Eventually, `write` will look exactly as `write_data` does today, and `write_data` will be removed.

## The schedule

This is the current proposed schedule for what changes will happen and when. The schedule is subject to change based on circumstances at the time but this is the plan as it stands today.

* [`v1.2.0`](https://github.com/hvac/hvac/milestone/37) -- The `Client.write_data` method is added
  * #1028 
* [`v2.0.0`](https://github.com/hvac/hvac/milestone/34) -- update `write` as follows:
  * remove the (explicit) arguments `path` and `wrap_ttl`, add `*args`
  * set value of `path` and `wrap_ttl` based on `*args` (for positional) and `**kwargs`
  * if `path` or `wrap_ttl` values were gotten from `**kwargs`, they won't be set as data (this preserves previous behavior), and we'll issue a warning suggesting the use of `write_data` or setting these positionally
  * if `data` exists in `kwargs` we'll also warn about its impending use (see next items) and suggest `write_data`
* [`v3.0.0`](https://github.com/hvac/hvac/milestone/36) -- "activate" a `data` parameter in `write`:
[...]

PR #1089, opened 2023-10-15
`Client.write` method breaking changes 2.0.0

### LINKED ISSUE #1079: 2.0.0 - implement breaking changes for `Client.write` method
See:
- https://github.com/hvac/hvac/issues/1034

briantist:
Implements:
- #1034 (v2 changes)

Resolves #1079 

This also adds a bunch of type hints and such, and some additional test coverage.

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -296,7 +296,7 @@ class Client:
         except exceptions.InvalidPath:
             return None
 
-    # TODO(v4.0.0): remove overload when write
+    # TODO(v4.0.0): remove overload when write doesn't use args and kwargs anymore
     @t.overload
     def write(
         self, path: str, write_ttl: t.Optional[str], **kwargs: t.Dict[str, t.Any]
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -1,4 +1,7 @@
 import os
+import typing as t
+
+from warnings import warn
 
 from hvac import adapters, api, exceptions, utils
 from hvac.constants.client import (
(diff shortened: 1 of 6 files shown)


briantist:
@M0NsTeRRR in case you're interested in reviewing

hvac/v1/__init__.py
```diff
@@ -251,35 +296,74 @@ def list(self, path):
         except exceptions.InvalidPath:
             return None
 
-    def write(self, path, wrap_ttl=None, **kwargs):
+    # TODO(v4.0.0): remove overload when write doesn't use args and kwargs anymore
+    @t.overload
+    def write(
+        self, path: str, write_ttl: t.Optional[str], **kwargs: t.Dict[str, t.Any]
```
M0NsTeRRR, line 302: `write_ttl` should be `wrap_ttl`

hvac/v1/__init__.py
```diff
@@ -18,6 +21,48 @@
...
+    *,
+    posvalue: t.Any = _sentinel,
+    method: str = "write",
+    replacement_method: str = "write_data",
+):
+    try:
+        value = dict.pop(member)
+    except KeyError:
+        if posvalue is not _sentinel:
+            return posvalue
```
M0NsTeRRR, line 42: Maybe its better to have only one return (outside of else at the end) and do `value = posvalue` and for the next elif `value = default` for readability ?

hvac/v1/__init__.py
briantist: Great catch, thank you!
```suggestion
        self, path: str, wrap_ttl: t.Optional[str], **kwargs: t.Dict[str, t.Any]
```

hvac/v1/__init__.py
briantist: Maybe personal preference but I prefer the `return` for readability, otherwise I have to follow the rest of the flow to see where the value gets used again and/or returned.

M0NsTeRRR:
LGTM (haven't time to test it)

PR #1095, opened 2023-10-17
Allow verifying recovery keys

wilsonge:
Currently the recovery mechanism only works for rekey's but not for recovery keys. This uses the same parameter we use elsewhere in the file to facilitate that

Documentation for the new path being used here: https://developer.hashicorp.com/vault/api-docs/system/rekey-recovery-key#submit-verification-key

### DIFF
--- a/hvac/api/system_backend/key.py
+++ b/hvac/api/system_backend/key.py
@@ -320,7 +320,7 @@ class Key(SystemBackendMixin):
             url=api_path,
         )
 
-    def cancel_rekey_verify(self):
+    def cancel_rekey_verify(self, recovery_key=False):
         """Cancel any in-progress rekey verification.
         This clears any progress made and resets the nonce. Unlike cancel_rekey, this only resets
         the current verification operation, not the entire rekey atttempt.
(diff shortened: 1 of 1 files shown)


hvac/api/system_backend/key.py
```diff
@@ -320,7 +320,7 @@ def read_backup_keys(self, recovery_key=False):
             url=api_path,
         )
 
-    def cancel_rekey_verify(self):
+    def cancel_rekey_verify(self, recovery_key=False):
```
briantist, line 323:
```suggestion
    def cancel_rekey_verify(self, *, recovery_key=False):
```
Let's make all of these keyword-only. The ability to pass a naked boolean value could be unclear, and future-looking, I'd like to try to avoid adding new positional parameters where possible, because it makes adding new parameters or changing ones in the future more strict since re-ordering is a breaking change.

briantist (commented):
Hi @wilsonge welcome! Thanks for submitting this. I haven't looked too closely yet and it will take some time before I'm able to. In the meantime, I would also like to see tests and documentation/examples for the change included in the PR.

Hopefully there's existing tests and documentation you could extend (I haven't looked yet); if not I understand that adding it from scratch could be a burden.

wilsonge:
> Hopefully there's existing tests and documentation you could extend (I haven't looked yet); if not I understand that adding it from scratch could be a burden.

There aren't :) I did check before I submitted. None of the recovery key based methods are documented or tested sadly.

hvac/api/system_backend/key.py
wilsonge: Are you willing to trade off the consistency of how we're implementing this in the other methods further up which already have the same parameter. If you're happy for the inconsistency happy to roll it out.

briantist: Thanks for pointing that out, I didn't see the other methods. I guess we'll leave them positional for consistency. 

If I had a comprehensive plan/guidance written out about this I might've gone the other way, but I haven't been able to write that yet.

briantist:
> > Hopefully there's existing tests and documentation you could extend (I haven't looked yet); if not I understand that adding it from scratch could be a burden.
> 
> There aren't :) I did check before I submitted. None of the recovery key based methods are documented or tested sadly.

It looks like we have some docs here: https://github.com/hvac/hvac/blob/main/docs/usage/system_backend/key.rst (published: https://hvac.readthedocs.io/en/stable/usage/system_backend/key.html)

And integration tests here: https://github.com/hvac/hvac/blob/main/tests/integration_tests/api/system_backend/test_key.py

PR #1100, opened 2023-10-28
Add support for sys/policies endpoint

### LINKED ISSUE #1025: Missing endpoint for "sys/policies/"
Hi,

Currently the policy backend works with the /sys/policy/ endpoint: https://github.com/hvac/hvac/blob/main/hvac/api/system_backend/policy.py
(hvac docs: https://hvac.readthedocs.io/en/stable/usage/system_backend/policy.html)

According the Hashicorp Vault documentation there exists another backend to manage policies (/sys/policies) and this is where my ACLs are stored. https://developer.hashicorp.com/vault/api-docs/system/policies.

Please either support adding a path parameter to the existing backend so that we can specify different endpoint or create a new backend just for this endpoint.

Thank you!

### LINKED ISSUE #1096: Add support for sys/policies endpoint
Would like to open a PR to add role governing and endpoint governing policies. Enterprise only feature. Please let me know if I am good to proceed! Thanks

https://developer.hashicorp.com/vault/api-docs/system/policies

tot19:
Resolves #1096 
Resolves #1025


tot19:
#1096

briantist (commented):
Thanks this is looking really good. I didn't fully look at it in detail yet but one thing I want to ask about is the `pretty_print` option.

I'm trying to understand why we would ever need or want to format the JSON that way since we're sending it to the Vault server and not really viewing it.

It adds complication to these methods for maintenance and testing.

tot19:
Added that because all of the /sys/policy methods used pretty_print... Will send up an amended PR now

briantist:
> Added that because all of the /sys/policy methods used pretty_print... Will send up an amended PR now.
> 
> I think the original intention is to allow for passing of policies as a dictionary. I'm not a big fan of that, but wanted to keep the two classes fairly consistent.

I get accepting a `dict` and then serializing it, the part I didn't really understand was why we would ever need to format the serialized JSON that way, and then create a new parameter to control that.

I wasn't too aware of what we were doing in the `/sys/policy` methods so I looked just now. I see that those are there but then I had the same question about those; it doesn't seem to make sense to me why we do that.

It looks like it was implemented here:
- https://github.com/hvac/hvac/pull/342

So the description there implies that when queried, even in the Vault UI, I guess Vault will show you the policy exactly as you sent it, and so it's nicer if the policy is formatted in a more readable way.

I haven't confirmed this, but I guess it makes some sense.

In that case, let's keep those, but let's also ensure the tests cover that.

Thanks!

briantist:
btw, if possible it's better (for me) that you do not squash commits locally; all commits will be squashed when the PR is merged. Having individual commits in the PR helps me review incremental changes faster. If you really prefer squashing for your local workflow for some reason that's fine, but it may slow down review.

> Gotcha, pushed up my commit before I saw updated comment. Will fix and open an issue for adding additional /sys/policy tests

Thanks, for test coverage I just meant for this PR, but of course if you want to add test coverage to existing code I would not complain 😅

tot19:
Will not squash locally after this commit. Thanks for the heads up and apologies for the nuisance (just force of habit)

Now that I have reviewed, I will keep pretty_print for ACL policies and drop for RGP/EGP since those are more like code blocks and don't really function same way. Added new test for ACL policy as well ... No new issue needed for /sys/policy because it already has a test that uses a dict policy.

briantist (approved):
Thanks again for another great PR @tot19 ! Sorry for the delays.

PR #1103, opened 2023-10-28
Add Vault 1.15.x to integration tests

### LINKED ISSUE #1075: Add Vault 1.15.x to integration tests
I tried to add Vault 1.15.x in #1074 but there's some behavior difference in the kubernetes auth method that I won't have time to investigate before 2.0.0.

We'll likely have to investigate and fix that at the same time as adding 1.15.x to CI.

tot19:
Looks like the requirement to have a valid JWT token was removed in version 1.15.0 [based on this pull request](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/207). Disabled the jwt error parameter (number 4) for versions 1.15.0 and greater. 

Resolves #1075

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -188,8 +188,7 @@ jobs:
           - "vault=1.12.*"
           - "vault=1.13.*"
           - "vault=1.14.*"
-          # - "vault=1.15.*"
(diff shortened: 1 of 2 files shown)


tot19:
#1075

briantist (approved):
Thanks for digging into this and helping to expand our CI coverage!

PR #1109, opened 2023-12-04
migrate from external mock package to stdlib unittest.mock

eli-schwartz:
It was moved to the stdlib in python 3.3, and the PyPI version is a backport for python versions less than 3.3.


briantist:
Hi @eli-schwartz , I apologize that I didn't get to look at this sooner. It turns out I ended up making these very changes in #1119 . It wasn't the purpose of the PR so I didn't go looking previous issues/PRs at the time. I still appreciate you putting this up and I hope you'll still open PRs for anything else you might find, thanks again!

PR #1118, opened 2023-12-29
Add support for Login MFA

cognifloyd:
Part of #984

Haven't tested this. But it looks complete based on the docs.

I tried to follow the format of the other hvac files, though I would prefer type hints over the doc strings.

Anything else required?


hvac/api/secrets_engines/identity/mfa/okta.py
```diff
@@ -0,0 +1,137 @@
...
+        :type username_format: str | unicode
+        :param base_url: If set, will be used as the base domain for API requests.
+            Examples: okta.com, oktaprevew.com, okta-emea.com
+        :type base_url: str | unicode
+        :param primary_email: If set, the username will match the Okta profile's
+            primary_email instead of the Okta profile's login.
+        :type primary_email: bool
+        :param use_passcode: If set, the user is reminded to use the passcode
+            upon MFA validation.
+        :type use_passcode: bool
```
cognifloyd, line 46: This will be introduced in a future version of vault once this PR is merged in vault:
https://github.com/hashicorp/vault/pull/24563

hvac/api/secrets_engines/identity/mfa/mfa_method_mixin.py
```diff
@@ -0,0 +1,92 @@
+from abc import ABCMeta, abstractmethod
+
+from hvac import utils
+from hvac.api.vault_api_base import VaultApiBase
+from hvac.constants.identity import DEFAULT_MOUNT_POINT
+
+
+class MfaMethodMixin(VaultApiBase, metaclass=ABCMeta):
```
cognifloyd, line 8: How do you feel about the naming of methods for the mfa methods?
You end up using this:
- `client.secrets.identity.mfa.okta.create(...)`
- `client.secrets.identity.mfa.okta.update(...)`
- `client.secrets.identity.mfa.okta.read(...)`
- `client.secrets.identity.mfa.okta.delete(...)`
- `client.secrets.identity.mfa.okta.list(...)`


Or should this instead be:
- `client.secrets.identity.mfa.okta.create_mfa_method(...)`
- `client.secrets.identity.mfa.okta.update_mfa_method(...)`
- `client.secrets.identity.mfa.okta.read_mfa_method(...)`
- `client.secrets.identity.mfa.okta.delete_mfa_method(...)`
- `client.secrets.identity.mfa.okta.list_mfa_methods(...)`

And should the totp-specific methods be renamed?
- `client.secrets.identity.mfa.totp.generate(...)` -> `client.secrets.identity.mfa.totp.generate_mfa_secret(...)`
- `client.secrets.identity.mfa.totp.admin_generate(...)` -> `client.secrets.identity.mfa.totp.admin_generate_mfa_secret(...)`
- `client.secrets.identity.mfa.totp.admin_destroy(...)` -> `client.secrets.identity.mfa.totp.admin_destroy_mfa_secret(...)`

tot19:
Thanks for the awesome contribution! 

I think we would need unit tests added for this since there is a lot of net new code. Do you need assistance with that?

cognifloyd:
I'm not sure what to unit-test because there is very little logic to test. Many other endpoint methods are untested. For example, I didn't see any unit tests for identity endpoints.

What form should be unit tested? Mock out requests and assert that the correct API path is used? That puts a magic string in more places increasing maintenance burden without adding much value.

I could break this up into 7 smaller PRs:
- 1 minor refactor
- 1 for the sys/mfa endpoint
- 1 for turning identity into a package w/ mfa login enforcement endpoints
- 4 (1/MFA method) for the MFA methods (PingID, TOTP, Duo, Okta)

Would that be helpful?

cognifloyd:
First mini-PR extracted from this PR: #1123

briantist:
Hi @cognifloyd , thanks, I haven't had a chance to look this over yet, sorry for the delay. Please hold off on splitting out to other PRs before getting more review.

cognifloyd:
When reviewing, it will be helpful to step through the commits.

briantist:
Hi @cognifloyd I haven't forgotten about this I've just been very short on time to put toward this project recently, I appreciate your patience.

---

> I'm not sure what to unit-test because there is very little logic to test. Many other endpoint methods are untested. For example, I didn't see any unit tests for identity endpoints.

> What form should the unit tests take? Mock out requests and assert that the correct API path is used? That puts a magic string in more places increasing maintenance burden without adding much value.

We have both unit and integration tests. 

In the integration tests, we run an actual Vault server so we typically try to actually test the methods as best we can, setting up the various backends, running test cases, etc.

In unit tests we don't run a Vault server, we often mock responses so that `requests` calls can be made normally and the responses library handles it.

Integration tests are sometimes harder to write, but they give us better real coverage (not just a number of lines), but the flip side is that it can be difficult or impossible to cover some uses in our tests and CI, especially when it requires 3rd party responses, so units can be really helpful for those.

Although there's not much logic in these methods, even testing that the right parameters get passed into the adapter call is useful coverage! Mocking might help with that to be able to check the parameters passed.

Take a look at existing tests to get a sense of how some existing things do it.

I know there are a lot of methods without tests in the project, filling out tests for those is something I'd love help with.

If you have a hard time filling out tests, hopefully we can get you some help, from me or @tot19 or anyone else who'd like to work with you on it.

Thank you again for the contribution it's definitely appreciated.

hvac/api/secrets_engines/identity/mfa/okta.py
cognifloyd: This will land in vault 2.17

PR #1119, opened 2023-12-30
fix readthedocs (RTD) public docsite, modernize some of the docs build/testing

### LINKED ISSUE #1053: Move docs dependencies into their own group
Now that we have poetry groups set up, with optional dev dependencies in their own group, I think we can further split off docs dependencies into their own group.

The reasons I want to do this:
- encourage more documentation contributions and make it easier for minimal testing of them locally
- to clarify dependencies for readthedocs (RTD)
- to aid in dependency installation for #1042 
- to give an option not to install these for folks who only want to run pytest locally
- make maintenance of these dependencies easier when we know they don't conflict with other dev dependencies
- to clarify dependencies around doctest

That last point is where I think we're going to hit some snags. Doctest is used to test code examples in the documentation, and we've done some weird things with it. It's a little hard to pick apart right now, and I think that stuff needs maintenance. It would be helpful for us to figure out what the future is for doctest, whether we want to lean into it and shore it up, or remove it (and its dependencies).

### LINKED ISSUE #1061: readthedocs: `latest` tag is broken due to bad default branch configuration
After we changed the default branch from `develop` to `main`, the `latest` tag in our docs is stuck on that branch's last commit: https://hvac.readthedocs.io/en/latest/changelog.html

This is because our RTD config was set up with that default branch, and not updated. However, I don't have access to the RTD account that "owns" the `hvac` project and so I cannot change the settings. I also tried importing the project into my own account but it wouldn't let me.

We will need to try contacting previous maintainers to see how we can get access.

In addition, the default branch is one of the properties we cannot set in the config file:
https://docs.readthedocs.io/en/stable/config-file/v2.html#migrating-from-the-web-interface

This _does not affect_ the `stable` tag (which is the default), which corresponds to the latest SCM tagged release:
https://hvac.readthedocs.io/en/stable/changelog.html

The `main` tag is also correct and is what `latest` should be:
https://hvac.readthedocs.io/en/main/changelog.html

briantist:
Fixes #1061 
Resolves #1053 

I got added to the RTD maintainers and was able to update several settings, like our defualt branch.
I've also enabled pull request builds, so we can see when a PR would break RTD (or in the case of this PR, fixes it), and maintainers can also view the generated documentation which is nice.

Fixing the build involved updating a lot of outdated requirements, so I also moved all the requirements into poetry as described in #1053 .

Along the way, updated the extremely our-of-date `sphinx` and `sphinx-rtd-theme` versions, so maybe the docs will even look a little nicer ✨.

The requirements upgrades needed some changes throughout other test code to remove some deprecated things.

I didn't go as far as trying to untangle our custom doctest thing and remove that, there's enough in this PR already and I want to get the public docsite fixed. That work will probably be part of or a pre-requisite for trying to move doctests into pytest, described a bit in #1105 .


briantist:
All changes here only affect dev/tests,a nd the public docsite, no changes to the library.

PR #1120, opened 2023-12-30
`Client.write_data` - remove potentially dangerous default

### LINKED ISSUE #1104: `Client.write_data` default value `{}` should be changed
https://github.com/hvac/hvac/blob/ede86220b783aa68acc63b23361ca6714a74f3a3/hvac/v1/__init__.py#L351

https://pylint.pycqa.org/en/latest/user_guide/messages/warning/dangerous-default-value.html

It seems this default was a bad idea.

As far as I know, all the code this value will go through will not mutate the value, but we should fix this before that situation changes.

I think we'll be able to default it to `None` and everything will still work correctly even if we pass it to `post` that way; worst case it doesn't and we default to `{}` inside the function if it's `None`.

Either way this change won't be breaking and will prevent future pain, breakage, and surprise.

---

I found this in a downstream project that uses pylint. It raises the question of whether we should maybe be using pylint, or perhaps #1059 would have caught it as well?

briantist:
Fixes #1104

### DIFF
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -348,7 +348,7 @@ class Client:
         self,
         path: str,
         *,
-        data: t.Dict[str, t.Any] = None,
+        data: t.Optional[t.Dict[str, t.Any]] = None,
         wrap_ttl: t.Optional[str] = None,
     ):
         """Write data to a path. Similar to write() without restrictions on data keys.
--- a/hvac/v1/__init__.py
+++ b/hvac/v1/__init__.py
@@ -348,7 +348,7 @@ class Client:
         self,
         path: str,
         *,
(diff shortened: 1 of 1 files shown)


tot19:
Looks good to me ... Good catch. One potential addition here would be to add a test to catch future instances of this? Know you mentioned pylint, and I generally have had good experience with it.

briantist:
Thanks for the review @tot19 ! I think I will address detection separately. Before adding a fourth(!) linter/checker to the project, I will probably evaluate this open PR https://github.com/hvac/hvac/pull/1059 to add `ruff`. Maybe it would have caught this too, and we can reduce the number of tools. In the meantime I'll merge this one.

PR #1121, opened 2023-12-30
try not skipping on cache hit

briantist:
Since #1119 , I've noticed that doctests run are failing, unable to find `sphinx-build`, unless I delete the cache and let it reinstall the dependencies. I guess the caching mechanism is not keeping the bin type files so the command `sphinx-build` just doesn't exist?

Trying to fix without completely uncaching...

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -88,7 +88,7 @@ jobs:
         cache: poetry
 
     - name: Install dependencies
-      run: poetry install --no-interaction --no-root --with dev
+      run: poetry install --no-interaction --no-root --with dev,docs
 
     - name: Install library
       run: poetry install --no-interaction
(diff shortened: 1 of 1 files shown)


briantist:
Seems to work! https://github.com/hvac/hvac/actions/runs/7366531054/job/20048899106?pr=1121#step:4:38

The first run did not find the cache, then I re-ran the docs job and it did find the cache. Running the install step again still had to install packages but not all of them, and it only took around 5 seconds.

The reason this is happening is because all the other runs don't install the `docs` group, so if another run populated the cache, then when it gets restored in this run, it does not include docs deps.

There's a possibility that in some workflow run, this job would be the first to populate a cache, and then the other jobs would all be restoring a cache that includes docs dependencies, but that is not any kind of problem.

---

Merging since this has no user-facing changes.

PR #1125, opened 2024-01-23
aws auth - add support for `iam_metadata` and `ec2_metadata` params

### LINKED ISSUE #1029: Add support for iam_metadata within aws identity integration
There is currently a gap between what is documented in [HVAC](https://hvac.readthedocs.io/en/stable/source/hvac_api_auth_methods.html#hvac.api.auth_methods.Aws.configure_identity_integration) vs [Vault's API](https://developer.hashicorp.com/vault/api-docs/auth/aws#configure-identity-integration). There is no parameter available to adjust the iam_metadata or ec2_metadata within configure_identity_integration.

danholodak:
Resolves #1029 

- Support the 2 metadata parameters for AWS configure identity integration
- Add integration tests for configure_identity_integration

### DIFF
--- a/hvac/api/auth_methods/aws.py
+++ b/hvac/api/auth_methods/aws.py
@@ -158,7 +158,7 @@ class Aws(VaultApiBase):
             To use only particular fields, select the explicit fields. To restore to defaults, send only a field of ``default``.
             Only select fields that will have a low rate of change for your ``iam_alias`` because each change triggers a storage
             write and can have a performance impact at scale.
-        :type iam_metadata: str | unicode, list
+        :type iam_metadata: str | unicode | list
         :param ec2_alias: Configures how to generate the identity alias when using the ec2 auth method. Valid choices
             are role_id, instance_id, and image_id. When role_id is selected, the randomly generated ID of the role is
             used. When instance_id is selected, the instance identifier is used as the identity alias name. When
@@ -170,7 +170,7 @@ class Aws(VaultApiBase):
             To use only particular fields, select the explicit fields. To restore to defaults, send only a field of ``default``.
             Only select fields that will have a low rate of change for your ``ec2_alias`` because each change triggers a storage
             write and can have a performance impact at scale.
-        :type ec2_metadata: str | unicode, list
+        :type ec2_metadata: str | unicode | list
         :param mount_point: The path the AWS auth method was mounted on.
         :type mount_point: str | unicode
         :return: The response of the request
--- a/hvac/api/auth_methods/aws.py
+++ b/hvac/api/auth_methods/aws.py
@@ -177,10 +177,10 @@ class Aws(VaultApiBase):
         :rtype: request.Response
         """
         if iam_alias is not None and iam_alias not in ALLOWED_IAM_ALIAS_TYPES:
-            error_msg = f"invalid iam alias type provided: '{iam_alias}'; supported iam alias types: '{','.join(ALLOWED_IAM_ALIAS_TYPES)}'"
+            error_msg = f"invalid iam alias type provided: '{iam_alias}' - supported iam alias types: '{','.join(ALLOWED_IAM_ALIAS_TYPES)}'"
             raise exceptions.ParamValidationError(error_msg)
         if ec2_alias is not None and ec2_alias not in ALLOWED_EC2_ALIAS_TYPES:
-            error_msg = f"invalid ec2 alias type provided: '{ec2_alias}'; supported ec2 alias types: '{','.join(ALLOWED_EC2_ALIAS_TYPES)}'"
+            error_msg = f"invalid ec2 alias type provided: '{ec2_alias}' - supported ec2 alias types: '{','.join(ALLOWED_EC2_ALIAS_TYPES)}'"
             raise exceptions.ParamValidationError(error_msg)
 
         params = utils.remove_nones(
--- a/hvac/api/auth_methods/aws.py
+++ b/hvac/api/auth_methods/aws.py
@@ -128,7 +128,12 @@ class Aws(VaultApiBase):
         return self._adapter.delete(url=api_path)
 
     def configure_identity_integration(
-        self, iam_alias=None, ec2_alias=None, mount_point=AWS_DEFAULT_MOUNT_POINT
+        self,
+        iam_alias=None,
+        ec2_alias=None,
+        mount_point=AWS_DEFAULT_MOUNT_POINT,
+        iam_metadata=None,
+        ec2_metadata=None,
     ):
         """Configure the way that Vault interacts with the Identity store.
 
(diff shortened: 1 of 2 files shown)


hvac/api/auth_methods/aws.py
```diff
@@ -128,7 +128,7 @@ def delete_config(self, mount_point=AWS_DEFAULT_MOUNT_POINT):
         return self._adapter.delete(url=api_path)
 
     def configure_identity_integration(
-        self, iam_alias=None, ec2_alias=None, mount_point=AWS_DEFAULT_MOUNT_POINT
+        self, iam_alias=None,  iam_metadata=None, ec2_alias=None, ec2_metadata=None, mount_point=AWS_DEFAULT_MOUNT_POINT
```
briantist, line 131: Let's move the new parameters to the end of the parameter list, that way we don't break anyone using positional arguments.

hvac/api/auth_methods/aws.py
```diff
@@ -146,34 +146,52 @@ def configure_identity_integration(
             select full_arn and then delete and recreate the IAM role, Vault won't be aware and any identity aliases
             set up for the role name will still be valid
         :type iam_alias: str | unicode
+        :param iam_metadata: The metadata to include on the token returned by the login endpoint.
+            This metadata will be added to both audit logs, and on the iam_alias. By default, it includes account_id 
+            and auth_type.Additionally, canonical_arn, client_arn, client_user_id, inferred_aws_region, inferred_entity_id, 
+            and inferred_entity_type are available. To include no metadata, set to "" via the CLI or [] via the API. 
+            To use only particular fields, select the explicit fields. To restore to defaults, send only a field of default. 
+            Only select fields that will have a low rate of change for your iam_alias because each change triggers a storage
```
briantist, line 154:
```suggestion
            This metadata will be added to both audit logs, and on the ``iam_alias``. By default, it includes ``account_id`` 
            and ``auth_type``. Additionally, ``canonical_arn``, ``client_arn``, ``client_user_id``, ``inferred_aws_region``, ``inferred_entity_id``, 
            and ``inferred_entity_type`` are available. To include no metadata, set to an empty list ``[]``. 
            To use only particular fields, select the explicit fields. To restore to defaults, send only a field of ``default``. 
            Only select fields that will have a low rate of change for your ``iam_alias`` because each change triggers a storage 
```
nits and formatting

hvac/api/auth_methods/aws.py
```diff
@@ -146,34 +146,52 @@ def configure_identity_integration(
...
         :param ec2_alias: Configures how to generate the identity alias when using the ec2 auth method. Valid choices
             are role_id, instance_id, and image_id. When role_id is selected, the randomly generated ID of the role is
             used. When instance_id is selected, the instance identifier is used as the identity alias name. When
             image_id is selected, AMI ID of the instance is used as the identity alias name
         :type ec2_alias: str | unicode
+        :param ec2_metadata: The metadata to include on the token returned by the login endpoint. This metadata will be 
+            added to both audit logs, and on the ec2_alias. By default, it includes account_id and auth_type. Additionally, 
+            ami_id, instance_id, and region, are available. To include no metadata, set to "" via the CLI or [] via the API. 
+            To use only particular fields, select the explicit fields. To restore to defaults, send only a field of default. 
+            Only select fields that will have a low rate of change for your ec2_alias because each change triggers a storage
```
briantist, line 166:
```suggestion
            added to both audit logs, and on the ``ec2_alias``. By default, it includes ``account_id`` and ``auth_type``. Additionally, 
            ``ami_id``, ``instance_id``, and ``region`` are available. To include no metadata, set to an empty list ``[]``. 
            To use only particular fields, select the explicit fields. To restore to defaults, send only a field of ``default``. 
            Only select fields that will have a low rate of change for your ``ec2_alias`` because each change triggers a storage 
```
nits and formatting

hvac/api/auth_methods/aws.py
```diff
@@ -146,34 +146,52 @@ def configure_identity_integration(
...
+            write and can have a performance impact at scale.
+        :type ec2_metadata: str | unicode, list
         :param mount_point: The path the AWS auth method was mounted on.
         :type mount_point: str | unicode
         :return: The response of the request
         :rtype: request.Response
         """
         if iam_alias is not None and iam_alias not in ALLOWED_IAM_ALIAS_TYPES:
-            error_msg = 'invalid iam alias type provided: "{arg}"; supported iam alias types: "{alias_types}"'
+            error_msg = f'invalid iam alias type provided: "{iam_alias}"; supported iam alias types: "{ALLOWED_IAM_ALIAS_TYPES}"'
```
briantist, line 175: The `error_msg.format()` line is what was supposed be templating these. `ALLOWED_EC2_ALIAS_TYPES` was supposed to be `join`ed, but it seems like that was never happening due to a mismatch between `alias_types` in the template and `environments` in the format call.

I'm ok with updating it to an f-string, but I think we still need to do the join somewhere, and we would need to remove the format call.

hvac/api/auth_methods/aws.py
```diff
@@ -146,34 +146,52 @@ def configure_identity_integration(
...
             raise exceptions.ParamValidationError(
                 error_msg.format(
                     arg=iam_alias, environments=",".join(ALLOWED_IAM_ALIAS_TYPES)
                 )
             )
         if ec2_alias is not None and ec2_alias not in ALLOWED_EC2_ALIAS_TYPES:
-            error_msg = 'invalid ec2 alias type provided: "{arg}"; supported ec2 alias types: "{alias_types}"'
+            error_msg = f'invalid ec2 alias type provided: "{ec2_alias}"; supported ec2 alias types: "{ALLOWED_EC2_ALIAS_TYPES}"'
```
briantist, line 182: see above comment

briantist (commented):
Thanks @danholodak ! I didn't look closely at the tests yet, just some initial comments. You'll also want to be sure you run the `black` formatter (and other linters) before committing:
```
poetry run black .
poetry run flake8 .
poetry run typos
```

hvac/api/auth_methods/aws.py
danholodak: Good call, took care of that. Do you resolve these threads when you re-review or should I be doing that?

hvac/api/auth_methods/aws.py
danholodak: Thanks, I missed that .format functionality - added the join into the f-string.

danholodak:
Thanks for the quick review, I believe I've addressed all the comments and I ran the linters. Looking forward to your thoughts on the tests.

hvac/api/auth_methods/aws.py
```diff
@@ -146,34 +151,44 @@ def configure_identity_integration(
...
         :rtype: request.Response
         """
         if iam_alias is not None and iam_alias not in ALLOWED_IAM_ALIAS_TYPES:
-            error_msg = 'invalid iam alias type provided: "{arg}"; supported iam alias types: "{alias_types}"'
-            raise exceptions.ParamValidationError(
-                error_msg.format(
-                    arg=iam_alias, environments=",".join(ALLOWED_IAM_ALIAS_TYPES)
-                )
-            )
+            error_msg = f'invalid iam alias type provided: "{iam_alias}"; supported iam alias types: "{",".join(ALLOWED_IAM_ALIAS_TYPES)}"'
```
briantist, line 180: f-strings only work with double quotes. You can make the quotes inside the strings into single quotes `'` or you can use an alternative method to get double-quotes in there: https://stackoverflow.com/q/17085354/3905079

As is, it's being treated by multiple statements separated by semi-colon, which is why flake8 is failing.

danholodak: Thanks for catching that, not sure why flake8 didn't catch it when I ran it on my end.

briantist:
minor ask: I would prefer if you don't squash commits on your end for the most part. It's helpful for me to see individual commits so I can review individual changes as we go along. I know some projects use "conventional commits" or have that workflow; in this project all commits in a PR are squashed into one on merge anyway, so no need to like "clean up" the commits locally.

If you really like to do it that way you can, it might slow me down a bit though, just my preference.

danholodak:
> minor ask: I would prefer if you don't squash commits on your end for the most part. 

Totally happy not to squash going forward, thanks for letting me know.

hvac/api/auth_methods/aws.py
```diff
@@ -146,34 +151,44 @@ def configure_identity_integration(
...
             set up for the role name will still be valid
         :type iam_alias: str | unicode
+        :param iam_metadata: The metadata to include on the token returned by the login endpoint.
+            This metadata will be added to both audit logs, and on the ``iam_alias``. By default, it includes ``account_id``
+            and ``auth_type``. Additionally, ``canonical_arn``, ``client_arn``, ``client_user_id``, ``inferred_aws_region``, ``inferred_entity_id``,
+            and ``inferred_entity_type`` are available. To include no metadata, set to an empty list ``[]``.
+            To use only particular fields, select the explicit fields. To restore to defaults, send only a field of ``default``.
+            Only select fields that will have a low rate of change for your ``iam_alias`` because each change triggers a storage
+            write and can have a performance impact at scale.
+        :type iam_metadata: str | unicode, list
```
briantist, line 161:
```suggestion
        :type iam_metadata: str | unicode | list
```

hvac/api/auth_methods/aws.py
```diff
@@ -146,34 +151,44 @@ def configure_identity_integration(
...
             used. When instance_id is selected, the instance identifier is used as the identity alias name. When
             image_id is selected, AMI ID of the instance is used as the identity alias name
         :type ec2_alias: str | unicode
+        :param ec2_metadata: The metadata to include on the token returned by the login endpoint. This metadata will be
+            added to both audit logs, and on the ``ec2_alias``. By default, it includes ``account_id`` and ``auth_type``. Additionally,
+            ``ami_id``, ``instance_id``, and ``region`` are available. To include no metadata, set to an empty list ``[]``.
+            To use only particular fields, select the explicit fields. To restore to defaults, send only a field of ``default``.
+            Only select fields that will have a low rate of change for your ``ec2_alias`` because each change triggers a storage
+            write and can have a performance impact at scale.
+        :type ec2_metadata: str | unicode, list
```
briantist, line 173:
```suggestion
        :type ec2_metadata: str | unicode | list
```

briantist (commented):
two final documentation nits, which I'll take care of committing

briantist (approved):
@danholodak thanks very much for this contribution!

PR #1136, opened 2024-02-21
Add sts_region parameter to auth.aws.configure

ewanoomen:
Adds the sts_region parameter to auth.aws.configure to configure the parameter mentioned here: https://developer.hashicorp.com/vault/api-docs/auth/aws#sts_region

### DIFF
--- a/hvac/api/auth_methods/aws.py
+++ b/hvac/api/auth_methods/aws.py
@@ -26,9 +26,9 @@ class Aws(VaultApiBase):
         endpoint=None,
         iam_endpoint=None,
         sts_endpoint=None,
-        sts_region=None,
         iam_server_id_header_value=None,
         mount_point=AWS_DEFAULT_MOUNT_POINT,
+        sts_region=None,
     ):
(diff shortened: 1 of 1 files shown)


hvac/api/auth_methods/aws.py
```diff
@@ -26,6 +26,7 @@ def configure(
         endpoint=None,
         iam_endpoint=None,
         sts_endpoint=None,
+        sts_region=None,
```
briantist, line 29: We'll need to move this to the end of the parameter list to avoid breaking users setting their arguments positionally.

briantist:
Hi @ewanoomen , could you fix the linting errors and move the parameter to the end of the list? I think we could get this merged pretty quickly.

ewanoomen:
> Hi @ewanoomen , could you fix the linting errors and move the parameter to the end of the list? I think we could get this merged pretty quickly.

Sorry for the late reply, I've moved the parameter and linting should be fixed as well

briantist (approved):

PR #1144, opened 2024-03-19
Adding namespace_in_state to jwt

### LINKED ISSUE #995: Add support for namespace_in_state field on JWT
Ref documentation: https://developer.hashicorp.com/vault/api-docs/auth/jwt#namespace_in_state

Request for adding of namespace_in_state field to jwt api.

To be implemented on the client.auth.jwt.configure method.  Thanks.

valleedelisle:
Adding this [parameter](https://developer.hashicorp.com/vault/api-docs/auth/jwt#namespace_in_state) for completeness.

Fix #995

### DIFF
--- a/hvac/api/auth_methods/jwt.py
+++ b/hvac/api/auth_methods/jwt.py
@@ -39,6 +39,7 @@ class JWT(VaultApiBase):
         jwks_ca_pem=None,
         jwt_validation_pubkeys=None,
         bound_issuer=None,
+        namespace_in_state=None,
         jwt_supported_algs=None,
         default_role=None,
         provider_config=None,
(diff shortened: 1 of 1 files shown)


hvac/api/auth_methods/jwt.py
```diff
@@ -39,6 +39,7 @@ def configure(
         jwks_ca_pem=None,
         jwt_validation_pubkeys=None,
         bound_issuer=None,
+        namespace_in_state=None,
```
briantist, line 42: Please move this parameter to the bottom of the list, that way clients supplying values positionally will not experience a breaking change.

briantist (commented):
Hello @valleedelisle welcome and thanks for putting up this PR!

hvac/api/auth_methods/jwt.py
valleedelisle: Makes sense. I moved it before `path`, let me know if that's fine.

briantist: Unfortunately we need to move it _after_ `path` since `path` can still be supplied positionally.

I would like to go through the entire library and make most of the parameters keyword-only, so that we can avoid this in the future (it comes up in almost every PR on existing methods), but I just haven't found the time to write up the policy and rollout plan (it's a breaking change, so I want to do it over the course of a few major releases, with plenty of warning for anyone using the arguments positionally).

valleedelisle: Yeah I wasn't sure. I always use `*kw` anyways, it's safer.

briantist (approved):
Thanks for contributing this @valleedelisle !

Issue #1142, opened 2024-03-17
Some tests use distutils module that has been removed in python 3.12


loqs:
[PEP 632](https://peps.python.org/pep-0632/) deprecated distutils in python 3.10 and its removal followed in python 3.12.

It is used in three test files:
https://github.com/hvac/hvac/blob/b61d05d9055e79b0c1a6b41a333d78fb7dd6aee0/tests/utils/server_manager.py#L10
https://github.com/hvac/hvac/blob/b61d05d9055e79b0c1a6b41a333d78fb7dd6aee0/tests/utils/hvac_integration_test_case.py#L8
https://github.com/hvac/hvac/blob/b61d05d9055e79b0c1a6b41a333d78fb7dd6aee0/tests/integration_tests/api/auth_methods/test_ldap.py#L1
Failure to find the module causes testing to failure during test discovery. The suggested replacement for distutils.spawn.find_executable as per PEP 632 is shutil.which.

briantist:
@loqs thanks for raising this. I remember making this replacement in some other project I was working on... and now can't remember which 😅 but the strange thing is, we already test against 3.12 and it seems like there are no issues
- https://github.com/hvac/hvac/pull/1073

Could you let me know where you are seeing failures? Running locally?

Also if you'd like to put up a PR to change this it would be appreciated.

loqs:
> Could you let me know where you are seeing failures? Running locally?

Yes locally on Arch Linux.  One explanation you are not seeing the issue would be if setuptools is installed as that [bundles distutils](https://gitlab.archlinux.org/archlinux/packaging/packages/python-hvac/-/merge_requests/1#note_172148).
 
> Also if you'd like to put up a PR to change this it would be appreciated.

Done.

briantist:
Related (probably should have caught this earlier):
- #1054 
- #1055

PR #1146, opened 2024-03-23
Replace distutils.spawn with shutil.which

loqs:
distutils was removed in python 3.12.
Closes https://github.com/hvac/hvac/issues/1142


briantist (approved):
Thanks so much for putting this up @loqs !

PR #1148, opened 2024-04-04
Add new exception types for HTTP status 405 and 412

### LINKED ISSUE #1093: Add an exception type for HTTP status code `412`
Right now there's no exception type for 412 status codes that get returned for eventual consistency failures, so we return `UnexpectedError` which is not very helpful on its own.

https://github.com/hvac/hvac/blob/8829ff20840743f89a90ec4319e8974fade18cd7/hvac/exceptions.py#L21

Adding a dedicated type would be good for being able to catch this, or at least getting more useful feedback, though I'll stress that retries are better handled during the request: https://hvac.readthedocs.io/en/stable/advanced_usage.html#retrying-failed-requests

mweigel:
# HTTP exception type changes

Any HTTP exception in `hvac` which does not have a pre-existing mapped status code to exception will return an `UnexpectedError` exeption.

We would like to add exception types for status codes [412 (`PreconditionFailed`)](#1093) and 405 (`UnsupportedOperation`).

## Breaking change

Existing code that wants to catch those exceptions may be written similarly to this:

```python
try:
    client.do_thing()
except UnexpectedError as e:
    # check status code in e and handle appropriately
```

If we then map those status codes to new exception types, the above code will fail to catch them.

As a result, we will do this change in two stages:

- This PR introduces the new classes, but **does not** map the existing status codes to them, so those exceptions will never be raised. This change will be added to `hvac` 2.2.0.
- #1149 will be introduced in `hvac` 3.0.0 and will map the status codes.

# Action needed

If you currently use `UnexpectedError` to catch status codes 405 or 412, update your code **before v3.0.0** to something like the following (using `UnsupportedOperation` as the example):

```python
try:
   client.do_thing()
except (UnexpectedError, UnsupportedOperation) as e:
    # handle
```

or if you want to handle them differently:

```python
try:
   client.do_thing()
except UnsupportedOperation as e:
    # handle
except UnexpectedError as e:
    # handle
```

This ensures that your code is ready to handle the new exception type. This is especially important if your project intends to support multiple major versions of `hvac`.

# Future exception types

If you are using `UnexpectedError` as a catch-all for any exception without a dedicated type, and you want to future proof against more HTTP exception types being added in the future, you may consider catching the base class `VaultError` instead or in addition to `:

```python
try:
    client.do_thing()
except Unauthorized:
    # handle unauthorized
except Forbidden:
    # handle forbidden
except VaultError as e:
    # conditionally handle any exception not named in a previous except block
```

---

* Prepare for #1093 by adding PreconditionFailed exception for status code 412
* Add UnsupportedOperation exception for status code 405

Adding the UnsupportedOperation exception is a reasonably significant change as there might be quite a bit of code expecting UnexpectedError on a 405. Can take it out if it's too much.

### DIFF
--- a/hvac/exceptions.py
+++ b/hvac/exceptions.py
@@ -49,6 +49,14 @@ class InvalidPath(VaultError):
     pass
 
 
+class UnsupportedOperation(VaultError):
+    pass
+
+
+class PreconditionFailed(VaultError):
+    pass
+
+
 class RateLimitExceeded(VaultError):
     pass
 


briantist:
Thanks for putting this up @mweigel ! I'll need some time to fully review it as I have a busy couple of weeks ahead.

One thing I should have mentioned in the original issue, is that while we can introduce the new exceptions in a minor release, we can't update the status map until a major release.

The reason is that if people have code that is expecting to catch `UnknownException` to deal with 412 and 405 now, their code will break once we raise a new exception type.

So I think we should split the PR; the new exception classes are introduced in the next (minor) release, and that lets folks update their `except` blocks to catch both `UnknownException` and the new one(s), then in the next major version we'll update the status map.

Does that make sense to you?

briantist (approved):
I want to edit the PR description as a signpost for people reading the changelog, to give instructions on what to do before merging, but this will bein the next release, thank you!

briantist:
@mweigel if you wouldn't mind also submitting the PR containing the status map changes and tests, that would be great, that way we have it prepared and reviewed and ready for 3.0.0.

mweigel:
Cool, pushed https://github.com/hvac/hvac/pull/1149

PR #1151, opened 2024-04-09
sys/tools/random

michaelkosir:
**New:**
- Added support for generating random bytes using `sys/tools/random`.
- Docs
- Tests

**Updated:**
- Transit: Added support for specifying entropy source (platform, seal, all) when generating random bytes.

**Reference:**
- https://developer.hashicorp.com/vault/api-docs/system/tools#generate-random-bytes
- https://developer.hashicorp.com/vault/api-docs/secret/transit#generate-random-bytes


hvac/api/secrets_engines/transit.py
```diff
@@ -603,7 +603,7 @@ def generate_data_key(
         )
 
     def generate_random_bytes(
-        self, n_bytes=None, output_format=None, mount_point=DEFAULT_MOUNT_POINT
+        self, n_bytes=None, output_format=None, source="", mount_point=DEFAULT_MOUNT_POINT
```
briantist, line 606: `source` must be moved to the end of the parameter list so that it doesn't break clients using positional arguments. We should also default it to `None` and then the method needs to handle that by not templating the value into the URL.

hvac/api/secrets_engines/transit.py
```diff
@@ -615,6 +615,10 @@ def generate_random_bytes(
         :type n_bytes: int
         :param output_format: Specifies the output encoding. Valid options are hex or base64.
         :type output_format: str | unicode
+        :param source: Specifies the source of the requested bytes. platform, the default, 
+            sources bytes from the platform's entropy source. seal sources from entropy augmentation (enterprise only). 
+            all mixes bytes from all available sources.
```
briantist, line 620:
```suggestion
        :param source: Specifies the source of the requested bytes. ``platform``, the default, 
            sources bytes from the platform's entropy source. ``seal`` sources from entropy augmentation (enterprise only). 
            ``all`` mixes bytes from all available sources.
```

hvac/api/system_backend/tools.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+from hvac import utils
+from hvac.api.system_backend.system_backend_mixin import SystemBackendMixin
+
+
+class Tools(SystemBackendMixin):
+    def generate_random_bytes(
+        self, n_bytes=None, output_format=None, source=""
```
briantist, line 8: Please default `source` to `None` and handle the templating needs in the method conditionally.

hvac/api/system_backend/tools.py
```diff
@@ -0,0 +1,37 @@
...
+            POST: /sys/tools/random(/:source)(/:bytes). Produces: 200 application/json
+
+        :param n_bytes: Specifies the number of bytes to return. This value can be specified either in the request body,
+            or as a part of the URL.
+        :type n_bytes: int
+        :param output_format: Specifies the output encoding. Valid options are hex or base64.
+        :type output_format: str | unicode
+        :param source: Specifies the source of the requested bytes. platform, the default, 
+            sources bytes from the platform's entropy source. seal sources from entropy augmentation (enterprise only). 
+            all mixes bytes from all available sources.
```
briantist, line 22:
```suggestion
        :param source: Specifies the source of the requested bytes. ``platform``, the default, 
            sources bytes from the platform's entropy source. ``seal`` sources from entropy augmentation (enterprise only). 
            ``all`` mixes bytes from all available sources.
```

briantist (commented):
Hi @michaelkosir , welcome and thanks for submitting this! I added a few initial comments, I will do a more in-depth review as soon as I find the time.

tests/integration_tests/api/system_backend/test_tools.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TestTools(HvacIntegrationTestCase, TestCase):
+    @parameterized.expand(
+        [
+            param(
+                "success",
+            ),
+        ]
+    )
+    def test_generate_random_bytes(
```
briantist, line 18: looks like we're not testing the `source` parameter, let's add some test cases and/or a new test method for testing the various `source` options

briantist (commented):
Seems like you may need to check the linting/formatting

PR #1155, opened 2024-04-13
add Vault 1.16.x to CI matrix

### LINKED ISSUE #1154: Add Vault 1.16.x to CI matrix
Vault 1.16 is out, need to update the matrix to test against it and see what breaks.
- https://github.com/hvac/hvac/blob/main/.github/workflows/lint-and-test.yml#L184
- https://github.com/hvac/hvac/issues/1152

briantist:
Resolves #1154

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -189,6 +189,7 @@ jobs:
           - "vault=1.13.*"
           - "vault=1.14.*"
           - "vault=1.15.*"
+          - "vault=1.16.*"
 
     steps:
     - uses: actions/checkout@v4
--- a/tests/integration_tests/api/auth_methods/test_jwt.py
+++ b/tests/integration_tests/api/auth_methods/test_jwt.py
@@ -217,15 +217,6 @@ class TestJWT(HvacIntegrationTestCase, TestCase):
             second=response.status_code,
         )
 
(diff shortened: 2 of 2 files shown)


briantist:
JWT error:
- > hvac.exceptions.InvalidRequest: error checking jwks URL https://127.0.0.1:53541/v1/identity/oidc/.well-known/keys, on post https://127.0.0.1:53541/v1/auth/jwt-test/config
hvac/utils.py:41: InvalidRequest
- https://github.com/hvac/hvac/actions/runs/8674145179/job/23786167385?pr=1155#step:8:105

This turns out to be a known issue in 1.16.1, and it seems that it will be fixed it 1.16.2:
- https://github.com/hashicorp/vault-plugin-auth-jwt/issues/294#issuecomment-2046107870
- https://developer.hashicorp.com/vault/docs/upgrading/upgrade-to-1.16.x#error-configuring-the-jwt-auth-method
- https://github.com/hashicorp/vault-plugin-auth-jwt/pull/290

briantist:
In their own test suite, they updated the test JWT to use a different signature, but I couldn't figure out where to do that in our own test suite, and since the issue claims it will be fixed in 1.16.2, I opted to just skip the test for now on Vault 1.16.1.

If we see it failing again with 1.16.2 we'll look into it further then.

PR #1161, opened 2024-04-15
Implement activedirectory.rotate_credentials and ldap.rotate_static_credentials

hkgnobody:
usage: client.secrets.activedirectory.rotate_credentials(name,[mount_point])

### DIFF
--- a/hvac/api/secrets_engines/active_directory.py
+++ b/hvac/api/secrets_engines/active_directory.py
@@ -179,3 +179,18 @@ class ActiveDirectory(VaultApiBase):
         return self._adapter.get(
             url=api_path,
         )
+
+    def rotate_credentials(self, name, mount_point=DEFAULT_MOUNT_POINT):
+        """This endpoint triggers a force rotate LDAP password for the associated account
+
+        :param name: Specifies the name of the role to request credentials from.
+        :type name: str | unicode
+        :param mount_point: Specifies the place where the secrets engine will be accessible (default: ad).
+        :type mount_point: str | unicode
+        :return: The response of the request.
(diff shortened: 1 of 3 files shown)


hkgnobody:
Ref: https://developer.hashicorp.com/vault/api-docs/secret/ad#rotate-role-credentials

Tylerlhess:
While this is a simple endpoint call passing to the api it might be worth a simple sanity check of a test case

hkgnobody:
@Tylerlhess not sure if it is what you suggested to have. 
Otherwise we can add another test to check the "last_password" after invoking rotate_credentials() is matching the "password" of the previous mocked request?

hkgnobody:
I made a mistake. implemented active_directory but make test case on ldap. Convert to DRAFT.

Tylerlhess:
> @Tylerlhess not sure if it is what you suggested to have. Otherwise we can add another test to check the "last_password" after invoking rotate_credentials() is matching the "password" of the previous mocked request?

Yeah that would actually be a good test. I think of having a test, even a dumb one, is a good way to make sure that other changes don't break this function from even being called. Yes validating that it actually worked is important but even a simple test that it runs without error catches most breaking changes.

hkgnobody:
I believe the [ad](https://developer.hashicorp.com/vault/api-docs/secret/ad) and [ldap](https://developer.hashicorp.com/vault/api-docs/secret/ldap) are very similar. That's why in the code base, they are almost identical. 
Therefore i am going to implement the rotate role credentials for both modules. 
Test case on expeccted status code was added for ldap, but i cannot locate the test file for AD.

PR #1172, opened 2024-06-04
 Implement the /sys/wrapping/wrap endpoint

### LINKED ISSUE #513: function /sys/wrapping/wrap
Is this function not avaliable? https://www.vaultproject.io/api/system/wrapping-wrap.html

if so, why? is it just an implementation choice?

clealb:
* Resolves #513
* New function wrap added that returns the response wrapped dictionary it was given

### DIFF
--- a/hvac/api/system_backend/wrapping.py
+++ b/hvac/api/system_backend/wrapping.py
@@ -28,7 +28,7 @@ class Wrapping(SystemBackendMixin):
             json=params,
         )
 
-    def wrap(self, payload=dict(), ttl=60):
+    def wrap(self, payload=None, ttl=60):
         """Wraps a serializable dictionary inside a wrapping token.
 
         Supported methods:
@@ -41,6 +41,9 @@ class Wrapping(SystemBackendMixin):
         :return: The JSON response of the request.
         :rtype: dict
         """
+
+        if payload is None:
+            payload = {}
 
         api_path = "/v1/sys/wrapping/wrap"
         return self._adapter.post(
--- a/hvac/api/system_backend/wrapping.py
+++ b/hvac/api/system_backend/wrapping.py
@@ -27,3 +27,24 @@ class Wrapping(SystemBackendMixin):
             url=api_path,
             json=params,
         )
+
+    def wrap(self, payload=dict(), ttl=60):
+        """Wraps a serializable dictionary inside a wrapping token.
+
+        Supported methods:
+            POST: /sys/wrapping/wrap. Produces: 200 application/json
+
+        :param payload: Specifies the data that should be wrapped inside the token.
+        :type payload: dict
+        :param ttl: The TTL of the returned wrapping token.
+        :type ttl: int
+        :return: The JSON response of the request.
+        :rtype: dict
+        """
+
+        api_path = "/v1/sys/wrapping/wrap"
+        return self._adapter.post(
+            url=api_path,
+            json=payload,
+            headers={"X-Vault-Wrap-TTL": "{}".format(ttl)}
+        )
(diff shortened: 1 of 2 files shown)


hvac/api/system_backend/wrapping.py
```diff
@@ -27,3 +27,24 @@ def unwrap(self, token=None):
...
+            POST: /sys/wrapping/wrap. Produces: 200 application/json
+
+        :param payload: Specifies the data that should be wrapped inside the token.
+        :type payload: dict
+        :param ttl: The TTL of the returned wrapping token.
+        :type ttl: int
+        :return: The JSON response of the request.
+        :rtype: dict
+        """
+
```
jeffwecan, line 44: This presumably runs into the 'ole [mutable default argument "gotcha"](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). To accomplish what you're after (i.e., an empty `payload` dict when no argument explicitly provided), it would be better to go with something like this:

```suggestion
    def wrap(self, payload=None, ttl=60):
        """Wraps a serializable dictionary inside a wrapping token.

        Supported methods:
            POST: /sys/wrapping/wrap. Produces: 200 application/json

        :param payload: Specifies the data that should be wrapped inside the token.
        :type payload: dict
        :param ttl: The TTL of the returned wrapping token.
        :type ttl: int
        :return: The JSON response of the request.
        :rtype: dict
        """
        if payload is None:
                payload = dict()
```

briantist: Agreed, though I would recommend `payload = {}`

clealb:
I added a commit to include your suggestions. Thanks for the review.

briantist:
One other thing: would you be able to implement some tests for this endpoint? You can see other tests (units and integration) in the project that could serve as a guide.

clealb:
The liniting error should be resolved. I also added an integration test for the new method.

tests/integration_tests/api/system_backend/test_wrapping.py
```diff
@@ -8,14 +8,12 @@ class TestWrapping(HvacIntegrationTestCase, TestCase):
     TEST_AUTH_METHOD_TYPE = "approle"
     TEST_AUTH_METHOD_PATH = "test-approle"
 
-    def setUp(self):
-        super().setUp()
```
briantist, line 12: What is the reason to remove the `setUp` method?

tests/integration_tests/api/system_backend/test_wrapping.py
```diff
@@ -31,3 +29,11 @@ def test_unwrap(self):
         logging.debug("unwrap_response: %s" % unwrap_response)
         self.assertIn(member="secret_id_accessor", container=unwrap_response["data"])
         self.assertIn(member="secret_id", container=unwrap_response["data"])
+
+    def test_wrap(self):
```
briantist, line 33: Thank you for including a test! We're missing coverage for `payload=None`, maybe `@parameterized.expand()` can be used to generalize the test?
https://github.com/hvac/hvac/blob/main/tests/integration_tests/api/system_backend/test_health.py#L17

tests/integration_tests/api/system_backend/test_wrapping.py
clealb: I moved the code from there to the unwrap function because it caused an invalid request error
after i added the test_wrap function. i think this is because it gets called twice. and since the test_wrap
doesn't need it I moved it into the test_unwrap function.

tests/integration_tests/api/system_backend/test_wrapping.py
clealb: I added `@parameterized.expand()`. The coverage should be at 100% now.

briantist: great! will need to reformat again with `black` though

clealb: oh. done

briantist (approved):
@clealb thanks for adding this and implementing tests!

PR #1176, opened 2024-06-21
feat: Raft Autopilot endpoints

### LINKED ISSUE #1165: Add raft autopilot set-configuration option
We currently need to enable dead server pruning, but as far as I can see [here](https://hvac.readthedocs.io/en/stable/_modules/hvac/api/system_backend/raft.html#Raft.read_raft_config), there is no method to set this parameter through hvac. Vault also does not provide this as an option in the config file either.

If needed and helpful, I can open a PR for this feature in the future.

kayra1:
closes #1165

This adds 3 endpoints from the autopilot feature of Vault as described in the issue #1165 .

I locally validated and wrote tests for these functions, but I added a skip reason stating that the integration tests currently only support Consul. I am not too familiar with the testing setup in this repo, but let me know if there is anything I can address there.

Thanks.

### DIFF
--- a/docs/usage/system_backend/raft.rst
+++ b/docs/usage/system_backend/raft.rst
@@ -131,7 +131,7 @@ Read Raft Cluster State
     client.sys.read_raft_cluster_state()
 
 Read Raft Autopilot Config
------------------------
+--------------------------
 
 :py:meth:`hvac.api.system_backend.Raft.read_raft_autopilot_config`
 
@@ -143,7 +143,7 @@ Read Raft Autopilot Config
     client.sys.read_raft_autopilot_config()
 
 Update Raft Autopilot Config
------------------------
+----------------------------
 
 :py:meth:`hvac.api.system_backend.Raft.update_raft_autopilot_config`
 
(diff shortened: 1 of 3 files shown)


briantist:
thanks @kayra1 and welcome! It'll take me a little time to review this more closely. In the meantime, take a look at the lint/test runs in CI, consider running linting, formatting, and tests locally to enable more rapid iteration so you don't have to wait for CI (I have to approve each new push for first time contributors).

cc @tot19 since I think you were interested as well

kayra1:
will do, thanks

kayra1:
hello @briantist . I was just wondering if there was anything missing from my end about this.

briantist:
@kayra1 I have not had a chance to look over it in detail. I would like to see more test coverage if possible. I know we don't use raft in the tests at all so integration is probably not feasible but unit tests might be.

kayra1:
@briantist I was basing the files i changed to this [PR](https://github.com/hvac/hvac/pull/1169), which seems to have done something similar. I will investigate the unit tests and add some if I can, and let you know if I run into any issues.

PR #1182, opened 2024-07-27
docs: fix AppRole.login docstring

evan-cohen:
Corrects the return description in the docstring of of `api.auth_methods.AppRole.login`


briantist (approved):
@evan-cohen thanks for this documentation fix!

PR #1185, opened 2024-08-06
Add decode token API endpoint

### LINKED ISSUE #1184: Use decode-token as part of root token generation
While looking at root token generation, I noticed the decode-token end point was not being used. In the test scripts, youare building a command line run to decode. Adding this would simplify using vault to decode the OTP encoded token at the end of the root generation.

The API has been there since 1.13 (as far we can currently go in the docs) and is located here : https://developer.hashicorp.com/vault/api-docs/system/decode-token

alainchiasson:
Adding decode-token to key.py simplifying the final decrypting/decoding of a generated root token. This uses the API endpoint sys/decode-token present since at least 1.13.x

Other than the function, I had rewritten two test cases to use decode token instead of the util.decode_genreated_root_token codethat forked a Vault CLI call ( vault operator ... ), that in the end called the same API 
 endpoint.

util.decode_generated_root_token was also removed, as it is only used by the test.

This also resolves #1184


alainchiasson:
Also added and tested the documentation.

alainchiasson:
I see that tests fail as the sys/decode-token was only added in 1.13.x

While I added the util.decode_generated_root_token and added an if lt 1.13 to select the util or my new code. The two questions are:                                     

- 1.13 is more than 3 versions back, do we stil need it ? I already made the change in the tests.                                                                                                                                  
- How do implement that at the code level - or just let it do a "Path not supported" exception ? Should I make an effort to add to an exception ? 

The failing documentation testing though, I'm not certain what version is being used - my installed version is 1.16

alainchiasson:
Just a quick follow-up. I stopped "fidling" with the code. Is there anything I can do to help move this forward? I have implemented a decode that can be done locally in HVAC - the code implements what is being done server side ( though go -> python ).

briantist:
thanks @alainchiasson ! I do apologize for the slow turnaround, just have a bit too much going on lately and that leaves less time I can dedicate to projects like this but hopefully I'll see some time opening up soon

PR #1186, opened 2024-08-09
Add num_uses and ttl parameters to Approle auth secret id generation

atoom:
Add the `num_uses` and `ttl` parameters to the `generate_secret_id` and `create_custom_secret_id` api methods instead of relying  on the approle default values for these parameters.

### DIFF
--- a/hvac/api/auth_methods/approle.py
+++ b/hvac/api/auth_methods/approle.py
@@ -237,6 +237,8 @@ class AppRole(VaultApiBase):
         metadata=None,
         cidr_list=None,
         token_bound_cidrs=None,
+        num_uses=None,
+        ttl=None,
         mount_point=DEFAULT_MOUNT_POINT,
         wrap_ttl=None,
     ):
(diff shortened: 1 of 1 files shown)


hvac/api/auth_methods/approle.py
```diff
@@ -237,6 +237,8 @@ def generate_secret_id(
         metadata=None,
         cidr_list=None,
         token_bound_cidrs=None,
+        num_uses=None,
+        ttl=None,
```
briantist, line 241: these need to be at the end of the parameter list in case of clients using position parameters

briantist (commented):
thanks for submitting this! could you add tests for the new parameters?

hvac/api/auth_methods/approle.py
atoom: @briantist fixed positional arguments and added some tests. Integration tests are the only thing actually verifying something but I still added the new arguments to the unit tests as well...

PR #1187, opened 2024-08-09
Add Feature Namespace Selector in Kubernetes create_role

kblase:
- Update function to reflect API capability
- Update unit and integration tests
- Update docs for kubernetes auth method

### DIFF
--- a/hvac/api/auth_methods/kubernetes.py
+++ b/hvac/api/auth_methods/kubernetes.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 """Kubernetes methods module."""
+import json
 from hvac import utils
+from hvac import exceptions
 from hvac.api.vault_api_base import VaultApiBase
 from hvac.utils import (
     validate_list_of_strings_param,
@@ -112,7 +114,8 @@ class Kubernetes(VaultApiBase):
         self,
         name,
         bound_service_account_names,
-        bound_service_account_namespaces,
+        bound_service_account_namespaces=None,
+        bound_service_account_namespace_selector=None,
         ttl=None,
         max_ttl=None,
         period=None,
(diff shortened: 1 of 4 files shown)


kblase:
Hey, 
could you please take a look at this PR and give some Feedback?
@hvac/hvac-maintainers

hvac/api/auth_methods/kubernetes.py
```diff
@@ -112,7 +113,8 @@ def create_role(
         self,
         name,
         bound_service_account_names,
-        bound_service_account_namespaces,
+        bound_service_account_namespaces=None,
+        bound_service_account_namespace_selector=None,
```
briantist, line 117: Right now, we require any new parameters to be at the end of the list, so that it doesn't break clients that use positional parameters (I'd like to eventually make most parameters in the library kw-only but that's a large undertaking).

We'll also need to account for `bound_service_account_namespaces` becoming optional now, I'd like to see the tests reflect that change by testing both no value passed as well as an explicit value.

kblase: I added an integration testcase where no `bound_service_account_namespaces` is empty. Do you need another test?

For the other concern I can restructure the parameters to append them at the end. Can you give me an example inside this repo to help me understand what the end result should look like?

briantist: No need to worry about kw-only right now.

For the test that might be ok yah, I don't see the CI run anymore and for some reason no coverage report was posted so let's see how that shakes out once the other change is pushed.

kblase: It seems like there are just doctests running.
When testing locally I get CA Errors. This seems not to be
```
FAILED tests/integration_tests/api/auth_methods/test_kubernetes.py::TestKubernetes::test_configure_5_missing_kubernetes_ca_cert - AssertionError: InvalidRequest not raised
```
and on test_list_roles I get:
```
FAILED tests/integration_tests/api/auth_methods/test_kubernetes.py::TestKubernetes::test_list_roles_0_success - hvac.exceptions.InvalidRequest:  The provided CA PEM data contains no valid certificates, on post https://127.0.0.1:48195/v1/auth/kubernetes-test/config
```

I can push another change when needed to trigger the CI again.

briantist: sorry I missed that last push. I've re-run CI. Looks like you've got some lint/formatting errors and more importantly the integration tests are failing, new error though, no certificate errors as far as I can tell.

I recommend running formatting and tests locally to give yourself a better feedback loop.

PR #1191, opened 2024-10-01
Added support for password policy api

kondotak:
added support for password policy api and associated unit tests

### DIFF
--- a/hvac/api/system_backend/policies.py
+++ b/hvac/api/system_backend/policies.py
@@ -234,3 +234,100 @@ class Policies(SystemBackendMixin):
         return self._adapter.delete(
             url=api_path,
         )
+
+    def list_password_policies(self):
+        """List all configured password policies.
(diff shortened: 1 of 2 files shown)


kondotak:
Is there a development version of enterprise available that I can use to correct the other failures?

tot19:
+1 from me on this. I could definitely use it in my environment. Still think this needs doc pages added. 

For enterprise only features, I have been adding integration tests that I run in my own environment, but then I am adding skips to them. You can check any of my commits for examples!

kondotak:
> +1 from me on this. I could definitely use it in my environment. Still think this needs doc pages added.
> 
> For enterprise only features, I have been adding integration tests that I run in my own environment, but then I am adding skips to them. You can check any of my commits for examples!

Thanks for the info, added the skipIf decorator and updates rst for docs

PR #1195, opened 2024-11-22
docs: remove notice for EOL python support dropping in 2022

clintonsteiner:
First thing you read on readthedocs


clintonsteiner:
Dropped 3.8 support here https://github.com/hvac/hvac/pull/1196

briantist:
What's the reasoning for removing the notice instead of changing the wording?

clintonsteiner:
Feel like it goes without saying to support "in-life" python versions, which is what would be left after the notice about policy changing after 2022

PR #1196, opened 2024-11-22
Remove python 3.8 and vault versions 1.11 and 1.12 for next major release

clintonsteiner:


briantist:
In accordance with [the policy](https://github.com/hvac/hvac/issues/877) this needs to be done in a major release.

clintonsteiner:
Sounds like this just needs to wait for a major release in that case? Thanks for taking the time to review

briantist:
it looks like there are a number of failures to address before this could be merged

clintonsteiner:
Rebased on other branches - this change can be merged before the next major release and features the removals of functionality
https://github.com/hvac/hvac/pull/1201 adds python 3.13 support
https://github.com/hvac/hvac/pull/1200 bumps gha version
https://github.com/hvac/hvac/pull/1199 adds vault version 1.17 and 1.18 for testing

briantist:
Thanks for these PRs @clintonsteiner ! I'm trying to figure out what's causing the timeouts in Vault 1.16+. Even https://github.com/hvac/hvac/pull/1183 which is an old dependabot PR that was passing previously is having the issue now so I don't think it has anything to do with your submissions, but we should have them passing. Since #1199 is also demonstrating the issue consistently in the new Vault versions I suspect some change was made in Vault and backported to 1.16 I just don't know what it is yet.

PR #1198, opened 2024-12-05
Add method to set the default issuer

### LINKED ISSUE #1197: Add method to set the default issuer
While v2.3.0 provides the method `hvac.api.secrets_engines.pki.list_issuers()`, the method `hvac.api.secrets_engines.pki.set_issuers()` to configure the default PKI issuer is missing.

stephan2012:
Add the missing method `hvac.api.secrets_engines.pki.set_issuers()` to configure the default PKI issuer.

Fixes #1197

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,9 +16,9 @@ copyright = "2018-2020, Ian Unruh, Jeffrey Hogan"
 author = "Ian Unruh, Jeffrey Hogan"
 
 # The short X.Y version
-version = "2.3.0"
+version = "2.4.0"
 # The full version, including alpha/beta/rc tags
-release = "2.3.0"
+release = "2.4.0"
 
 
 # -- General configuration ---------------------------------------------------
--- a/hvac/api/secrets_engines/pki.py
+++ b/hvac/api/secrets_engines/pki.py
@@ -799,7 +799,21 @@ class Pki(VaultApiBase):
         )
 
     def set_issuers(self, params, mount_point=DEFAULT_MOUNT_POINT):
-        """Set issuers configuration"""
+        """Set issuers configuration.
+
+        Configure the default issuer.
+
+        Supported methods:
+
+            POST: /{mount_point}/config/issuers. Produces: 200 application/json
+
+        :param mount_point: The "path" the method/backend was mounted on.
+        :type mount_point: str | unicode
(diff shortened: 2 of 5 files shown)


hvac/api/secrets_engines/pki.py
```diff
@@ -798,6 +798,30 @@ def list_issuers(self, mount_point=DEFAULT_MOUNT_POINT):
             url=api_path,
         )
 
+    def set_issuers(self, params, mount_point=DEFAULT_MOUNT_POINT):
```
briantist, line 801: This method should be probably be named `set_issuers_configuration` since it corresponds to that API: https://developer.hashicorp.com/vault/api-docs/secret/pki#set-issuers-configuration

We also don't want to take a generic `params` parameter. Instead we should accept a parameter called `default` (or maybe `default_issuer` to be more specific), and a `default_follows_latest_issuer` parameter so that these can be validated and documented properly.

The tests should then reflect these changes and test for success and failure conditions.

docs/conf.py
```diff
@@ -16,9 +16,9 @@
 author = "Ian Unruh, Jeffrey Hogan"
 
 # The short X.Y version
-version = "2.3.0"
+version = "2.4.0"
 # The full version, including alpha/beta/rc tags
-release = "2.3.0"
+release = "2.4.0"
```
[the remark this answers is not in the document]
stephan2012: Done.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "hvac"
-version = "2.3.0"
+version = "2.4.0"
```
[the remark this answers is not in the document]
stephan2012: Done.

hvac/api/secrets_engines/pki.py
stephan2012: Using `set_issuers_configuration` vs. `set_issuers` is somewhat inconsistent with `list_issuers` (titled [_Read issuers configuration_](https://developer.hashicorp.com/vault/api-docs/secret/pki#read-issuers-configuration) in the API docs, by the way).

stephan2012: `default_follows_latest_issuer` is optional in the HTTP API. Taking it as an argument with an implicit default would probably unintentionally change the configured value.

PR #1201, opened 2024-12-16
add support for python 3.13

clintonsteiner:
* greenlet package required version bump to 3.10 which is minimum with support for python3.13
* https://github.com/hvac/hvac/pull/1196 per pr - python 3.8 will be dropped in major release


clintonsteiner:
Test failures are hopefully fixed in https://github.com/hvac/hvac/pull/1205

Issue #1209, opened 2025-01-19
consul 1.20.2 causes test_health.py to fail


deamen:
While working on writing integration tests for a PR using test_health.py as a reference, I encountered persistent timeout errors. After thorough investigation over several days, I identified the issue: Consul version 1.20.x caused the Vault HA cluster to stop responding to queries, leading to timeouts.

Reverting to Consul version 1.19.2 resolved the issue, allowing the integration tests to execute successfully without timeouts.

I believe the failures in https://github.com/hvac/hvac/actions/runs/12662420080/job/35341372162 is caused by this.

Here is the script I use to troubleshoot:

```
#!/usr/bin/python3
import subprocess
import hvac
import json
import time
import warnings
from urllib3.exceptions import InsecureRequestWarning
import requests

# Suppress only the InsecureRequestWarning
warnings.simplefilter("ignore", InsecureRequestWarning)

# Kill existing consul and vault processes
subprocess.run(["pkill", "-f", "consul agent"])
subprocess.run(["pkill", "-f", "vault-ha"])

consul_http_port = 8500
consul_command = [
    "consul",
    "agent",
    "-dev",
    "-disable-host-node-id",
    "-serf-lan-port=44695",
    "-serf-wan-port=54333",
    "-server-port=56835",
    "-grpc-port=-1",
    "-grpc-tls-port=-1",
    "-bind=127.0.0.1",
    f"-http-port={consul_http_port}",
    "-dns-port=-1",
]

consul_process = subprocess.Popen(
    consul_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE
)

time.sleep(5)

vault_config = {
    "listener": {
        "tcp": {
            "tls_cert_file": "tests/config_files/server-cert.pem",
            "tls_key_file": "tests/config_files/server-key.pem",
            "address": "127.0.0.1:8200"
        }
    },
    "disable_mlock": True,
    "default_lease_ttl": "768h",
    "max_lease_ttl": "768h",
    "storage": {
        "consul": {
            "address": "127.0.0.1:8500",
            "path": "vault/"
        }
    }
}
vault_addr = "https://127.0.0.1:8200"

vault_config_path = "/tmp/vault-ha-node1.json"
with open(vault_config_path, "w") as f:
    f.write(json.dumps(vault_config))

vault_command = ["vault", "server", "-config=" + vault_config_path]

vault_process = subprocess.Popen(
[...]

PR #1211, opened 2025-01-20
Add get_leader method to get the leader node from a list of vault urls

### LINKED ISSUE #552: HighAvailablity ho to use the client
Hi all, 
I've setted up two vault machine (namely `vault1` and `vault2`) for HA (fallback).

The clients allow only a single url to be put as client url. i've set `vault1.mydomain.com`

If the `vault1` restarts then it becomes sealed and looses all the infos regarding the HA, that's to send the data to `vault2` . At the same time, `vault2` knows to be the active one. Bascially the HA does not work.

How can I let the code to switch to the master? 
is there a way to put a list of url and let the system check what to use? does it automatically check the `ha_status` ? or should I implement a fallback when possible?

or there should be something configured somewhere else? (ex: have NGINX in front that does some checkings and stuff)

many thanks.

deamen:
This PR fixes #552 by adding a simple, client-side way to handle high availability with Vault. Basically, it figures out the leader node from a list of Vault URLs, so smaller organizations don’t have to worry about setting up (or paying for) a separate load balancer.

This PR is blocked by #1209 , the integration test will always fail with `Read time out` message when using consul 1.20.2 and 1.20.1

### DIFF
--- a/docs/usage/system_backend/leader.rst
+++ b/docs/usage/system_backend/leader.rst
@@ -54,7 +54,7 @@ Get Leader
 Examples
 ````````
 
-.. testcode:: sys_get_leader
+.. code:: python
 
     import hvac
     client = hvac.Client(cluster_url=['https://127.0.0.1:8200', 'https://127.0.0.1:8202','https://127.0.0.1:8204'])
@@ -63,7 +63,5 @@ Examples
     print('Leader: %s' % leader)
 
 Example output:
-
-.. testoutput:: sys_get_leader
 
     Leader is: https://127.0.0.1:8200
--- a/docs/usage/system_backend/leader.rst
+++ b/docs/usage/system_backend/leader.rst
@@ -44,3 +44,26 @@ Examples
 
     client = hvac.Client(url='https://127.0.0.1:8200')
     client.sys.step_down()
+
+Get Leader
+------------------
+
+.. automethod:: hvac.api.system_backend.Leader.get_leader
+   :noindex:
+
+Examples
+````````
+
+.. testcode:: sys_get_leader
+
+    import hvac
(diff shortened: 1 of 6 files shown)


deamen:
I need some help with the `Documentation Test`:

```
File "usage/system_backend/leader.rst", line 113, in sys_get_leader
Failed example:
    import hvac
    client = hvac.Client(cluster_url=['https://127.0.0.1:8200/', 'https://127.0.0.1:8202/','https://127.0.0.1:8204/'])

    leader = client.sys.get_leader()
    print('Leader: %s' % leader)
Exception raised:
      File "/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/doctest.py", line 1368, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest sys_get_leader[0]>", line 4, in <module>
        leader = client.sys.get_leader()
                 ^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/hvac/hvac/hvac/api/system_backend/leader.py", line 69, in get_leader
        raise LeaderNotFoundError("No leader found in the cluster URL list")
    hvac.exceptions.LeaderNotFoundError: No leader found in the cluster URL list
```
Since the URL I provided in the example are fake ones, the example is supposed to fail.

How can I get it to pass the test?

deamen:
The `Documentation Tests` failure is caused by installing greenlet 3.03 with python3.13 as in https://github.com/python-greenlet/greenlet/issues/392#issuecomment-2446961814.

Don't know why python3.13 is used in the test, I cannot reproduce in ubuntu container with manual installation.

deamen:
@briantist  I think I have figured out why the `Documentation Tests` is failing in https://github.com/hvac/hvac/issues/1212. In your opinion, is there anything else I need to do to get this PR merged?

PR #1215, opened 2025-03-11
Typo corrections

klukashka:


hvac/api/secrets_engines/kv_v2.py
```diff
@@ -283,7 +283,7 @@ def delete_secret_versions(self, path, versions, mount_point=DEFAULT_MOUNT_POINT
         :type path: str | unicode
         :param versions: The versions to be deleted. The versioned data will not be deleted, but it will no longer be
             returned in normal get requests.
-        :type versions: int
+        :type versions: list of int
```
briantist, line 286: I believe the correct format would be like this
```suggestion
        :type versions: list[int]
```

klukashka: OK, I thought you type hint it in such a way. I replaced it to 'list[int]'

klukashka:
why is name parameter (which should be a string, I guess) set to False?
https://github.com/hvac/hvac/blob/main/hvac/api/secrets_engines/pki.py#L684

klukashka:
There is used a built-in 'type' as a variable. 
https://github.com/hvac/hvac/blob/main/hvac/api/secrets_engines/pki.py#L515
There is also such a thing in other files, that is an example.
Shouldn't it be replaced with something like '_type'?

klukashka:
Another question. What is the actual type of 'allowed_users_template'? There are no examples of usage for ssh in the docs and in the method its default value is an empty string, though it's mentioned as boolean.
https://github.com/hvac/hvac/blob/main/hvac/api/secrets_engines/ssh.py#L129

PR #1216, opened 2025-03-12
Changed greenlet to strong 3.0.0 version

klukashka:


briantist:
Unfortunately it looks like that version has broken more stuff, please feel free to try more things here. It might also be helpful to reproduce the issue locally so that you can iterate faster. Thanks for your help on this!

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
```
briantist, line 1: Please use poetry 1.6.1 for this, sometimes new versions introduce changes are not compatible

Issue #499, opened 2019-07-15
Hvac accepts a session that is not a requests Session


surfer190:
This may be expected functionality, but you can send a dict as the session

Recreate with:

```
import hvac

session = {'hello': 'world'}
client = hvac.Client(session=session)

print(client.is_authenticated())
```

Version of hvac: `hvac==0.9.3`

Can we add a check to ensure that the session is an instance of `requests.Session`?

jeffwecan:
Yeah agreed that behavior is a bit funky. Adding a conditional to validate the correct class / parent class seems reasonable.

surfer190:
Running this code now gives:

    Traceback (most recent call last):
      File "/hvac_bug/check.py", line 5, in <module>
        client = hvac.Client(session=session)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/hvac_bug/env/lib/python3.11/site-packages/hvac/v1/__init__.py", line 136, in __init__
        self._adapter = adapter(
                        ^^^^^^^^
      File "/hvac_bug/env/lib/python3.11/site-packages/hvac/adapters.py", line 95, in __init__
        if session.verify:
           ^^^^^^^^^^^^^^
    AttributeError: 'dict' object has no attribute 'verify'

This is not a bug but nice to raise exception informing user they are not using the package correctly.

Something like:

```
if not isinstance(session, requests.Session):
    raise TypeError("hvac session is expected to by of type 'requests.Session'")
```

PR #1223, opened 2025-08-01
Validate that session argument is a session object (#499)

ahamilto:
This PR is for Issue #499 and adds both a check that the session parameter is of the correct object and a test case that the correct exception is thrown when a non-session object is passed in. It also updates the test suite to ensure that the session mocks are of the correct type.

### DIFF
--- a/hvac/adapters.py
+++ b/hvac/adapters.py
@@ -7,7 +7,7 @@ from abc import ABCMeta, abstractmethod
 import requests
 import requests.exceptions
 
-from hvac import utils
+from hvac import exceptions, utils
 from hvac.constants.client import DEFAULT_URL
 
 
@@ -90,6 +90,13 @@ class Adapter(metaclass=ABCMeta):
         if not session:
             session = requests.Session()
             session.cert, session.verify, session.proxies = cert, verify, proxies
+        elif not isinstance(session, requests.Session):
+            raise exceptions.ParamValidationError(
+                "unsupported session type argument provided {arg}, supported types: {supported_types}".format(
+                    arg=session,
+                    supported_types=requests.Session,
+                )
+            )
         # fix for issue 991 using session verify if set
         else:
             if session.verify:
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -2,13 +2,14 @@ import pytest
 
 from unittest import mock
 
+import requests
 from hvac.adapters import Adapter
 
 
 class MockAdapter(Adapter):
     def __init__(self, *args, **kwargs):
         if "session" not in kwargs:
-            kwargs["session"] = mock.MagicMock()
+            kwargs["session"] = mock.MagicMock(spec_set=requests.Session())
         super().__init__(*args, **kwargs)
 
     def request(self, *args, **kwargs):
--- a/tests/unit_tests/test_adapters.py
+++ b/tests/unit_tests/test_adapters.py
@@ -319,3 +319,10 @@ class TestAdapterVerify(TestCase):
             c = Client()
         assert c._adapter.session.proxies == proxies
         assert c._adapter.session
+
+    def test_session_object_validation(self):
+        s = {"hello": "world"}
+        with self.assertRaises(exceptions.ParamValidationError) as context:
+            c = Client(session=s)
+
+        self.assertTrue("unsupported session type argument provided" in str(context.exception))


Tylerlhess:
As a second to this concern about the Session objects we will also want to ensure that the sessions are kept. 

I remember having issues with this while writing the certificate handling. I am not sure with the request capable libraries if they are using the same libraries that requests relies on but I remember having to set up the cert logic just above this 4 years ago and seem to remember the sessions objects certificates not being sticky to the client objects. I forget if I was able to resolve that though unfortunately. Just some food for thought as you are editing the code in that area.

ahamilto:
@briantist thank you for the welcome. I am looking to get more involved in HVAC maintenance since it is a library that my company uses heavily and I am trying to get some time allocated to support it. I have also joined the gitter channel though I couldn't tell how active it was.

For this PR, I can try to rework it to check for the "necessary characteristics" of a session-type object. If I can get that working, I'll also try to quickly update the test cases to support the alternative library that was mentioned. I'll do a quick check on the integration tests as well. I was able to run `make test` locally after my changes with no more errors than there were before but I'll do some more extensive testing. If there are unrelated test fixes, would you prefer them as a separate PR or added onto this one?

@Tylerlhess I believe the adapter class pulls the certificate data out of the session object if one is provided so I think that was the fix you implemented. I will try to do some testing around TLS auth to validate. We use Vault TLS authentication extensively at my company so I've got a pretty good handle on it.

briantist:
> @briantist thank you for the welcome. I am looking to get more involved in HVAC maintenance since it is a library that my company uses heavily and I am trying to get some time allocated to support it. 

That's great! Could definitely use help with it, my time for putting into this has dropped to nearly zero. See also:
- https://github.com/hvac/hvac/discussions/994

> I have also joined the gitter channel though I couldn't tell how active it was.
I have never used the gitter channel, I think it may be dead 😅 we do have a discord (which, while not really "active" is alive, and I will see messages there). Link is in the discussion I linked above.


> For this PR, I can try to rework it to check for the "necessary characteristics" of a session-type object. If I can get that working, I'll also try to quickly update the test cases to support the alternative library that was mentioned. 

I don't want to be prescriptive on that alternative library so we don't need to add it to the tests, but definitely open on doing a pseudo-check ("necessary characteristics") that will help to address the concern in #499 (mainly, failing faster and with a better error message).

> I'll do a quick check on the integration tests as well. I was able to run `make test` locally after my changes with no more errors than there were before but I'll do some more extensive testing. If there are unrelated test fixes, would you prefer them as a separate PR or added onto this one?

If separate test changes are needed, they should be in a dedicated PR.

Tylerlhess:
> @Tylerlhess I believe the adapter class pulls the certificate data out of the session object if one is provided so I think that was the fix you implemented. I will try to do some testing around TLS auth to validate. We use Vault TLS authentication extensively at my company so I've got a pretty good handle on it.

All I remember is a week of ripping my hair out over why the certs weren't attaching while building the Client object but would work if you built and authenticated in the same action.  Yeah we started using it with puppet certs in that former life. Simplified sooo much auth.

hvac/adapters.py
```diff
@@ -91,14 +91,16 @@ def __init__(
             session = requests.Session()
             session.cert, session.verify, session.proxies = cert, verify, proxies
         # fix for issue 991 using session verify if set
-        else:
+        elif self.check_session_object(session):
```
Ousret, line 94: just stumbled upon that PR, I really want hvac to be future proof regarding that validation as we're using this in production.

```suggestion
        elif hasattr(session, "adapters") and isinstance(session.adapters, dict):
```

this is largely sufficient.

Tylerlhess: Ok just making sure I am understanding this correctly. All we need to do to verify a session is a valid session object is make sure it has an adapters attribute and that adapters attribute is a dictionary?

Just double checking that this is a universal truth for the other requests compatible libraries. It completely makes sense that it would be that simple. I'm just dotting I's here but I think this sounds true. (It's late but if I don't do it now I will forget I saw it.)

Ousret: > is a valid session object

yes.

> that this is a universal truth for the other requests compatible libraries

to my best knowledge anything that want to be "requests" compatible must implement the concept of "adapters". otherwise it's a breaking change.

ahamilto:
I have simplified the session check per the recommendation of @Ousret . I had to remove the dict type check since that breaks the unit tests (PowerMock does not get the types of all the attributes correct it appears). I also updated the tests to test with both a real and fake session and check for the appropriate response. I attempted to add niquests as well, however, I ran into an issue with Poetry 1.6.1 getting stuck in a dependency loop. I could either rewrite the lock file with poetry 2.0+ (which worked in my testing) and add niquests as a dev dependency for the tests or leave it as is for now since that is not officially supported.

ahamilto:
@briantist I have updated this PR with simplified testing and added tests for niquests. All tests that were passing before the PR still pass as well as the new tests added. Please let me know if there are any other changes or improvements you would like made.

ahamilto:
@briantist I have removed the niquests tests as requested.

PR #1245, opened 2026-04-17
Add read_namespace and patch_namespace to sys/namespaces

### LINKED ISSUE #1242: Add read_namespace and patch_namespace to sys/namespaces
## Feature request: read and patch namespace

The `sys/namespaces` module currently supports `create_namespace`, `list_namespaces`, and `delete_namespace`, but the Vault API exposes two additional endpoints that aren't yet covered:

- **GET `/v1/sys/namespaces/{path}`** — read a single namespace's metadata
- **PATCH `/v1/sys/namespaces/{path}`** — update namespace custom metadata

These are used in enterprise environments to inspect namespace configuration and to update custom metadata without recreating the namespace.

**Vault API reference:**
- https://developer.hashicorp.com/vault/api-docs/system/namespaces

I'd be happy to submit a small PR adding `read_namespace` and `patch_namespace` methods following the existing patterns in `namespace.py`. Just wanted to check in here first in case there's context I'm missing or a preferred approach. Thanks for maintaining such a useful library!

tot19:
Closes #1242

## What this adds

- `read_namespace(path)` — `GET /v1/sys/namespaces/{path}` to retrieve a single namespace's metadata
- `patch_namespace(path, custom_metadata=None)` — `PATCH /v1/sys/namespaces/{path}` to update namespace custom metadata in place

Both follow the existing patterns in `namespace.py` and `system_backend_mixin.py`. Integration tests are included under `tests/integration_tests/api/system_backend/test_namespace.py` with an `@skipIf(not is_enterprise(), ...)` guard, matching the existing test style.

The diff is intentionally minimal — no docs, no changelog, no unrelated changes. Happy to add anything the project needs or adjust based on feedback. Thank you for your time reviewing this!

### DIFF
--- a/docs/usage/system_backend/namespace.rst
+++ b/docs/usage/system_backend/namespace.rst
@@ -60,6 +60,42 @@ Example output:
     {"request_id":"<redacted>","lease_id":"","renewable":false,"lease_duration":0,"data":{"key_info":{"testns/":{"id":"ekiUn","path":"testns/"}},"keys":["testns/"]},"wrap_info":null,"warnings":null,"auth":null}
 
 
+Read Namespace
+--------------
+
+.. automethod:: hvac.api.system_backend.Namespace.read_namespace
+   :noindex:
+
+Examples
+````````
+
+.. testcode:: sys_namespace
+    :skipif: not test_utils.is_enterprise()
+
+    import hvac
+    client = hvac.Client(url='https://127.0.0.1:8200')
+
+    client.sys.read_namespace(path='testns')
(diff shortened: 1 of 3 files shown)


briantist (approved):
The coverage has missing lines, but I think that's just because it's an enterprise only feature and the enterprise versions we have in CI are too old so the tests can't really run in our CI.

briantist (changes requested):
Actually, could you add unit tests at least? That way we get _something_?

PR #1247, opened 2026-04-17
Add License class for /sys/license/status endpoint

### LINKED ISSUE #1244: Add license status endpoint (/sys/license/status)
## Feature request: license status API support

Vault Enterprise exposes a license status endpoint that is useful for automated license monitoring and compliance checks:

- **GET `/v1/sys/license/status`** — returns the current license information including expiry, feature flags, and customer details

This is a single read-only endpoint with no parameters and isn't currently implemented in hvac.

**Vault API reference:**
- https://developer.hashicorp.com/vault/api-docs/system/license

I'd be happy to contribute a small PR adding a `read_license_status` method if that would be useful. It's a trivial addition — just wanted to flag the gap and check if there's a preferred location in the module before proceeding. Thanks!

tot19:
Closes #1244

## What this adds

A new `License` class in `hvac/api/system_backend/license.py` with one method:

- `read_license_status()` — `GET /v1/sys/license/status`

The class is wired into `SystemBackend` alongside the other system backend mixins. An integration test is in `tests/integration_tests/api/system_backend/test_license.py`, guarded by `@skipIf(not is_enterprise(), ...)`.

This is as small as it gets — a single read-only endpoint. Happy to add anything the project needs or adjust if there's a preferred location or naming convention. Thanks for your time!

### DIFF
--- a/docs/usage/system_backend/index.rst
+++ b/docs/usage/system_backend/index.rst
@@ -11,6 +11,7 @@ System Backend
    key
    leader
    lease
+   license
    mount
    namespace
    policies
--- a/docs/usage/system_backend/license.rst
+++ b/docs/usage/system_backend/license.rst
@@ -1,0 +1,22 @@
+License
+=======
+
+.. contents::
+   :local:
+   :depth: 1
+
+Read License Status
+-------------------
+
+.. automethod:: hvac.api.system_backend.License.read_license_status
+   :noindex:
+
+Examples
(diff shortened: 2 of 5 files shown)


briantist (commented):
This looks good to me but has the same issue as the other PR where the lack of modern enterprise versions in our CI means we can't run integration tests on it. Could unit tests be added as well?

